PR #11, opened 2015-01-27
Better embed: Added Custom Stylesheet support and Fixed jQuery loading last

freibuis:
2 changes

1) Added Custom Style Sheet option into the embed editor.
    This will allow the user to make changes from the CMS with out needing to editor the filesystem. 
2) a)Moved the following to the head to allow jQuery to load first
        
        
        
  b) load $javaScriptContent before
```$javaScriptContent .= '<script type="text/javascript">';
```$javaScriptContent .= '   var options = ' . json_encode($options) . ';';
```$javaScriptContent .= '   $(document).ready(function() { EmbedInit(); });';
```$javaScriptContent .= '</script>';
fires. otherwise jQuery will not load
 Since moving to 1.7 jQuery code fails to fire because its loaded after the script. moving this into the head fixed this issue

### CODE REVIEW

**Comment by dasgarner** [comment]
Hi,

Thanks for the contribution. The JavaScript should be at the end of the page so that the page renders faster, not in the head. (common practice for fast loading pages).

The other changes look great to me - so if you could revert your change to the HTML Template, Ill merge.

Thanks,
Dan

**Comment by freibuis** [comment]
The problem with loading the jquery last is if you try to load other frameworks that use jquery they fail because jquery is not loaded yet.  Also bootstrap framework is an example of what needs to be loaded in the head other wise the page will format after page load properly.

I changed page render from 
```
embedscript
<script src="modules/preview/vendor/jquery-1.11.1.min.js"></script>
<script src="modules/preview/xibo-layout-scaler.js"></script>
```
to render
```
<script src="modules/preview/vendor/jquery-1.11.1.min.js"></script>
<script src="modules/preview/xibo-layout-scaler.js"></script>
embedscript
```
that way if we load other frameworks within the "embedscript" then wont fail to load if they rely on jquery. (simpleweatherjs for example)

Do you have any objections to putting the style content () in the HEAD? that way we could use the stylesheet to load js if the user wants and remove formating after page load issues.

**Comment by dasgarner** [comment]
I've no problem with the library being loaded before the script content from the module - I totally agree with you that jquery should be the first JavaScript include.

You've done that by moving the following to the top

``` php
// Include some vendor items
$javaScriptContent .= '<script src="' . (($isPreview) ? 'modules/preview/vendor/' : '') . 'jquery-1.11.1.min.js"></script>';
$javaScriptContent .= '<script src="' . (($isPreview) ? 'modules/preview/' : '') . 'xibo-layout-scaler.js"></script>';
```

However I do still maintain that the JavaScript includes as a whole should go at the bottom of the page just before the `</body>` tag.

STYLESHEETCONTENT is fine in the HEAD

freibuis:
Done, Moved the Javascript content back where you wanted it.

After testing. my issues arisen again for stylesheets the require jquery to be loaded first. "$ is not defined"

The workaround is to use the new style sheet section to load jquery src via a```<script>``` tag there.  if you load the same script version from the same location its ok as the second load is ignored. issue sidestepped. the first load is slow.. all following usages should be instant as its already cached.

Would you be willing for me to add a checkbox that allows the user to load the jquery src in the head? of course this would come with a "***Warning***" that this is not advised unless you have having issues with scripts that require jquery to be loaded first

dasgarner:
Thanks for all this!

Could you provide an example of where you get that error message? I feel like I am missing something  obvious, but in my mind if all the JavaScript is slower than the jQuery include, then jQuery will always be there by the time it comes to use it.

In principle I have no problem with an advanced tab that has a check box for switching that include around - the example will help me with an exact use case.

Cheers

freibuis:
I have debugged my issue. I loaded bootstrap.min.js in the new style sheet area..This loaded before jquery.

libraries need to load from the javascript section.

dasgarner:
Fantastic - glad you have it working.

PR #15, opened 2015-01-30
Fixed typo and removed 'tuttle' from manual URL

Speedy-Gonzalez:
Fixed typo and removed 'tuttle' from manual URL


dasgarner:
Great, thanks!

PR #19, opened 2015-02-07
Update region.data.class.php

Speedy-Gonzalez:
Changed AddRegion width and height to 250, to make new regions better to handle.

I think this would be an improvement because the very small regions are sometimes hard to handle. Sometimes the options menu gets over the hole region so you can't drag you region with the mouse.

### DIFF
--- a/lib/data/region.data.class.php
+++ b/lib/data/region.data.class.php
@@ -83,7 +83,7 @@ class Region extends Data
      * @param $regionid Object[optional]
      * @return string The region id
      */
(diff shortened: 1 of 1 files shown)


dasgarner:
@Speedy-Gonzalez thanks again for helping out with these fixes.

As you are contributing more, would you mind signing the CLA in xibosignage/xibo ? It helps protect the project for the future. We actually don't believe you need to for small changes such as these, but it may be worth it for peace of mind.

PR #27, opened 2015-02-10
Alternative Theme - Presentation Dan

laercionit:


dasgarner:
There are still a lot of library files in this pull request - surely they don't need to be there?

laercionit:
Dan, in this case sent you the theme directory - default-v2 whole.
There is nothing to do merge, is the whole directory so you can see the visual change.

dasgarner:
There are too many changes to pull this into my branch - we need to get only the changed files from the new theme so that we can compare what you have done in some sane way...

I can't look through 800+ files, I just don't have the time!

laercionit:
Dan, I sent all the theme directory of files, regardless update, just so you can see the result.

After its approval, just send the changed files.

PR #34, opened 2015-02-24
Change xibo-forms.js to solve (a part of) issue 262

Speedy-Gonzalez:
Changed the way how style is applied to the active editor instance

### DIFF
--- a/theme/default/js/xibo-forms.js
+++ b/theme/default/js/xibo-forms.js
@@ -37,16 +37,23 @@ var text_callback = function(dialog)
     // Conjure up a text editor
     CKEDITOR.replace("ta_text");
     
-    CKEDITOR.instances["ta_text"].on('instanceReady', function() {
+    CKEDITOR.instances["ta_text"].on('instanceReady', function(activeInstance) {
         var scale = $('#layout').attr('designer_scale');
 
-        $("#cke_ta_text .cke_contents").css({
-            background: $('#layout').css('background-color')
+        // Set something sensible based on the color of the layout background and get the current background color
+        var color = $c.complement($("#layout").data().backgroundColor);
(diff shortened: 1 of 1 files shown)


Speedy-Gonzalez:
Please have a look at the code changes introduced in this pull request, but don't pull yet, there is still a problem left.

It is now possible to switch between the source view and the normal editor view without losing the background color.

Problem: The code sets the complementary color. That works correctly. But after clicking on "Save" the text color is lost and text color will be `color: #333`.
Solution: Choosing a color from the color picker solves the problem, because this adds a `<span></span>` with an style element.

We have to find a way to write the color style to the span after loading the editor without overriding changes made by the user.

Or what do you think about that?

theme/default/js/xibo-forms.js
```diff
@@ -37,16 +37,23 @@ var text_callback = function(dialog)
...
-        });
-        
-        $("#cke_ta_text iframe").css({
-            "background": "transparent"
+        // Set something sensible based on the color of the layout background and get the current background color
+        var color = $c.complement($("#layout").data().backgroundColor);
+        var bg = $("#layout").data().backgroundColor;
+
+        // Apply styles on the active editor instance
+        activeInstance.editor.document.getBody().setStyle('color', color);
```
dasgarner: The intention was to do this on Line 14 - but it seems like I only took into account media with templates (like tickers). Plain old text modules don't have templates and therefore they won't be hit by the code on Line14.

I think that code is correct - but it needs to be changed to output the span if there isn't a template field present (we can tell by the override checkbox field being present)

Speedy-Gonzalez:
Found a better solution. I will open a new pull request to keep the commit history cleaner.

See https://github.com/xibosignage/xibo-cms/pull/35

PR #35, opened 2015-02-26
Issue262 source view leads to color change v2

Speedy-Gonzalez:
Issue https://github.com/xibosignage/xibo/issues/262

I found a better way to fix this problem (and keep most of your code as it is).

Now everything should work. I've tested it with text and ticker fields.

I added also some `#Color#` tags in three ticker templates. Now the color should automatically change to the complementary color.


dasgarner:
Great!! Thanks very much.

PR #47, opened 2015-03-25
jquery.boxfit for RSS Ticker Module

fogsag:


dasgarner:
I think this looks good - and I think I know what you are trying to achieve with it. I have a few questions about whether boxfit should be applied before or after the layout scaler (I am thinking before might be more appropriate? or maybe it doesn't matter).

In any case can we discuss the details on [community](https://community.xibo.org.uk/c/features)?

fogsag:
Thank you for the fast reply! I am not sure how the layout scaler works 
exactly, thanks for the tip, we have to check that. For sure, it is 
better to discuss the details.
I have now created an account and created an issue 
https://community.xibo.org.uk/t/jquery-boxfit-plugin-for-rss-ticker-text-resize/301
Hope that is ok.

-- 
FOGS AG - weblab.li
Birkenweg 6
FL-9490 Vaduz

www.weblab.li
Tel. +423 770 69 69

fogsag:
Messed up

fogsag:
The new one https://github.com/screen-li/xibo-cms/tree/xibo_ticker_resize

PR #52, opened 2015-04-05
Added pointer cursor to preview navigation buttons

Tarabass:
also it seems that the css class page-header is not used any longer, but not sure about that


dasgarner:
Thanks!

PR #53, opened 2015-04-15
Fix problem that API DataSetDelete and DataSetDataDelete always result i...

tkimuranet:
...n failure.

These two API calls fails like this log.
`<errormsg>Undefined property: PermissionManager::$delete</errormsg>`
`<errornum>8</errornum>`
`<errortype>Notice</errortype>`
`<scriptname>/home/tkimuranet/xibo-cms/lib/service/rest.class.php</scriptname>`
`<scriptlinenum>1526</scriptlinenum>`

API module referes PermissionManager's undefined property 'delete'.
Change 'delete' to 'del'.


dasgarner:
Thanks - thats great.

PR #54, opened 2015-04-17
Ticker image from content

fogsag:
Please test/review.
If there is no image attached to the feed as media enclosure, this patch takes the first image from content (if not found from description) of an rss item.

### DIFF
--- a/modules/ticker.module.php
+++ b/modules/ticker.module.php
@@ -964,12 +964,21 @@ class ticker extends Module
 
 			if ($link == NULL) {
 				$dom = new DOMDocument;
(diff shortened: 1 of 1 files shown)


dasgarner:
What's the context for doing this? Do you have an example where the Media Enclosure is empty, but the content contains HTML with an image tag?

fogsag:
Lot of sites (i guess 50%) are not using media attachments.
See Drupal / Wordpress, Facebook Feeds or on http://simplepie.org/demo/ are some example feeds.
for example http://www.engadget.com/rss.xml

PR #59, opened 2015-05-19
page error

designermax:
To enable custom error page.


dasgarner:
This makes perfect sense, thank you.

PR #82, opened 2015-07-15
Forgotten, Hash

designermax:


dasgarner:
The "forgotten" process isn't hooked up to anything on the UI side because i'm not really happy with the way it works.

It should email a time limited nonce rather than a password - and that is what it will do when it gets implemented in 1.8.

That being said your fix is perfectly correct to make it work as it stands now

PR #118, opened 2015-10-16
Add SAML support using php-saml toolkit.

pitbulk:
Notice that in order to execute SLO, the workflow->slo parameter of settings must be enabled and /saml/logout must be accessed.


dasgarner:
I've created a `feature/saml` branch - can you open a new pull request into that branch please? Also would it be possible to include an example `settings.php` (just the SAML bits) in a comment on the new pull request?

Thanks!

PR #119, opened 2015-10-18
Add SAML support using php-saml toolkit.

pitbulk:
Notice that in order to execute SLO, the workflow->slo parameter of settings must be enabled and /saml/logout must be accessed.


dasgarner:
Many thanks!

PR #123, opened 2015-10-30
SAML Integration

dasgarner:


kyledecot:
Any word on if this is going to get merged in? We use shibboleth for our SSO and this would be a great addition.

dasgarner:
We are waiting for a 3rd party to test on a live system as we don't have one - if you'd be happy to do that instead we would certainly be able to merge it based on the results of your testing rather than theirs.

I suspect they will be a number of months yet.

PR #130, opened 2016-01-06
Fix SQL query

Jimmy89:
The menuitem.Sequence was missing in the Select statement, while being used in the Order By statement.
See problem and solution log here:
https://community.xibo.org.uk/t/admin-panel-does-not-show-navigation-bar/4601/9


dasgarner:
This looks fine, thank you!

dasgarner:
https://github.com/xibosignage/xibo/issues/678

deonvisser:
I still have this issue in 1.7.9. Menu only works for Super Admin and not for any other users.

PR #131, opened 2016-01-25
Detect image format, convert to preferred format if necessary and pos…

ajiwo:
Detect image format, convert to preferred format if necessary and possible.

Check the image format from data submitted by a Player and do conversion if
the incoming format is different with preferred format, provided that the php
extension, imagick in this case, is available and enabled, and supports conversion
from the detected format to preferred one.

On successful conversion, the `$screenShot` variable will contain the converted
binary data.
On failure conversion and when the incoming format != preferred format,
the method returns `false`.

On detection failure and conversion failure, the method flow doesn't change.


dasgarner:
Closed due to subsequent PR

PR #132, opened 2016-01-25
Detect image format, convert to preferred format if necessary and possible.

ajiwo:
Please ignore the [previous](https://github.com/xibosignage/xibo-cms/pull/131) PR.

Detect image format, convert to preferred format if necessary and possible.

Check the image format from data submitted by a Player and do conversion if
the incoming format is different with preferred format, provided that the php
extension, imagick in this case, is available and enabled, and supports conversion
from the detected format to preferred one.

On successful conversion, the `$screenShot` variable will contain the converted
binary data.
On failure conversion and when the incoming format != preferred format,
the method returns `false`.

On detection failure and conversion failure, the method flow doesn't change.

xibosignage/xibo#689

### DIFF
--- a/lib/Xmds/Soap4.php
+++ b/lib/Xmds/Soap4.php
@@ -466,10 +466,12 @@ class Soap4 extends Soap
                 $screenShotImg = Img::make($screenShot);
             } catch (\Exception $e) {
                 if ($this->display->isAuditing == 1)
-                    Log::debug($e->getMessage());
+                    Log::debug($imgDriver . " - " . $e->getMessage());
             }
-            if($screenShotImg !== false)
+            if($screenShotImg !== false) {
+                Log::debug("Use " . $imgDriver);
                 break;
+            }
         }
 
         if($screenShotImg !== false) {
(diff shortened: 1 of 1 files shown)


dasgarner:
This code looks great - thank you. Can it be easily tidied up to use Intervention/Image, which is what the rest of 1.8 uses for image manipulation?

http://image.intervention.io/api/make

ajiwo:
If 1.8 is using it, I think the patch should also use it. I'll  take a look and try out Intervention/Image with the patch.

ajiwo:
I can't catch exception from Intervention/Image  as the method throwing the exception is a private,
seems like the library assumes php extension to load is always available, it's the library internals, not CMS related. Maybe I'll file an issue to Intervention/Image later, if it is their bug, I'm not sure. Cannot wrap my code with the library atm.

dasgarner:
That looks really good now - so we have an implementation which converts from the provided format (assuming it can be understood by gd/imagik) and saves as jpg which the web portal knows how to display.

There is just the small matter of your CLA pull request in the main repo and then I will get this merged into 1.8

Thanks

ajiwo:
> so we have an implementation which converts from the provided format (assuming it can be understood by gd/imagik) and saves as jpg which the web portal knows how to display.

Yes, likely so.

dasgarner:
Excellent, thank you for your contribution and work to get to this stage. I will merge that in now.

ajiwo:
It's a pleasure.

PR #137, opened 2016-02-07
Refresh designer preview after reordering timeline item.

ajiwo:
xibosignage/xibo#709

Should be applicable to 1.7 series too.


dasgarner:
That looks fine to me - thank you

PR #180, opened 2016-05-26
Feature - Campaign Preview

dasgarner:
See - https://community.xibo.org.uk/t/campaign-preview/6110 for feature description.


lib/Controller/Campaign.php
```diff
@@ -144,6 +144,16 @@ public function grid()
                 'text' => __('Schedule Now')
             );
 
+            // Preview
+            // // FIXME : check permissions pour ce bouton
```
dasgarner: I don't think you have anything to fix here - we wouldn't be in the loop unless we had view permissions for the Campaign in question.

lib/Controller/Campaign.php
```diff
@@ -577,4 +587,46 @@ public function unassignLayout($campaignId)
...
+     *  operationId="campaignPreview",
+     *  tags={"campaign"},
+     *  summary="Preview Campaign",
+     *  description="Preview an existing Campaign"
+     * )
+     */
+    public function preview($campaignId)
+    {
+        $campaign = $this->campaignFactory->getById($campaignId);
+	// FIXME : pas de checkEditable
```
dasgarner: Same as above - probably only "view" permissions

lib/routes.php
```diff
@@ -155,6 +155,7 @@
 $app->post('/campaign', '\Xibo\Controller\Campaign:add')->name('campaign.add');
 $app->put('/campaign/:id', '\Xibo\Controller\Campaign:edit')->name('campaign.edit');
 $app->delete('/campaign/:id', '\Xibo\Controller\Campaign:delete')->name('campaign.delete');
+$app->get('/campaign/:id/preview', '\Xibo\Controller\Campaign:preview')->name('campaign.preview');
```
dasgarner: This should be in `routes-web.php` as it won't be appropriate to get that HTML back from the API/Console/etc.

web/install/index.php
```diff
@@ -1,133 +0,0 @@
-<?php
```
dasgarner: Please can you revert the deletion of this file.

dasgarner:
I've made some comments in the code for you to review - thanks!

There are also some file mod changes - from non-executable to executable which will need to be reverted before we can merge.

Many thanks,
Dan

LesCrous-Xibo:
Hi Dan, 

I think it's better now.
Thanks to you, 
Gwen

dasgarner:
Great, those changes look good. I'll wait for your subsequent alterations (discussed in the community topic) before merging.

LesCrous-Xibo:
`PHP Fatal error:  Uncaught exception 'PDOException' with message 'SQLSTATE[HY000] [2003] Can't connect to MySQL server on 'mysql' (113)'` 
it seems to be something that has nothing to do with the code

alexharrington:
Yes certainly. I'm just looking at that now for you :)

alexharrington:
Retest this please

LesCrous-Xibo:
Thanks Alex, it seems to work fine now

contributors/LesCrous.md
```diff
@@ -0,0 +1,13 @@
...
+Agreement, version 1.0, with MD5 checksum
+2141746be10ee23794f402369cf644fc.
+
+I furthermore declare that I am authorized and able to make this
+agreement and sign this declaration.
+
+Signed,
+
+Olivier Perrot
+https://github.com/LesCrous
```
dasgarner: Sorry if the instructions were not clear - but this file needs to go in the xibosignage/xibo repository and **not** in this one.

lib/Controller/Campaign.php
```diff
@@ -609,6 +617,7 @@ public function preview($campaignId)
         {
             $duration += $layout->duration ;
             $extendedLayouts[] = ['layout' => $layout,
+                                  'duration' => gmdate('H:i:s', $layout->duration),
```
dasgarner: This formatting should really be the responsibility of the View and not the Controller (because it is display formatting). You can do that in your Twig file by using the `|date` filter on L16 of `views/campaign-preview.twig`

views/campaign-page.twig
```diff
@@ -63,6 +64,7 @@ This is the template for the campaign page
             "columns": [
                 { "data": "campaign" },
                 { "data": "numberLayouts" },
+                { "orderable": false, "data": "totalDuration" },
```
dasgarner: Would H:i:s be good here?

```
"data": function (data, type) {
                            if (type != "display")
                                return data.totalDuration;

                            return moment().startOf("day").seconds(data.totalDuration).format("H:mm:ss");
                        }
```

views/campaign-preview.twig
```diff
@@ -13,19 +13,19 @@ This is the template for the campaign page
...
+            <p><b>{% trans "number of layouts" %}</b> : {{ campaign.numberLayouts }} </p>
             <div class="row clearfix">
-              {% for layout in layouts %}
+              {% for extendedLayout in extendedLayouts %}
                 <div class="col-sm-6 col-md-4 col-lg-3">
                   <div class="embed-responsive embed-responsive-4by3">
-                      <div class="embed-responsive-item preview-container">
-                        {% include 'layout-preview-partial.twig' with {'layout': layout, 'previewOptions': previewOptions } %}
-                      </div>
+                        <div class="embed-responsive-item preview-container" data-url="/layout/preview/{{ extendedLayout.layout.layoutId }}">
```
dasgarner: Should use `{{ urlFor("<<theroutename>>", { "id": extendedLayout.layout.layoutId }) }}` so that the path is always resolved correctly (people might be running in virtual folders, sub folders, etc)

dasgarner:
I've added some more suggestions - I hope you don't mind!

gwen21:
Hi Dan, Thank you for your review, 

I'll consider all your remarks. I'm very sorry for the contributing markdown file pollution in this repository, I'm gonna put it in the `xibosignage/xibo` repository now.

LesCrous-Xibo:
Hi everybody, 

I think it's all done now for this feature. Dan, I have modified the `query()` method in the `CampaignFactory`, it seemed a good idea to me but you are allowed to tell me that it wasn't. 
I'm waiting for your reply.
Cheers,
Gwen

dasgarner:
Hi Gwen,

I think your approach with the `CampaignFactory` is good - it means we don't have the payload of calculating the SUM unless the option is provided, which is great.

I will look through the complete `diff` again today and make sure... but I think everything is OK and it will get merged.

Thanks for your contribution here and taking the time to make these adjustments - much appreciated.

dasgarner:
Merged - thanks Gwen. Best wishes!

LesCrous-Xibo:
Thank you very much Dan and  Alex ! I'm very happy to have contributed to `xibo-cms` and I hope that it is only a beginning. 
Cheers

PR #209, opened 2016-09-23
Feature/1.8.0 template modules

dasgarner:


lib/Widget/ForecastIo.php
```diff
@@ -502,29 +577,44 @@ public function getResource($displayId = 0)
...
+                
+                if (count($quantity) > 1)
+                  $offset = $quantity[2];
+            }
+
             // Pull it out, and run substitute over it for each day
-            $dailySubs = '';
             // Substitute for every day (i.e. 7 times).
-            for ($i = 0; $i < 7; $i++) {
+            for ($i = $offset; $i < $itterations + $offset; $i++) {
```
dasgarner: I am not convinced this is correct - lets say you want all 7 days, starting day 2 - `$itterations = 7` and `$offset = 2`, that means the loop will go until `$i < 9` ?

lib/Widget/ForecastIo.php
```diff
@@ -502,29 +577,44 @@ public function getResource($displayId = 0)
...
-
         // JavaScript to control the size (override the original width and height so that the widget gets blown up )
         $options = array(
             'previewWidth' => $this->getSanitizer()->getDouble('width', 0),
             'previewHeight' => $this->getSanitizer()->getDouble('height', 0),
             'originalWidth' => $this->region->width,
             'originalHeight' => $this->region->height,
-            'scaleOverride' => $this->getSanitizer()->getDouble('scale_override', 0)
+            'scaleOverride' => $this->getSanitizer()->getDouble('scale_override', 0),
+            'widgetDesignWidth' => $this->getOption('widgetOriginalWidth'),
```
dasgarner: It would be great if you could sanitize these so they are integers

`'widgetDesignWidth' => $this->getSanitizer()->int($this->getOption('widgetOriginalWidth')),`

That means they will end up in JS as integers, which I think will work more consistently across browsers.

modules/forecastio/5-day-custom-large-bg.template.json
```diff
@@ -0,0 +1,10 @@
+{
+ "id":"5-day-fullhd-background",
+ "value":"5 days, current day in a customized background",
```
dasgarner: We might want to work on this name - we can discuss

dasgarner:
The 3 failing tests are fixed upstream - no need for them to block this PR.

lib/Widget/ForecastIo.php
dasgarner: Are you sure?

lib/Widget/ForecastIo.php
dasgarner: +1

PR #217, opened 2016-09-30
1.8.0-rc1 - Forecast Templates - Final

dasgarner:


lib/Service/DateServiceInterface.php
```diff
@@ -19,7 +19,7 @@
      * @param string $format
      * @return string
```
dasgarner: It would be nice to add the parameter to the method docblock

maurofmferrao: Done :+1:

PR #222, opened 2016-10-14
Further en_US fixes

guruevi:
This fixes various errors when using the en_US locale.
This also fixes the Layout Editor not working on Touch Screen devices (like tablets).


dasgarner:
Can you confirm that the translations you did inside the `mo` file were originally done in Launchpad? Or do they need to be imported?

The fix for touch events on tables needs to be reviewed - would it be possible to separate that into a separate branch/PR so that it can be pulled cleanly into our test environment? I think we only need the `min.js` file and can leave out the readme, etc.

guruevi:
Hi,

I downloaded the files from LaunchPad but used another editor and then committed to GitHub.

I can separate it out. I pulled it from another Git repo, I'm pretty sure we can get rid of the README (given it does not contain license information). Feel free to reject.

dasgarner:
Launchpad is our master - before each release we download the current translation set from Launchpad and apply it to this repository.

I believe we can import your `en_US.mo` file into Launchpad - @alexharrington is that possible?

Regarding the other changes - yes it would be fantastic if you could separate them out. If you don't mind waiting until the above translation issue is resolved, then you could simply drop the `mo` file from your commit.

I'll make some other comments against your diff regarding the other changes.

Many thanks for your patience (we've been slow replying - sorry!)

web/theme/default/libraries/calendar/js/language/en-US.js
```diff
@@ -0,0 +1,115 @@
...
+	window.calendar_languages = {};
+}
+// Here you define the language and Country code. Replace en-US with your own.
+// First letters: the language code (lower case). See http://www.loc.gov/standards/iso639-2/php/code_list.php
+// Last letters: the Country code (upper case). See http://www.iso.org/iso/home/standards/country_codes/country_names_and_code_elements.htm
+<<<<<<< HEAD
+window.calendar_languages['en-US'] = {
+=======
+window.calendar_languages['en-GB'] = {
+>>>>>>> bd93088372a8d6c95ef285bd5cc17d5a6b574ae3
```
dasgarner: I don't think this should be here? It looks some a line from a failed merge?

web/theme/default/libraries/jquery-ui-touch-punch/README.md
```diff
@@ -0,0 +1,41 @@
+# jQuery UI Touch Punch
```
dasgarner: We don't need to readme or the unminified js - just the `.min.js` will be enough.

alexharrington:
I'm afraid it's not possible to import mo format files in to Launchpad.

If you have the corresponding PO format it may be possible to add them that way?

guruevi:
@alexharrington I looked over the Launchpad site and can't seem to add the PO file. Any where I can send it to?

I also fixed the JavaScript files in a recent commit.

web/theme/default/libraries/calendar/js/language/en-US.js
guruevi: Correct. Fixed it.

web/theme/default/libraries/jquery-ui-touch-punch/README.md
guruevi: Deleted them in a commit

alexharrington:
I've never done it that way myself either, but it does say it can upload PO files. If you send me the file I'm happy to try for you.

dasgarner:
Sorry, but this is not a patch we can take.

guruevi:
I forgot this branch was tracked into this ticket, sorry.

PR #226, opened 2016-11-11
Additional Templates for Finance - Stocks and Currencies

maurofmferrao:
https://github.com/xibosignage/xibo/issues/949


dasgarner:
Can you add the title/description of the module to `locale/dbtranslate.php` so that they end up as translation strings?

Normally this is taken care of by `__()` or `{% trans "" %}` but because the module name/description are in the DB, we also need to add them to the code somewhere.

dasgarner:
> Stock Exchange Info

It might be better to make this configurable? Perhaps as a form option?

dasgarner:
> The duration specified is per page otherwise it is per feed.

Perhaps this should say "per item" instead of "per feed" ?

dasgarner:
Should this have a background colour by default? transparent might be better?

image

dasgarner:
test this please

PR #228, opened 2016-11-30
1.8.0-rc2 bug fixes and small enchancements

dasgarner:


lib/Controller/Campaign.php
```diff
@@ -68,17 +74,19 @@ class Campaign extends Base
      * @param ConfigServiceInterface $config
      * @param CampaignFactory $campaignFactory
      * @param LayoutFactory $layoutFactory
+     * @param TagFactory $tagFactory
```
dasgarner: interesting positioning of this param comment - perhaps it should be at the end? :)

PR #232, opened 2016-12-21
Updates to test suite for 1.8.0-rc2

dasgarner:


dasgarner:
Merged in https://github.com/xibosignage/xibo-cms/pull/233

PR #234, opened 2017-01-05
Bugfix/1.8.0 rc3

dasgarner:


web/modules/xibo-image-render.js
```diff
@@ -0,0 +1,79 @@
...
+            var reloadTimes = $self.data("reload-times");
+            
+            // Loop an infinite number of times ( maxTries == -1 ) or until the loop reach options.maxTries times
+            if (reloadTimes < options.maxTries || options.maxTries == -1) {
+                
+                // Create a timeout using the options reload time
+                setTimeout(function() {
+                        
+                        // Try to change source to the original
+                        $self.unbind("error").attr("src", originalSrc).bind("error", handleImageError);
```
dasgarner: Do you need an unbind here? It should already be unbound at this point?

lib/Widget/DataSetView.php
```diff
@@ -393,11 +394,12 @@ public function getResource($displayId = 0)
 
         $javaScriptContent .= '<script type="text/javascript" src="' . $this->getResourceUrl('xibo-layout-scaler.js') . '"></script>';
         $javaScriptContent .= '<script type="text/javascript" src="' . $this->getResourceUrl('xibo-dataset-render.js') . '"></script>';
+        $javaScriptContent .= '<script type="text/javascript" src="' . $this->getResourceUrl('xibo-image-render.js') . '"></script>';
 
         $javaScriptContent .= '<script type="text/javascript">';
         $javaScriptContent .= '   var options = ' . json_encode($options) . ';';
         $javaScriptContent .= '   $(document).ready(function() { ';
-        $javaScriptContent .= '       $("#DataSetTableContainer").dataSetRender(options); $("body").xiboLayoutScaler(options);';
+        $javaScriptContent .= '       $("#DataSetTableContainer").dataSetRender(options); $("body").xiboLayoutScaler(options); $("#content").find("img").xiboImageRender(options); ';
```
dasgarner: Can we make sure that this should be on `#content` for DataSetView? The other renderer uses `#DataSetTableContainer`

web/modules/xibo-image-render.js
```diff
@@ -0,0 +1,79 @@
...
+                        
+                        // Increase the control var and set it to the element
+                        reloadTimes++; 
+                        $self.data("reload-times", reloadTimes); 
+                }, options.reloadTime);
+            }
+        };
+        
+        // Original image source
+        var originalSrc = $self.attr("src");
```
dasgarner: I wonder if it would be better to store this in `$self.data` ?

web/modules/xibo-image-render.js
maurofmferrao: You're right, that is bound just above that line

lib/Widget/DataSetView.php
maurofmferrao: It was working this way, the Module uses #content as well, but we can use the #DataSetTableContainer as it contains all the DataSet module html.

dasgarner:
test this please

lib/Widget/Stocks.php
```diff
@@ -392,10 +399,36 @@ public function getResource($displayId = 0)
         if (!$items = $this->getYql()) {
             return '';
         }
+        
+        if( $this->getOption('overrideTemplate') == 0 ) {
+            
+            $templates = $this->templatesAvailable();
```
dasgarner: It seems like we do this a lot? Perhaps there is an opportunity to standardise access to these templates by putting methods for:

 - templatesAvailable
 - getTemplateById

Into the `ModuleWidget` class?

alexharrington:
retest this please

PR #262, opened 2017-05-22
Fix this function on a page without table

guruevi:
If we use this function on a page that does not have a global variable named table a bug happens. We cannot test simply to see if the table variable exists because table exists as a built-in function in modern browsers (console.table is aliased to table).

I've tested the "if (backgroundChanges && table)" construction which due to it existing causes weirdness. This is the only way to test if table actually exists.


dasgarner:
This looks fine to me - thank you for your contribution.

PR #263, opened 2017-05-30
Fix object.zIndex is not a function error

guruevi:
zIndex is not actually a function in newer versions of jQuery UI
https://jqueryui.com/upgrade-guide/1.12/#removed-zindex


dasgarner:
Thank you.

How did this bug present itself in the application - I will create an issue for it and link it up.

guruevi:
@dasgarner The issue presented itself when I updated jQuery-UI to the latest version because I ran into a bug with Safari. 

This seems to fix any issues Xibo may have with the newer jQuery library so it should be safe to update to the latest jQuery-UI and a later version of the jQuery core.

PR #274, opened 2017-10-03
Take advantage of "php-saml" toolkit

flysen:
php-saml provide metadata.php to extract and validate the SP and IDP data provided by the array. To make this cleaner I recommend to not use the current settings.php file to add SAML data. I've created a samlSettings.php to hold the array. Also created the folder samltools to host SAML tools useful for the integration. 

metadata.php takes the array parse it and display a xml that you can sent to your 3:rd party federation (IDP).

I have not tested the full work-flow using SAML because I needed to get hold of the metadata first and then I have ~3weeks for approval.

### DIFF
--- a/lib/Helper/Install.php
+++ b/lib/Helper/Install.php
@@ -264,6 +264,10 @@ define('SECRET_KEY', '$secretKey');
 // \$middleware = [];
 // \$authentication = ;
 
+// \$authentication = new \Xibo\Middleware\SAMLAuthentication();
+// \$samlFile = realpath(dirname(__FILE__) . '/samlSettings.php');
+// require_once $samlFile;
+
 END;
 
         if (!fwrite($fh, $settings))
--- a/web/_toolkit_loader.php
+++ b/web/_toolkit_loader.php
@@ -1,0 +1,23 @@
+<?php
+
+// Create an __autoload function 
+// (can conflicts other autoloaders)
+// http://php.net/manual/en/language.oop5.autoload.php
+$libDir = realpath(dirname(__FILE__) . '/../vendor/onelogin/php-saml/lib/Saml2/'). '/';
+$extlibDir = realpath(dirname(__FILE__) . '/../vendor/onelogin/php-saml/extlib/'). '/';
+
+// Load composer
+if (file_exists('vendor/autoload.php')) {
+    require 'vendor/autoload.php';
+}
+
+// Load now external libs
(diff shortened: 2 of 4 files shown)


dasgarner:
Thank you for your PR.

You can do all of this already by putting your settings in `settings.php` and calling `/saml/metadata` from a web browser (or making a web request in another way).

The toolkit, etc is not required.

Unless I am missing something of course - which is very possible!

flysen:
@dasgarner the medtadata.php provide a proper xml parsed file from the SP ready to be signed by the IDP.
Here is an example of the array provided from [array SAML](https://xibo.org.uk/manual/en/users_saml.html):
`<md:EntityDescriptor validUntil="2017-10-05T07:25:30Z" cacheDuration="PT604800S" entityID="http://xibo-cms.example.com/saml/metadata"><md:SPSSODescriptor AuthnRequestsSigned="false" WantAssertionsSigned="false" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol"><md:SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="http://xibo-cms.example.com/saml/sls"/><md:NameIDFormat>urn:oasis:names:tc:SAML:2.0:nameid-format:emailAddress</md:NameIDFormat><md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="http://xibo-cms.example.com/saml/acs" index="1"/></md:SPSSODescriptor></md:EntityDescriptor>`

dasgarner:
Here is the output from the existing route:

image

They are the same, no?

To get this output, you enable the SAML auth middleware as per the instructions - and visit the /saml/metadata route, and then you have your IdP information.

I am reluctant to provide another non-managed application entry point, especially one that includes its own autoloader.

flysen:
@dasgarner :+1: 
Exactly what I was trying to achieve, big thanks. I did two things really wrong.

1. I ended the route with a trailing "/" https://localhost/saml/metadata/ - That is wrong.
2. When I visited the page https://localhost/saml/metadata I did it from a browser which gave me just one line out put:

`urn:oasis:names:tc:SAML:2.0:nameid-format:emailAddress`

Using wget show the complete xml or using "view-source" in a browser (and of course Postman).

Sorry my bad
/Fredrik

dasgarner:
No problem - thank you for contributing!

PR #288, opened 2017-12-12
Various Improvements as documented in issue comments

dasgarner:
xibosignage/xibo#1320

Broadly speaking, the improvements are:
 - fixed some issues installing/upgrading
 - combined into 1 table
 - moved some controller helper logic into the factory
 - adjusted the form text and structure


dasgarner:
@LukyLuke assuming you don't have any objections I will merge this into the feature branch and apply it to some test systems here for more testing.

LukyLuke:
Thanks @dasgarner 
I would be pleased if you merge it :)

I will have time as of next week to improve the RGraph module to use this new Feature better (combine of multiple Datasets and throw away the direct use of webservice APIs).

dasgarner:
@LukyLuke after you've made those changes would you be interested in us looking to see if we could include the rgraph module into the core product also?

PR #338, opened 2018-03-12
DataSet Filter by [DisplayName]

JosefWinkler:
You can filter a DataSet by [DisplayName]. 
So I think it's better to read.

e.g. 
Filter:  Room = [DisplayName]
DataSet:   Room = Room-101

### DIFF
--- a/lib/Entity/DataSet.php
+++ b/lib/Entity/DataSet.php
@@ -379,10 +379,14 @@ class DataSet implements \JsonSerializable
         // Get the Latitude and Longitude ( might be used in a formula )
         if ($displayId == 0) {
             $displayGeoLocation = "GEOMFROMTEXT('POINT(" . $this->config->GetSetting('DEFAULT_LAT') . " " . $this->config->GetSetting('DEFAULT_LONG') . ")')";
+            //Show DataSet-Row on Preview if Filter-KeyField empty
+            $displayName = '';
         }
         else {
             $displayGeoLocation = '(SELECT GeoLocation FROM `display` WHERE DisplayID = :displayId)';
             $params['displayId'] = $displayId;
+            //Get displayName from Database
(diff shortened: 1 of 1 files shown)


lib/Entity/DataSet.php
```diff
@@ -379,10 +379,14 @@ public function getData($filterBy = [], $options = [])
...
         if ($displayId == 0) {
             $displayGeoLocation = "GEOMFROMTEXT('POINT(" . $this->config->GetSetting('DEFAULT_LAT') . " " . $this->config->GetSetting('DEFAULT_LONG') . ")')";
+            //Show DataSet-Row on Preview if Filter-KeyField empty
+            $displayName = '';
         }
         else {
             $displayGeoLocation = '(SELECT GeoLocation FROM `display` WHERE DisplayID = :displayId)';
             $params['displayId'] = $displayId;
+            //Get displayName from Database
+            $displayName = $this->getStore()->select('Select display from display where displayid = '.$displayId ,[])[0]['display'];
```
dasgarner: You'd be better off using the DisplayFactory for this and handling the situation where a display doesn't exist (this would also allow us to extend with other properties if we chose to.

Something like:

```
try {
  $displayName = $display->display;
} catch (NotFoundException $notFoundException) {
  // set them to empty
  $displayName = '';
}
```

lib/Entity/DataSet.php
```diff
@@ -432,6 +436,9 @@ public function getData($filterBy = [], $options = [])
         if ($filter != '') {
             // Support display filtering.
             $filter = str_replace('[DisplayId]', $displayId, $filter);
+            // Support display filtering by [DisplayName]
+            $filter = str_replace('[DisplayName]', $displayName , $filter);
```
dasgarner: We will need to make sure we add single quotes around our display name (if they aren't already in the filter)

PR #340, opened 2018-03-14
Support for text/html content type xibosignage/xibo#1452

PeterMis:


lib/Widget/Ticker.php
```diff
@@ -821,7 +821,7 @@ private function getRssItems($isPreview, $text)
                 $client = new Client();
                 $response = $client->get($feedUrl, $this->getConfig()->getGuzzleProxy([
                     'headers' => [
-                        'Accept' => 'application/rss+xml, application/rdf+xml;q=0.8, application/atom+xml;q=0.6, application/xml;q=0.4, text/xml;q=0.4'
+                        'Accept' => 'application/rss+xml, application/rdf+xml;q=0.8, application/atom+xml;q=0.6, application/xml;q=0.4, text/xml;q=0.4, text/html'
                     ],
```
dasgarner: Is there a reason you missed the `q=` part of the header? I can't recall whether this is valid syntax or not - but perhaps you did that to indicate no relative quality?

PR #345, opened 2018-03-21
Playlists for Swift

dasgarner:
First phase of adding Playlist support into the Xibo CMS for Swift (2.0)


dasgarner:
retest this please

PR #359, opened 2018-04-10
Release18

araby123:
to test with our privet cloud hosting in company


dasgarner:
The release18 branch is for 1.8 and develop is for 2.0 - it is not possible to merge develop into release18

PR #367, opened 2018-04-17
Bugfix/1.8.9 pack2

PeterMis:


lib/Controller/Display.php
```diff
@@ -1538,6 +1539,8 @@ public function setDefaultLayout($displayId)
 
         $display->defaultLayoutId = $layoutId;
         $display->save(['validate' => false]);
+        if ($defaultLayoutId != $display->defaultLayoutId)
```
dasgarner: we have a better way to do this now:

```
$display->hasPropertyChanged('defaultLayoutId)
```

lib/Controller/Display.php
```diff
@@ -1525,6 +1525,7 @@ public function defaultLayoutForm($displayId)
     public function setDefaultLayout($displayId)
     {
         $display = $this->displayFactory->getById($displayId);
+        $defaultLayoutId = $display->defaultLayoutId;
```
dasgarner: unneeded with the below suggestion

web/theme/default/js/xibo-forms.js
```diff
@@ -14,8 +14,9 @@ var text_callback = function(dialog, extraData) {
     var scale = $layout.attr('designer_scale');
     var regionWidth = $("#region_" + $layout.data().currentRegionId).attr("width");
     var regionHeight = $("#region_" + $layout.data().currentRegionId).attr("height");
-    var applyContentsToIframe = function() {
-        $("#cke_ta_text iframe").contents().find("head").append("" +
+    var applyContentsToIframe = function(field) {
+        console.log('Applying iframe adjustments to ' + field);
```
dasgarner: Good to comment out console.logging

PR #371, opened 2018-05-08
Session.php fix remoteaddr from Proxy

levic92:
Session information was displaying the proxy server's IP address instead of client address. I found this issue - https://github.com/xibosignage/xibo/issues/606 - and it seems like it was fixed when referencing display IP addresses but not session, so I added the same function found in https://github.com/xibosignage/xibo-cms/blob/500f3e16fa525c7428a3bc1f98383a0bde6697d0/lib/Xmds/Soap.php#L1735-L1748


dasgarner:
test this please

dasgarner:
This looks great, thank you.

I'll create an issue for it, link it up and then get it merged.

levic92:
Thanks!

dasgarner:
https://github.com/xibosignage/xibo/issues/1509

PR #396, opened 2018-07-26
Docker: Move all root-privileged work from start script to the image

LukyLuke:
Move all root-privileged work from start script to the Dockerfile to be able to use Xibo in a Non-Root configuration like OpenShift.

Due to very strict security policies it's not allowed in OpenShift to rn any root privileged commands inside the docker container. Therefore I moved all setup tasks from the entrypoint.sh to the Dockerfile.


dasgarner:
@alexharrington will have to review this and discuss with you.

docker/entrypoint.sh
```diff
@@ -126,42 +126,21 @@ then
...
 then
-    # Update /etc/periodic/15min/cms-db-backup with current environment (for cron)
-    /bin/sed -i "s/^MYSQL_USER=.*$/MYSQL_USER=$MYSQL_USER/" /etc/periodic/15min/cms-db-backup
-    /bin/sed -i "s/^MYSQL_PASSWORD=.*$/MYSQL_PASSWORD=$MYSQL_PASSWORD/" /etc/periodic/15min/cms-db-backup
-    /bin/sed -i "s/^MYSQL_HOST=.*$/MYSQL_HOST=$MYSQL_HOST/" /etc/periodic/15min/cms-db-backup
-    /bin/sed -i "s/^MYSQL_PORT=.*$/MYSQL_PORT=$MYSQL_PORT/" /etc/periodic/15min/cms-db-backup
-    /bin/sed -i "s/^MYSQL_DATABASE=.*$/MYSQL_DATABASE=$MYSQL_DATABASE/" /etc/periodic/15min/cms-db-backup
+    # Define the secret if not done already
+    SECRET_KEY=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8) && \
+    /bin/sed -i "s/define('SECRET_KEY','');/define('SECRET_KEY','$SECRET_KEY');/" /var/www/cms/web/settings.php
```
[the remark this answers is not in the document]
LukyLuke: It will, but it doesn't matter becasue `define('SECRET_KEY','');` isn't there. It will be replaced on the first run with `define('SECRET_KEY','.........');`

LukyLuke:
This was the plan. To move all commands which needs root permission to the build. So the system will be setup correctly in the image and therefore on the first run (folders, files, permissions, cron).
Only the Environment-Variables will be setup everytime when you start the container. Therefore I moved all the MYSQL-Stuff into a separated file which can be changed every time you start the container and probably define a different environment variable by parameter.

Dockerfile
```diff
@@ -145,9 +145,27 @@ RUN mkdir -p /var/www/cms/library/temp &&  \
     mkdir -p /run/apache2 && \
     rm /etc/apache2/conf.d/info.conf && \
     rm /etc/apache2/conf.d/userdir.conf && \
-    touch /CMS-FLAG && \
-    addgroup ssmtp && \
-    chmod 777 /tmp
+    addgroup ssmtp
+
+# Initial configuration
+RUN test ! -f /var/www/SETUP-COMPLETED && \
```
dasgarner: this file won't ever be there when the container builds, so this will always run - can remove the test and file completely?

dasgarner:
OK - I think I need Alex's expertise on it :smile: 

I think `Dockerfile.dev` also needs a tweak so that it doesn't create the flag file - a minor point I know.

Dockerfile
LukyLuke: It's created some lines below. but probably a better solution would be to check for the config files itself and copy it if it does not exist.
I guess this was done so in the `entrypoint.sh` like this to not override existing developer configurations or so. If this is not needed, best solution would be to just copy the files without checking anything 😄

alexharrington:
I will review this properly on my return, but the issue I see is that the
permissions will be correct in the image, but not in the volume mounts, and
so unless people manually set correct permissions on the file system it
won't work as it stands.

alexharrington:
Will Apache start on port 80 under a non-root user? My understanding is that it requires root to bind to high-numbered ports, and then drops privilege to the user configured after that.

I appreciate that 80 external to the container can be directed to any port inside the container if Apache were running on a higher numbered port, however it would be a breaking change for any person not using the cms_custom-ports.yml compose file (and potentially for those using that but with a reverse proxy via an override yml file) which will be the majority of users.

For us, Docker is about making the process of setting up and maintaining the CMS as easy as possible, so I'd be really averse to introducing anything that causes a breaking change, or that means users have to manually change things outside the container (eg filesystem permissions) to have it work as expected.

I will go through the PR with some questions though as some of the work can probably be moved over to the build stage. In that case, perhaps you can move your changes in to an alternative Dockerfile (as we do for Dev) and then maintain that going forward (I can't do so as we don't have an environment to test in). You can then build that Dockerfile locally and push to your local repo for deployment.

alexharrington:
Another issue that has occurred reading through the code is that I'm going to be adding a new method to include a CA certificate in to the container (which requires running update-ca-certificates as root) so that the user can dynamically include a root CA (for instances where they're running the container behind a filtering proxy server that issues its own 'fake' certificates). I'm not sure it's possible to do that in your model unless they built the container locally.

LukyLuke:
I will ping you when it's working and open a new Pull-Request :o)

PR #411, opened 2018-09-04
Bugfix/1.8.12 pack1

PeterMis:


lib/Controller/User.php
```diff
@@ -748,6 +956,49 @@ public function changePassword()
...
+     */
+    public function forceChangePassword()
+    {
+        // Save the user
+        $user = $this->getUser();
+        $newPassword = $this->getSanitizer()->getString('newPassword');
+        $retypeNewPassword = $this->getSanitizer()->getString('retypeNewPassword');
+
+        if ($newPassword == null || $retypeNewPassword == '')
+            throw new \InvalidArgumentException(__('Please enter the password'));
```
dasgarner: Can you change your InvalidArgumentExceptions to the one that extends from XiboException?

lib/Controller/User.php
```diff
@@ -748,6 +956,49 @@ public function changePassword()
...
+
+        if ($newPassword != $retypeNewPassword)
+            throw new \InvalidArgumentException(__('Passwords do not match'));
```
dasgarner: As above

lib/Controller/User.php
```diff
@@ -740,6 +945,9 @@ public function changePassword()
         $newPassword = $this->getSanitizer()->getString('newPassword');
         $retypeNewPassword = $this->getSanitizer()->getString('retypeNewPassword');
 
+        if ($newPassword == null || $retypeNewPassword == '')
+            throw new \InvalidArgumentException(__('Please enter the password'));
```
dasgarner: As below if possible

lib/Controller/User.php
```diff
@@ -748,6 +956,49 @@ public function changePassword()
...
+            'data' => $user
+        ]);
+    }
+
+    /**
+     * Force User Password Change
+     */
+    public function forceChangePasswordPage()
+    {
+        $this->getState()->template = 'user-force-change-password-page';
```
dasgarner: we need something in here that redirects the user to the homepage if their change password flag isn't set to 1.

PR #416, opened 2018-09-20
Dockerfiles: Remove Unneeded mkdir

amayer5125:
The directory will be created when any of the other `mkdir` commands are called since  `-p` flag is used.


alexharrington:
You're correct it's unnecessary, but I think it aids readability, and since it happens within a single layer there's no penalty for keeping it. What's the problem this is causing you please?

amayer5125:
No problem. I was just looking over the project and noticed it was redundant.

PR #419, opened 2018-10-10
Update DataSetFactory.php

chalda-pnuzig:
Wrong usage of getDate function, date is always null.
See https://community.xibo.org.uk/t/remote-dataset-does-not-parse-date-fields/16099


PeterMis:
Fixed here - https://github.com/xibosignage/xibo/issues/1677

PR #431, opened 2018-11-27
Pull in release18 and update test suite for v2 changes.

dasgarner:


dasgarner:
All tests passing aside from one Cypress test - notified relevant person who will pull and get that passing.

PR #451, opened 2019-01-08
Feature/proof of play israt

ifarzana:


lib/Controller/Stats.php
```diff
@@ -45,6 +46,10 @@ class Stats extends Base
      * @var StorageServiceInterface
      */
     private $store;
+    /**
```
dasgarner: A carriage return would be tidier ;)

lib/Controller/Stats.php
```diff
@@ -263,128 +274,56 @@ public function grid()
...
-            $mediaSql = '';
-            $i = 0;
-            foreach ($mediaIds as $mediaId) {
-                $i++;
-                $mediaSql .= ':mediaId_' . $i . ',';
-                $params['mediaId_' . $i] = $mediaId;
+        /*Clear the display_ids array and add the user selected display in the array*/
+        if($displayId != null){
+            $display_ids = array();
+            if(!in_array($displayId, $display_ids)){
```
dasgarner: As we refactor old code we make a habit of changing the names of variables where they don't fit - `$display_ids` could be refactored to `$displayIds`

lib/Controller/Stats.php
```diff
@@ -524,62 +460,43 @@ public function export()
         $displayId = $this->getSanitizer()->getInt('displayId');
 
         // Get an array of display id this user has access to.
-        $displayIds = array();
+        $display_ids = array();
```
dasgarner: Same as above regarding naming convention

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        $this->store->isolated($sql, $data);
+
+    }
+
+    /** @inheritdoc */
+    public function addTagStat($statData)
+    {
+
```
dasgarner: Can we add a query for this too?

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        return ['statData'=> $rows];
+
+    }
+
+    /** @inheritdoc */
+    public function deleteStats($maxage, $fromDt = null, $options = [])
+    {
+        try {
```
dasgarner: Here I think we should specify some default options (considering `$options` is null). Something like:

```
$options = array_merge([
  'limit' => 10000,
   /*etc*/
], $options);```

File: lib/Storage/MySqlTimeSeriesStore.php [hunk 6]
```diff
@@ -0,0 +1,371 @@
...
+                // Break if we've exceeded the maximum attempts.
+                if ($i >= $options['maxAttempts'])
+                    break;
+            }
+
+            $this->log->debug('Deleted Stats back to ' . $maxage . ' in ' . $i . ' attempts');
+
+            $message = ' - ' . __('Done.');
+
+            return ['message' => $message];
```
dasgarner: Passing a message back would certainly work, but is it a good design pattern?

Would it be more "PHP" if it returned a count of records deleted or an exception on failure (we'd have to throw a general exception rather than a PDO one as other time series stores might not be PDO).

We could then have our message set by the calling method as before.

lib/Storage/TimeSeriesStoreInterface.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/*
```
dasgarner: We need to get Copyrights set in all new/modified files. The correct one to use is:

```
/**
 * Copyright (C) 2019 Xibo Signage Ltd
 *
 * Xibo - Digital Signage - http://www.xibo.org.uk
 *
 * This file is part of Xibo.
 *
 * Xibo is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * any later version.
 *
 * Xibo is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
 * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
 */
```

lib/Xmds/Soap.php
```diff
@@ -1447,40 +1464,52 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
...
+            } elseif ($type == 'layout') {
+                $layoutStats[] = [
+                    'type' => $type,
+                    'statDate' => $now,
+                    'fromDt' => $fromdt,
+                    'toDt' => $todt,
+                    'scheduleId' => $scheduleId,
+                    'displayId' => $this->display->displayId,
+                    'layoutId' => $layoutId,
+                ];
```
dasgarner: The other `if` statement needed here is called `event` and records tag data.

lib/Controller/Stats.php
dasgarner: So I am not sure if this logic is right...

Take the following example... there are 5 displays, id 1,2,3,4,5
The user has permissions for id 1,2,3 but not 4,5
The user filters on id 1: They should only get stats for id 1.
The user then filters on id 4: They should get 0 records.
The user then filters on id 1 and 4: They should only get stats for id 1

I think the above code always gives them stats for 1,2,3 ?

lib/Storage/MySqlTimeSeriesStore.php
ifarzana: We do not have tag stat records in MySQL. I need to pass tagStat array from doSubmitStats() of Soap.php. I can imagine the tagStat array as below: 

```
$tagStats[] = [
    'type' => $type,
    'statDate' => $now,
    'fromDt' => $fromdt,
    'toDt' => $todt,
    'scheduleId' => $scheduleId,
    'displayId' => $this->display->displayId,
    'tag' => $tag,
];
```

dasgarner: Yes I think that is sensible.

lib/Xmds/Soap.php
ifarzana: I added as below: when $type == 'event'
```
} elseif ($type == 'event') {
    $tagStats[] = [
        'type' => $type,
        'statDate' => $now,
        'fromDt' => $fromdt,
        'toDt' => $todt,
        'scheduleId' => $scheduleId,
        'displayId' => $this->display->displayId,
        'tag' => $tag,
    ];
}
```

lib/Storage/MySqlTimeSeriesStore.php
dasgarner: I am not sure what happens if you pass `null` into `array_merge` ?

PR #452, opened 2019-01-14
PHPUnit Tests for MySqlTimeSeriesStore

ifarzana:


tests/integration/StatisticsLayoutTest.php
```diff
@@ -0,0 +1,235 @@
...
+        $this->assertObjectHasAttribute('data', $object, $this->client->response->body());
+    }
+
+    /**
+     * Check if proof of play statistics are correct
+     */
+    public function testProof()
+    {
+        // Create a Display
+        $display = $this->createDisplay();
```
dasgarner: We are creating a specific display and layout in this test - could we have these items in `setup` instead? and then in `teardown` could we just call delete on these objects?

If we only add specific items in setup and then delete them in teardown, i think we should be able to remove all of the other setup/teardown code?

tests/integration/StatisticsLayoutTest.php
```diff
@@ -0,0 +1,235 @@
...
+    protected $startLayouts;
+    protected $startDisplays;
+
+    /**
+     * setUp - called before every test automatically
+     */
+    public function setup()
+    {
+        parent::setup();
+        $this->startMedias = (new XiboLibrary($this->getEntityProvider()))->get(['start' => 0, 'length' => 10000]);
```
dasgarner: this records the "state of the system" - query if we need that for this test - see below comments.

tests/integration/StatisticsLayoutTest.php
```diff
@@ -0,0 +1,235 @@
...
+        $this->startDisplays = (new XiboDisplay($this->getEntityProvider()))->get(['start' => 0, 'length' => 10000]);
+    }
+
+    /**
+     * tearDown - called after every test automatically
+     */
+    public function tearDown()
+    {
+        // tearDown all media files that weren't there initially
+        $finalMedias = (new XiboLibrary($this->getEntityProvider()))->get(['start' => 0, 'length' => 10000]);
```
dasgarner: this tries to delete any differences between the end of test state and the start of test state - query if we need this - see below comments.

dasgarner:
test this please

PR #454, opened 2019-01-21
Feature/player upgrades

PeterMis:
Improvements to the Player Application Upgrades


db/migrations/20190121092556_player_upgrade_and_override_config_migration.php
```diff
@@ -0,0 +1,76 @@
...
+
+        // add overrideConfig column to display table
+        $overrideConfigColumn = $displayTable->hasColumn('overrideConfig');
+        if (!$overrideConfigColumn)
+            $displayTable->addColumn('overrideConfig', 'text');
+
+        $displayTable->save();
+
+        // Get system user
+	    $user = $this->fetchRow("SELECT userId FROM `user` WHERE userTypeId = 1");
```
dasgarner: indentation

lib/Controller/Display.php
```diff
@@ -500,7 +505,10 @@ function grid()
             /* @var \Xibo\Entity\Display $display */
             if (in_array('displaygroups', $embed)) {
                 $display->load();
-            } else {
+            } elseif (in_array('overrideconfig', $embed)) {
```
dasgarner: shouldn't this be a new IF statement after this one - what if you want to include displaygroups and overrideconfig?

lib/Controller/Display.php
```diff
@@ -794,6 +796,23 @@ function editForm($displayId)
             $layouts = [];
         }
 
+        $playerVersions  = $this->playerVersionFactory->query(null, ['playerType' => $display->clientType]);
+        $versions = [];
+
+        // Go through the player versions from factory and supply it as a parameter versions to edit form - on the form we default to value set previously on the profile or to the newest Code version if it's a new display profile.
+        foreach ($playerVersions as $playerVersion) {
```
dasgarner: opportunity to refactor this into a method on the display? `$display->getSetting('versionMediaId', overrideOnly:true)`

lib/Controller/Display.php
```diff
@@ -1025,6 +1047,44 @@ function edit($displayId)
         $display->timeZone = $this->getSanitizer()->getString('timeZone');
         $display->displayProfileId = $this->getSanitizer()->getInt('displayProfileId');
 
+        $currentProfile = $display->getSettings();
```
dasgarner: Not clear why we copy all current settings into the override config here? shouldn't we only do the ones that are set (which at the moment can only be the versionMediaId) ?

lib/Controller/DisplayProfile.php
```diff
@@ -305,10 +310,18 @@ public function editForm($displayProfileId)
     {
         // Create a form out of the config object.
         $displayProfile = $this->displayProfileFactory->getById($displayProfileId);
+        // Get the Player Version for this display profile type
+        $playerVersions  = $this->playerVersionFactory->query(null, ['playerType' => $displayProfile->type]);
```
dasgarner: I thought our edit form used a select2 to look up this data (calling GET /playersoftware) ?

lib/Controller/PlayerSoftware.php
```diff
@@ -0,0 +1,418 @@
...
+        if (!$this->getUser()->checkDeleteable($media))
+            throw new AccessDeniedException();
+
+        $version->load();
+        $media->load();
+
+        // Delete
+        $version->delete();
+        $media->setChildObjectDependencies($this->layoutFactory, $this->widgetFactory, $this->displayGroupFactory, $this->displayFactory, $this->scheduleFactory, $this->playerVersionFactory);
+        $media->delete();
```
dasgarner: you probably want to delete the media record before you delete the version record, because mediaId is a foreign key on version

lib/Controller/PlayerSoftware.php
```diff
@@ -0,0 +1,418 @@
...
+        echo $ssspDocument->saveXML();
+    }
+
+    /**
+     * @param string $cmsKey
+     * @param int $displayId
+     * @throws NotFoundException
+     * @throws InvalidArgumentException
+     */
+    public function getVersionFile($cmsKey, $displayId)
```
dasgarner: new lines are available completely free of charge :-)

lib/Entity/Display.php
```diff
@@ -319,6 +319,12 @@ class Display implements \JsonSerializable
      */
     public $tags;
 
+    /**
+     * @SWG\Property(description="Media Id of the installer file overridden per display")
+     * @var int
+     */
+    public $versionMediaId = null;
```
dasgarner: this isn't needed anymore is it? data contained in `overrideConfig` ?

lib/Entity/Display.php
```diff
@@ -820,8 +853,18 @@ private function setConfig()
 
                 $displayProfile = $this->displayProfileFactory->getUnknownProfile($this->clientType);
             }
+            if (isset($this->overrideConfig)) {
```
dasgarner: I see what you're trying to do here, but I think the logic needs to be after you've loaded the display profile settings - something like:

```
// load my config

// load in the display profile config

// if I have been asked to in options, merge my display overrides with the config i have from the display profile

File: lib/Factory/MediaFactory.php [hunk 10]
```diff
@@ -536,6 +536,10 @@ public function query($sortOrder = null, $filterBy = [])
 
         $body .= " WHERE 1 = 1 ";
 
+        if ($this->getSanitizer()->getString('notPlayerSoftware', $filterBy) == 1) {
```
dasgarner: getString?

lib/Factory/ModuleFactory.php
```diff
@@ -578,6 +578,10 @@ public function query($sortOrder = null, $filterBy = array())
             $params['regionSpecific'] = $this->getSanitizer()->getInt('regionSpecific', $filterBy);
         }
 
+        if ($this->getSanitizer()->getString('notPlayerSoftware', $filterBy) == 1) {
```
dasgarner: getString?

lib/Factory/PlayerVersionFactory.php
```diff
@@ -0,0 +1,290 @@
...
+        $sql = '
+        SELECT DISTINCT player_software.player_type AS type 
+        FROM player_software
+        ORDER BY type ASC
+        ';
+
+        foreach ($this->getStore()->select($sql, $params) as $row) {
+            $entries[] = $version = $this->createEmpty()->hydrate($row);
+        }
+        $this->getLog()->debug('PLAYER DISTINCT TYPES ENTRIES ARE ' . json_encode($entries, JSON_PRETTY_PRINT));
```
dasgarner: tidy log?

lib/Widget/PlayerSoftware.php
```diff
@@ -0,0 +1,122 @@
...
+     * @return PlayerVersionFactory
+     */
+    private function getPlayerVersionFactory()
+    {
+        return $this->getApp()->container->get('playerVersionFactory');
+    }
+
+    public function postProcess($media)
+    {
+        $this->getLog()->debug('MEDIA IN POST PROCESS IS ' . json_encode($media, JSON_PRETTY_PRINT));
```
dasgarner: tidy logs?

lib/Xmds/Soap.php
```diff
@@ -423,6 +429,22 @@ protected function doRequiredFiles($serverKey, $hardwareKey, $httpDownloads)
 
         // Create a comma separated list to pass into the query which gets file nodes
         $layoutIdList = implode(',', $layouts);
+        if (isset($this->display->overrideConfig))
```
dasgarner: refactor - should all be replace-able with `$display->getSetting('versionMediaId');` ?

lib/Xmds/Soap5.php
```diff
@@ -95,11 +96,14 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
                 $displayElement->setAttribute('status', 0);
                 $displayElement->setAttribute('code', 'READY');
                 $displayElement->setAttribute('message', 'Display is active and ready to start.');
-                $displayElement->setAttribute('version_instructions', $display->versionInstructions);
 
                 // Display Settings
-                $settings = $display->getSettings();
-
+                if (isset($this->display->overrideConfig))
```
dasgarner: refactor - should be just $this->display->getSettings();`

we always want to consider overrides here?

lib/Xmds/Soap5.php
```diff
@@ -112,11 +116,29 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
                         $arrayItem['value'] = $this->getConfig()->getSetting('XMR_PUB_ADDRESS');
                     }
 
+                    if (strtolower($arrayItem['name']) == 'versionmediaid' && $arrayItem['value'] != null ) {
```
dasgarner: refactor - $display->getSetting('versionMediaId') should be enough

lib/Controller/DisplayProfile.php
PeterMis: We do indeed, but I think I need to query it here (and similar in Display controller) and pass versions to the twig as a "field.options" and  with the field.value in select 2 dropdown, to have correct default / previously selected version correctly displayed there?

dasgarner: You only need to add 1 item to it, which is the currently selected one (if there is one). So if you have a getByMediaId in your playerVersionFactory you could use that, along with the `$display->getSetting('versionMediaId', <only get the overrides>);`

Imagine that there are 100 versions - the above would load all of those into a list, and never show them (because as soon as you open the list, the query fires to get the paged version).

PR #457, opened 2019-01-22
Misc issues before beta code freeze

dasgarner:


dasgarner:
Failed tests relate to removing items from the timeline form request, replacing them with a call to `/playlist`. This will be done in another branch.

PR #458, opened 2019-01-23
Feature/player upgrades18

PeterMis:


lib/Factory/ModuleFactory.php
```diff
@@ -576,6 +576,10 @@ public function query($sortOrder = null, $filterBy = array())
                 $params['regionSpecific'] = $this->getSanitizer()->getInt('regionSpecific', $filterBy);
             }
 
+            if ($this->getSanitizer()->getInt('notPlayerSoftware', $filterBy) == 1) {
```
dasgarner: wouldn't it be better to filter on type here instead of name?

```
AND media.type <> \'playersoftware\'
```

lib/Controller/Library.php
```diff
@@ -309,8 +324,9 @@ function displayPage()
         $this->getState()->template = 'library-page';
         $this->getState()->setData([
             'users' => $this->userFactory->query(),
-            'modules' => $this->moduleFactory->query(['module'], ['regionSpecific' => 0, 'enabled' => 1]),
-            'groups' => $this->userGroupFactory->query()
+            'modules' => $this->moduleFactory->query(['module'], ['regionSpecific' => 0, 'enabled' => 1, 'notPlayerSoftware' => 1]),
+            'groups' => $this->userGroupFactory->query(),
+            'validExt' => implode('|', $this->moduleFactory->getValidExtensions(['notPlayerSoftware' => 1]))
```
dasgarner: Why the change to a `|` ?

PR #460, opened 2019-01-23
Swift 2.0.0 beta

PeterMis:
Minor fixes for player software


lib/Factory/ModuleFactory.php
```diff
@@ -579,7 +579,7 @@ public function query($sortOrder = null, $filterBy = array())
         }
 
         if ($this->getSanitizer()->getInt('notPlayerSoftware', $filterBy) == 1) {
-            $body .= ' AND name <> "Player Software" ';
+            $body .= ' AND module <> "playersoftware" ';
```
dasgarner: these need to be `'` and not `"` - MySQL does support `"` but it is not considered good practice to use them

PR #472, opened 2019-01-31
Rename mediamanager to mediadashboard

PeterMis:


lib/routes-web.php
```diff
@@ -40,8 +40,8 @@
 $app->get('/statusdashboard', '\Xibo\Controller\StatusDashboard:displayPage')->name('statusdashboard.view');
 $app->get('/statusdashboard/displays', '\Xibo\Controller\StatusDashboard:displays')->name('statusdashboard.displays');
 $app->get('/icondashboard', '\Xibo\Controller\IconDashboard:displayPage')->name('icondashboard.view');
-$app->get('/mediadashboard', '\Xibo\Controller\MediaManager:displayPage')->name('mediamanager.view');
```
dasgarner: While I agree that this would be the correct approach to fixing the issue - we can't put in a DB change without releasing 1.8.13.

If we changed these lines to:

```
$app->get('/mediamanager', '\Xibo\Controller\MediaManager:displayPage')->name('mediamanager.view');
$app->get('/mediamanager/data', '\Xibo\Controller\MediaManager:grid')->name('mediaManager.search');
```

then I believe we could fix it without having any of the associated changes or DB mods?

PR #476, opened 2019-02-10
RemoteDatasets always truncates

LukyLuke:
RemoteDatasets has been cleared all the time because the clearRate Property is a string and not an int

### DIFF
--- a/lib/Entity/DataSet.php
+++ b/lib/Entity/DataSet.php
@@ -573,7 +573,7 @@ class DataSet implements \JsonSerializable
      */
     public function isTruncateEnabled()
     {
-        return $this->clearRate !== 0;
+        return intval($this->clearRate) !== 0;
     }
 
     /**


dasgarner:
Thanks Luke - we should fix this slightly further upstream by ensuring that the factory parses out this properly as an int (which will mean it gets fixed elsewhere too).

See https://github.com/xibosignage/xibo-cms/blob/develop/lib/Factory/DataSetFactory.php#L284

Happy to do this - or if you'd like to modify your PR?

Thanks (and a happy new year btw!).

LukyLuke:
Hey dasgarner,
happy new year to you guys too!

That's right, it's a better place there.
There is no need for change the PullRequest if you can fix this in direct commit - that's even better :)

Thanks for you work!

dasgarner:
Thanks Luke - we've committed something into 1.8/2.0 for it.

PR #485, opened 2019-02-19
Swift 2.0.0 beta3

PeterMis:


lib/Entity/Playlist.php
```diff
@@ -518,6 +518,10 @@ public function delete($options = [])
             }
         }
 
+        // region delete, the parentId is the regionPlaylist playlistId, we want to remove all linked subplaylists in that regionPlaylist
```
dasgarner: can we expand this comment so that we say why it is safe to do this here - something like:

> we are deleting a region so any sub-playlists attached to it will be deleted - we want to remove all link records from the closure table using the parentId as that is the ID of the region specific playlist.

dasgarner: comment suggested because it took me a long while to work out whether this was safe or not

PR #487, opened 2019-02-20
Feature/2.0.0 beta.3 pack1

dasgarner:


lib/Entity/Schedule.php
```diff
@@ -151,6 +164,12 @@ class Schedule implements \JsonSerializable
      */
     public $recurrenceRepeatsOn;
 
+    /**
+     * @SWG\Property(description="Recurrence monthly repeats on - 0 is day of month, 1 is weekday of week")
+     * @var string
```
PeterMis: that should be int?

dasgarner: yes it should, well spotted!

PR #492, opened 2019-02-26
OSX-428 - Bulk updating Layout and Media

ifarzana:


lib/Controller/Layout.php
```diff
@@ -726,6 +733,83 @@ function unretire($layoutId)
...
+
+        $enableStat = $this->getSanitizer()->getCheckbox('enableStat');
+
+        $layout->enableStat = $enableStat;
+        $layout->save();
+
+        // Return
+        $this->getState()->hydrate([
+            'httpStatus' => 204,
+            'message' => sprintf(__('For Layout %s Enable Stats Collection is set to %s'), $layout->layout, ($layout->enableStat == 1) ? 'On': 'Off')
```
dasgarner: Please can we translate On and Off strings too?

lib/Controller/Library.php
```diff
@@ -330,6 +330,84 @@ function displayPage()
...
+
+        $enableStat = $this->getSanitizer()->getString('enableStat');
+
+        $media->enableStat = $enableStat;
+        $media->save();
+
+        // Return
+        $this->getState()->hydrate([
+            'httpStatus' => 204,
+            'message' => sprintf(__('For Media %s Enable Stats Collection is set to %s'), $media->name, $media->enableStat)
```
dasgarner: Did you want to have your On/Off string here too (it might be nice instead of showing 0 / 1)

views/library-page.twig
```diff
@@ -428,5 +428,21 @@
             });
             $(dialog).find('.modal-body').append($input);
         }
+
+        function setEnableStatMultiSelectFormOpen(dialog) {
+
+            var $select = $('<select id="enableStat" name="enableStat" class="form-control">' +
+                '<option value="Off">Off</option>' +
```
dasgarner: Can we translate on/off/inherit here too?

lib/Controller/Library.php
ifarzana: This is already a string i.e., on/off/inherit

dasgarner: `$media->enableStat` is an integer?

ifarzana: $media->enableStat is a string

lib/Controller/Layout.php
ifarzana: `sprintf(__('For Layout %s Enable Stats Collection is set to %s'), $layout->layout, ($layout->enableStat == 1) ? __('On') : __('Off'))` updated

lib/Controller/Library.php
ifarzana: as it is a string, I have done as below:

> sprintf(__('For Media %s Enable Stats Collection is set to %s'), $media->name, __($media->enableStat))

dasgarner: ah I understand! yes that is perfect

db/migrations/20190213160914_add_global_stat_setting_migration.php
```diff
@@ -21,6 +21,8 @@
  */
 
 use Phinx\Migration\AbstractMigration;
+use Jenssegers\Date\Date;
```
dasgarner: I am not sure if phinx will be able to use this reference?

dasgarner: Phinx isn't loaded like one of our normal endpoints, and I don't know if it loads the vendor files or not (I think perhaps not?)

You might need to use PHP core DateTime libs?

lib/Controller/Stats.php
```diff
@@ -272,20 +272,29 @@ public function grid()
...
 
-        // Get an array of display id this user has access to.
+        // Do not filter by display if super admin and no display is selected
+        // Super admin will be able to see stat records of deleted display, we will not filter by display later
         $displayIds = [];
-        foreach ($this->displayFactory->query() as $display) {
-            $displayIds[] = $display->displayId;
-        }
+        if ( !$this->getUser()->isSuperAdmin() )
+        {
```
dasgarner: { should be on the same line as the if ;-)

lib/Controller/Stats.php
```diff
@@ -319,11 +328,13 @@ public function grid()
...
             // If the media name is empty, and the widgetid is not, then we can assume it has been deleted.
             $widgetName = ($widgetName == '' &&  $widgetId != 0) ? __('Deleted from Layout') : $widgetName;
+            $displayName = $this->getSanitizer()->string($row['display']);
+            $layoutName = $this->getSanitizer()->string($row['layout']);
 
             $entry['type'] = $this->getSanitizer()->string($row['type']);
             $entry['displayId'] = $this->getSanitizer()->int(($row['displayId']));
-            $entry['display'] = $this->getSanitizer()->string($row['display']);
-            $entry['layout'] = $this->getSanitizer()->string($row['layout']);
+            $entry['display'] = ($displayName != "") ? $displayName : __('Not Found');
```
dasgarner: `''` please :-)

PR #493, opened 2019-02-27
Dockerfile for Openshift / Non-Root environments

LukyLuke:
Next try for a special Image for Openshift.

The entrypoint for Openshift `openshift.sh` should be a replacement for the existing entrypoint.sh (see last commit 1e69fbd )

What I'm not happy about at the moment is, that it is one more Dockerfile with just some small changes. But as far as I know it's not possible to "import" a dockerfile into an other, or is it?

What I can try is to use the default xibo-image and just change all the things downstream what is needed specially for nonroot-environments.

Btw: Actually this is for the "development" branch because in our setup I need some fixes already :-)
And the new Layout-Editor you've done is amazing compared to the previous 👍


dasgarner:
Closing this out - if you ever revisit please do feel free to open another! Thanks

PR #494, opened 2019-02-28
Swift 2.0.0 beta3

PeterMis:


lib/Entity/Module.php
```diff
@@ -155,6 +155,10 @@ public function validate()
     {
         if (!v::intType()->validate($this->defaultDuration))
             throw new InvalidArgumentException(__('Default Duration is a required field.'), 'defaultDuration');
+
+        if (!v::alnum(',')->validate($this->validExtensions)) {
```
dasgarner: Does this work if its empty?

lib/Factory/DisplayFactory.php
```diff
@@ -120,6 +120,7 @@ public function getById($displayId, $showTags = false)
     public function getByLicence($licence)
     {
         $displays = $this->query(null, ['disableUserCheck' => 1, 'license' => $licence]);
+        $this->getLog()->debug('DISPLAY LOG LICENCE ' . $licence );
```
dasgarner: if we need to leave this in, can it be reworded and not capitals?

PeterMis: we don't need it, there is a chance I forgot to remove it.

lib/Entity/Module.php
PeterMis: no, it does not...

Do we want something like that? 🤔 

```
        if (v::notEmpty()->validate($this->validExtensions)) {
            if (!v::alnum(',')->validate($this->validExtensions)) {
                throw new InvalidArgumentException(__('Comma separated file extensions only please, without the .'),'validExtensions');
            }
        }
```

dasgarner:
```
 if (!empty($this->validExtensions) && !v::alnum(',')->validate($this->validExtensions)) {
            throw new InvalidArgumentException(__('Comma separated file extensions only please, without the .'), 'validExtensions');
        }
```

Did you mean that?

PR #496, opened 2019-03-04
Bugfix/1.8.13 pack2

PeterMis:


lib/Entity/DataSet.php
```diff
@@ -514,6 +515,11 @@ public function getData($filterBy = [], $options = [])
                             // Pull out the column name and date format
                             $details = explode(',', str_replace(')', '', str_replace('$dateFormat(', '', $column->formula)));
 
+                            if (isset($details[2])) {
+                                $language = str_replace(' ', '', $details[2]);
+                            }
+
+                            $this->date->setLocale($language);
```
dasgarner: does this set the locale to `""` if the language isn't specified?

lib/Entity/DataSet.php
```diff
@@ -504,6 +504,7 @@ public function getData($filterBy = [], $options = [])
 
         // Are there any client side formulas?
         if (count($clientSideFormula) > 0) {
+            $language = '';
```
dasgarner: any reason this is declared above? could it all be one statement inside the `if (isset($details[2])) {` check?

lib/Entity/DataSet.php
PeterMis: If it's not set in Formula, then yes, however in the end it would be set to en - as that's the default for the date translator.

Perhaps we should do something like that (covers both your comments) 🤔 

``` 
                            if (isset($details[2])) {
                                $language = str_replace(' ', '', $details[2]);
                            } else {
                                $language = $this->config->getSetting('DEFAULT_LANGUAGE', 'en_GB');
                            }
```
This seems to correctly translate to the language set in Formula and if it is not set there, then it translates to the CMS language.

dasgarner: that sounds good!

PR #513, opened 2019-03-19
Swift 2.1.0 alpha

PeterMis:


lib/Controller/DisplayGroup.php
```diff
@@ -1736,4 +1749,181 @@ public function command($displayGroupId)
...
+
+        // get an array of assigned media files
+        /** @var Media[] $assignedFiles */
+        $assignedFiles = $this->mediaFactory->getByDisplayGroupId($displayGroupId);
+
+        $copyMembers = $this->getSanitizer()->getCheckbox('copyMembers', 0);
+        $copyTags = $this->getSanitizer()->getCheckbox('copyTags', 0);
+        $copyAssignments = $this->getSanitizer()->getCheckbox('copyAssignments', 0);
+
+        if (!$this->getUser()->checkEditable($displayGroup)) {
```
dasgarner: We could make this more efficient if we did the permissions check before we queried for the rest of the data

lib/Controller/DisplayGroup.php
```diff
@@ -1736,4 +1749,181 @@ public function command($displayGroupId)
...
+     * )
+     */
+    public function copy($displayGroupId)
+    {
+        // get display group object
+        /** @var \Xibo\Entity\DisplayGroup $displayGroup */
+        $displayGroup = $this->displayGroupFactory->getById($displayGroupId);
+
+        // get an array of assigned displays
+        /** @var Display[] $membersDisplays */
```
dasgarner: These type hints shouldn't be necessary if the Factory has the correct `@returns` docblock on it

db/migrations/20190315134628_add_bandwidth_limit_column_to_displaygroup_migration.php
```diff
@@ -0,0 +1,38 @@
...
+class AddBandwidthLimitColumnToDisplaygroupMigration extends AbstractMigration
+{
+    /** @inheritdoc */
+    public function change()
+    {
+        $displayGroupTable = $this->table('displaygroup');
+
+        if (!$displayGroupTable->hasColumn('bandwidthLimit')) {
+            $displayGroupTable
+                ->addColumn('bandwidthLimit', 'integer', ['default' => 0, 'null' => true])
```
dasgarner: are we sure that we need the field nullable if the default value is 0? if so, do we handle a mixture of nulls and 0's in this field when we query for it?

lib/Entity/DisplayGroup.php
```diff
@@ -768,6 +780,7 @@ private function edit()
             'isDynamic' => $this->isDynamic,
             'dynamicCriteria' => $this->dynamicCriteria,
             'dynamicCriteriaTags' => $this->dynamicCriteriaTags,
+            'bandwidthLimit' => $this->bandwidthLimit,
```
dasgarner: depending on the answer to the migration question, we might need to protect this from null's

lib/Helper/XiboUploadHandler.php
```diff
@@ -50,17 +50,30 @@ protected function handle_form_data($file, $index)
 
             // Get some parameters
             if ($index === null) {
-                if (isset($_REQUEST['name']))
+                if (isset($_REQUEST['name'])) {
```
dasgarner: great!

lib/Xmds/Soap.php
```diff
@@ -1846,19 +1846,24 @@ protected function getIp()
      * Check we haven't exceeded the bandwidth limits
      *  - Note, display logging doesn't work in here, this is CMS level logging
      *
+     * @param int $displayId The Display ID
      * @return bool true if the check passes, false if it fails
+     * @throws NotFoundException
      */
-    protected function checkBandwidth()
+    protected function checkBandwidth($displayId = 0)
```
dasgarner: Do we ever need to call this without providing a display?

lib/Xmds/Soap.php
```diff
@@ -1846,19 +1846,24 @@ protected function getIp()
...
     {
         // Uncomment to enable auditing.
         //$this->logProcessor->setDisplay(0, true);
 
+        $display = $this->displayFactory->getById($displayId);
```
dasgarner: Can we use `$this->display` ?

views/authed.twig
```diff
@@ -241,10 +241,16 @@
             <td>
                 <span class="preview"></span>
             </td>
-            <td class="title">
+            <td class="title" title="Name">
```
dasgarner: Translations are probably needed here, somehow...

views/display-form-edit.twig
```diff
@@ -187,6 +187,23 @@
                         {% endif %}
                         {{ forms.datetime("auditingUntil", title, auditing, helpText) }}
 
+                        {% set title %}{% trans "Bandwidth limit" %}{% endset %}
+                        {% set helpText %}{% trans "The bandwdith limit that should be applied. Enter 0 for no limit." %}{% endset %}
```
dasgarner: bandwdith

views/authed.twig
PeterMis: I'm not sure how to do that there 🤔

PeterMis:
Retest this please

PR #520, opened 2019-04-02
OSX-456 Layout ID is changed in Stat table

ifarzana:


db/migrations/20190326163016_create_layout_history_table_migration.php
```diff
@@ -0,0 +1,56 @@
...
+    public function change()
+    {
+        $table = $this->table('layouthistory', ['id' => 'layoutHistoryId']);
+        $table->addColumn('campaignId', 'integer')
+            ->addColumn('layoutId', 'integer')
+            ->addColumn('publishedDate', 'datetime')
+            ->addForeignKey('campaignId', 'campaign', 'campaignId')
+            ->create();
+
+        $this->execute('INSERT INTO `layouthistory` (campaignId, layoutId, publishedDate)
```
dasgarner: Can we make a comment explaining what this is doing and why?

lib/Entity/Layout.php
```diff
@@ -680,9 +680,7 @@ public function delete($options = [])
             } catch (NotFoundException $notFoundException) {
                 $this->getLog()->info('No draft to delete for a Layout in the Draft state, odd!');
             }
-        }
-
-        // Delete Permissions
+        }// Delete Permissions
```
dasgarner: is this a whoops?

lib/Factory/LayoutFactory.php
```diff
@@ -288,6 +289,65 @@ public function getById($layoutId)
...
+     * Get CampaignId from layout history
+     * @param int $layoutId
+     * @return int campaignId|null
+     * @throws NotFoundException
+     */
+    public function getCampaignId($layoutId)
+    {
+        $campaignId = null;
+
+        if ($layoutId == 0)
```
dasgarner: can we have long form if statements with `{}` please

lib/Factory/LayoutFactory.php
```diff
@@ -288,6 +289,65 @@ public function getById($layoutId)
...
+     * @throws NotFoundException
+     */
+    public function getLatestLayoutId($campaignId)
+    {
+
+        if ($campaignId == null)
+            throw new NotFoundException();
+
+        try {
+            $row = $this->getStore()->select('SELECT MAX(layoutId) AS layoutId FROM `layouthistory` WHERE campaignId = :campaignId LIMIT 1', ['campaignId' => $campaignId]);
```
dasgarner: You don't need a LIMIT on a MAX without grouping (there will always be 1 record).

lib/Factory/LayoutFactory.php
```diff
@@ -288,6 +289,65 @@ public function getById($layoutId)
         return $layouts[0];
     }
 
+    /**
+     * Get CampaignId from layout history
+     * @param int $layoutId
+     * @return int campaignId|null
+     * @throws NotFoundException
+     */
+    public function getCampaignId($layoutId)
```
dasgarner: Maybe we should call this `getCampaignIdFromLayoutHistory()` just so it is clear where it goes for that information?

lib/Factory/LayoutFactory.php
```diff
@@ -288,6 +289,65 @@ public function getById($layoutId)
...
+        return $campaignId;
+    }
+
+    /**
+     * Get latest layoutId by CampaignId from layout history
+     * @param int campaignId
+     * @return int layoutId
+     * @throws NotFoundException
+     */
+    public function getLatestLayoutId($campaignId)
```
dasgarner: Maybe we should call this `getLatestLayoutIdFromLayoutHistory()` just so it is clear where it goes for that information?

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -126,7 +136,7 @@ public function getStatsReport($fromDt, $toDt, $displayIds, $layoutIds, $mediaId
 
         $body .= ' WHERE stat.type <> \'displaydown\'
                 AND stat.end > :fromDt
-                AND stat.start <= :toDt
+                AND stat.start <= DATE_ADD(:toDt, INTERVAL 1 DAY)
```
dasgarner: I'm not clear why we are adding a day to the toDt here?

If we do need to do that, then should we do that in PHP rather than MySQL?

ifarzana: We don't see any stat records of to date in the grid if we don't add a date. I will change it to PHP date.

lib/Entity/Layout.php
ifarzana: o_o

ifarzana:
retest this please

ifarzana:
retest this please

ifarzana:
retest this please

PR #525, opened 2019-04-03
Html preview CSS override enable for white label

maurofmferrao:


web/theme/default/css/html-preview.css
```diff
@@ -0,0 +1,21 @@
+/*
```
dasgarner: Can we have the new standard copyright?

PR #532, opened 2019-04-08
OSX-443 Time Disconnected Stats are inaccurate

ifarzana:


views/stats-proofofplay-page.twig
```diff
@@ -193,7 +193,31 @@
                             if (type != "display")
                                 return "";
 
-                            return moment().startOf("day").seconds(data).format("H:mm:ss")
+                            let durationData = moment.duration(data, "seconds");
```
dasgarner: We need to think about how this gets translated into different languages (the works months, day, etc)

dasgarner:
https://github.com/xibosignage/xibo/issues/1743

PR #538, opened 2019-04-10
Feature: Can search layouts filtered by ParentId

CodeIntern:
Hello,

One of our needs was to be able to find, through the API, the "draft" Layout given the original "published" Layout's id.

This change allows a user to pass a `parentId` parameter to the `api/layout` endpoint to filter Layouts whose parent Layout has the given id.


dasgarner:
Jenkins: test this please

dasgarner:
@PeterMis - would you mind looking over this when you have time and give your review. Thanks.

dasgarner:
Ok, we will have to Cherry Pick into release20 if we think that is needed, otherwise I will merge Monday. Tha is.

PR #548, opened 2019-04-27
Maint: Merge release20 into develop

dasgarner:


dasgarner:
> Only one comment, perhaps we should include my latest merged PR to release20 here as well?

Yes we will definitely need to merge again before 2.0.2

PR #561, opened 2019-05-09
OSX-516 Auto-report data model and logic

ifarzana:


db/migrations/20190509101525_create_report_schedule_table_migration.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @inheritdoc */
+    public function change()
+    {
+        $table = $this->table('reportschedule', ['id' => 'reportScheduleId']);
+        $table->addColumn('name', 'string')
+            ->addColumn('reportName', 'string')
+            ->addColumn('filterCriteria', 'text')
+            ->addColumn('schedule', 'string')
+            ->addColumn('lastRunDt', 'integer', ['default' => 0])
+            ->addColumn('userId', 'integer')
```
dasgarner: A foreign key here to the user table would be good.

db/migrations/20190509101525_create_report_schedule_table_migration.php
```diff
@@ -0,0 +1,44 @@
...
+            ->addColumn('createdDt', 'integer')
+            ->create();
```
dasgarner: if we use `save()` here, then I think we can make use of rollback in some circumstances

db/migrations/20190509102648_create_saved_report_table_migration.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Class CreateSavedReportTableMigration
+ */
+class CreateSavedReportTableMigration extends AbstractMigration
+{
+    /** @inheritdoc */
+    public function change()
+    {
+        $table = $this->table('saved_report', ['id' => 'savedReportId']);
```
dasgarner: We have a few foreign keys we could add here, mediaId, reportScheduleId and userId

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+            $cron = \Cron\CronExpression::factory($reportSchedule->schedule);
+            $nextRunDt = $cron->getNextRunDate(\DateTime::createFromFormat('U', $reportSchedule->lastRunDt))->format('U');
+            $reportSchedule->nextRunDt = $nextRunDt;
+
+            $reportSchedule->reportName = $this->reportService->getReportByName($reportSchedule->reportName)->description;
+
+            switch ($reportSchedule->schedule) {
+
+                case ReportSchedule::$SCHEDULE_DAILY:
+                    $reportSchedule->schedule = 'Run once a day, midnight';
```
dasgarner: We should probably make these translatable strings.

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+        $this->getState()->setData($reportSchedules);
+    }
+
+    /**
+     * Report Schedule Add
+     */
+    public function reportScheduleAdd()
+    {
+
+        $this->getLog()->debug('Add Report Schedule');
```
dasgarner: Should some of this logic be in the report class? Not all reports will have a type/layout for example?

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+        ]);
+    }
+
+    /**
+     * Displays an Add form
+     */
+    public function addReportScheduleForm()
+    {
+
+        $type = $this->getSanitizer()->getParam('type', 'layout');
```
dasgarner: Are these things we might expect the report class to deal with?

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+     * Report Schedule Delete Form
+     */
+    public function deleteReportScheduleForm($reportScheduleId)
+    {
+        $reportSchedule = $this->reportScheduleFactory->getById($reportScheduleId);
+
+        if (!$this->getUser()->checkDeleteable($reportSchedule))
+            throw new AccessDeniedException(__('You do not have permissions to delete this report schedule'));
+
+//         var_dump($reportSchedule->name);
```
dasgarner: we can take these out of the final PR

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+    //</editor-fold>
+
+    //<editor-fold desc="Saved report">
+
+    /**
+     * Saved report Grid
+     */
+    public function savedReportGrid()
+    {
+        $reportName = 'summaryReport';
```
dasgarner: What is this used for?

lib/Controller/Report.php
```diff
@@ -0,0 +1,520 @@
...
+
+    /**
+     * Displays the Saved Report Page
+     */
+    public function displaySavedReportPage()
+    {
+        // Call to render the template
+        $this->getState()->template = 'saved-report-page';
+        $this->getState()->setData([
+            'savedReports' => $this->savedReportFactory->query()
```
dasgarner: it would be better to have this as a separate `grid` type call, so that we can take advantage of paging, etc. If there are 10000 saved reports we wont be able to load them like this.

lib/Factory/ReportScheduleFactory.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        // Sorting?
+        $sql .= 'ORDER BY ' . implode(',', $sortOrder);
+
+        foreach ($this->getStore()->select($sql, $params) as $row) {
+            $entries[] = $this->createEmpty()->hydrate($row);
+        }
+
+        return $entries;
```
dasgarner: We should add our standard paging logic here in case there are a lot of schedules to query.

lib/Report/ReportTrait.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * @var SanitizerServiceInterface
+     */
+    private $sanitizerService;
+
+    /**
+     * @var DisplayFactory
+     */
+    private $displayFactory;
```
dasgarner: I might be wrong, but it feels like these factories are specific to the report we are dealing with and not all reports will need them (and might need different ones).

This might be a good use case for a service locator factory where we have a `setDependencies($container)` method which takes the DI container as a parameter - each individual report class can then grab the factories it needs.

dasgarner: This comment refers to the layout, media and display factories which would seem to be specific to the report we've made

lib/Report/SummaryReport.php
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /** @inheritdoc */
+    public function getSavedReportResults($savedreportId)
+    {
+
+        $savedReport = $this->savedReportFactory->getById($savedreportId);
+
+        // Open a zipfile and read the json
+        $zipFile = $this->getConfig()->getSetting('LIBRARY_LOCATION') . $savedReport->storedAs;
```
dasgarner: this bit of logic here probably belongs in the report service rather than the specific summary report - meaning we should enforce all reports save/retrieve their data in the same way

lib/Report/SummaryReport.php
```diff
@@ -0,0 +1,232 @@
...
+            throw new InvalidArgumentException(__('No displays with View permissions'), 'displays');
+
+        $labels = [];
+        $countData = [];
+        $durationData = [];
+        $backgroundColor = [];
+        $borderColor = [];
+
+        // Call the time series interface get daily summary report
+        $result =  $this->getTimeSeriesStore()->getDailySummaryReport($displayIds, $diffInDays, $type, $layoutId, $mediaId, $reportFilter, $groupByFilter, $fromDt, $toDt);
```
dasgarner: So this bit is interesting - we should discuss!
Having our clever abstract report classes doesn't help if we eventually have to put all the report methods in the time series stores.

lib/Service/ReportService.php
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * @var SavedReportFactory
+     */
+    private $savedReportFactory;
+
+    /**
+     * @var DisplayFactory
+     */
+    private $displayFactory;
```
dasgarner: Again it feels like these factories are specific to our report - not generic to the service - see prior comment

lib/Controller/Report.php
ifarzana: We need the reportName to be passed as parameter to preview a saved report. I have changed the hard coded reportName.

PR #563, opened 2019-05-10
OSX-149 Implement Two Factor Authentication xibosignage/xibo#1777

PeterMis:


lib/Controller/Login.php
```diff
@@ -407,4 +419,157 @@ private function generateEmailBody($subject, $body)
...
+
+        // if our user has email two factor enabled, we need to send the email with code now
+        if ($user->twoFactorTypeId === 1) {
+
+            if ($user->email == '') {
+                throw new NotFoundException('No email');
+            }
+
+            $mailFrom = $this->getConfig()->getSetting('mail_from');
+            $tfa = new TwoFactorAuth('Xibo Signage');
```
dasgarner: Should this be the value in global settings?

lib/Controller/Login.php
```diff
@@ -407,4 +419,157 @@ private function generateEmailBody($subject, $body)
...
+     */
+    public function twoFactorAuthValidate()
+    {
+        $user = $this->userFactory->getByName($_SESSION['tfaUsername']);
+        $result = false;
+        $updatedCodes = [];
+
+        if (isset($_POST['code'])) {
+            $tfa = new TwoFactorAuth('Xibo Signage');
+            $result = $tfa->verifyCode($user->twoFactorSecret, $this->getSanitizer()->string($_POST['code']));
```
dasgarner: any reason you don't use getSanitizer()->getString() here?

lib/Entity/User.php
```diff
@@ -1378,4 +1403,40 @@ public function getUserOptions()
...
+        $this->twoFactorTypeId = 0;
+        $this->twoFactorSecret = NULL;
+        $this->twoFactorRecoveryCodes = NULL;
+
+        $sql = 'UPDATE `user` SET twoFactorSecret = :twoFactorSecret,
+                  twoFactorTypeId = :twoFactorTypeId,
+                  twoFactorRecoveryCodes =:twoFactorRecoveryCodes
+               WHERE userId = :userId';
+
+        $params = array(
```
dasgarner: Old school array syntax

views/user-form-edit-profile.twig
```diff
@@ -0,0 +1,195 @@
...
+                    <p>{% trans "Here are your recovery codes, please make sure to store them in a safe place, like password manager." %}</p>
+                    <p>{% trans "Recovery codes will become active only after this form is successfully saved" %}</p>
+                    <p id="codesList" class=""></p>
+                </div>
+            </form>
+        </div>
+    </div>
+
+    <script type="text/javascript">
+        $(document).ready(function() {
```
dasgarner: I am quite surprised this works because this whole form is returned via AJAX. This should probably be a function in `authed.twig` which is then called on the form load callback?

PR #582, opened 2019-06-03
Timezone implementation

ifarzana:


dasgarner:
jenkins: retest this please

lib/Factory/ReportScheduleFactory.php
```diff
@@ -153,10 +153,10 @@ public function query($sortOrder = null, $filterBy = [])
             $body .= " AND reportschedule.reportScheduleId = :reportScheduleId ";
         }
 
-        // Owner filter
-        if ($this->getSanitizer()->getInt('userId', 0, $filterBy) != 0) {
-            $body .= " AND reportschedule.userid = :userId ";
-            $params['userId'] = $this->getSanitizer()->getInt('userId', 0, $filterBy);
+        if ( $this->getSanitizer()->getCheckbox('onlyMySchedules') == 1) {
```
dasgarner: is it useful to have both of these things?

db/migrations/20190603083836_change_stat_table_start_end_column_migration.php
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use Phinx\Migration\AbstractMigration;
+
+/**
+ * Class AddPreviousRunDateReportScheduleMigration
+ */
+class ChangeStatTableStartEndColumnMigration extends AbstractMigration
```
dasgarner: Q: how does this play into migrating older stats? we've assumed here that we will be dropping them all, but we know in reality that we can't do that.

PR #590, opened 2019-06-18
Brooks 2.1.0 beta

PeterMis:


PeterMis:
Retest this please

lib/Entity/Tag.php
```diff
@@ -542,4 +735,101 @@ private function unlinkDisplayGroups()
...
+    public function updateTagValues($values)
+    {
+        $this->getLog()->debug('Tag options were changed, the following values need to be removed ' . json_encode($values));
+
+        foreach ($values as $value) {
+            $this->getLog()->debug('removing following value from lktag tables ' . $value);
+
+            if ($this->getStore()->exists('SELECT * FROM `lktagcampaign` INNER JOIN `tag` on lktagcampaign.tagId = tag.tagId WHERE `tag` = :tag AND value = :value', ['value' => $value, 'tag' => $this->tag])) {
+
+                $this->getStore()->update('UPDATE `lktagcampaign` SET `value` = null WHERE tagId = :tagId AND value = :value',
```
dasgarner: I dont think you need to do any of the prior checking - you can just run the update statement because if the where clause doesn't match, it won't update anything.

lib/Factory/LayoutFactory.php
```diff
@@ -452,6 +452,18 @@ public function getByBackgroundImageId($backgroundImageId)
...
     }
 
+    /**
+     * @param string $tag
+     * @return Layout[]
+     * @throws NotFoundException
+     */
+    public function getByTag($tag)
+    {
+        $layouts = $this->query(null, ['disableUserCheck' => 1, 'tags' => $tag, 'exactTags' => 1]);
```
dasgarner: you could just return without assigning to $layouts ?

ui/src/core/xibo-forms.js
```diff
@@ -1114,4 +1114,122 @@ function userProfileEditFormOpen() {
...
+                url: $('#'+formId).data().gettag,
+                type: "GET",
+                data: {
+                    name: tagN,
+                },
+                beforeSend: function () {
+                    $("#loadingValues").addClass('fa fa-spinner fa-spin loading-icon')
+                },
+                success: function (response) {
+                    if (response.success) {
```
dasgarner: What happens if the response isn't a success - should we have an else here that shows a system error message?

ui/src/core/xibo-forms.js
```diff
@@ -1114,4 +1114,122 @@ function userProfileEditFormOpen() {
...
+            $('#tagValue').focus();
+        }
+    });
+
+    $(formSelector).on('itemRemoved', function(event) {
+
+        if(tagN === event.item) {
+            $('#tagValueRequired').addClass('hidden');
+            $('.save-button').prop('disabled', false);
+            $('#tagValue').html('');
```
dasgarner: you can chain these together for efficiency

views/campaign-form-add.twig
```diff
@@ -25,14 +25,23 @@
 {% block formHtml %}
     <div class="row">
         <div class="col-md-12">
-            <form id="campaignAddForm" class="XiboForm form-horizontal" method="post" action="{{ urlFor("campaign.add") }}">
+            <form id="campaignAddForm" class="XiboForm form-horizontal" method="post" action="{{ urlFor("campaign.add") }}" data-gettag="{{ data.gettag }}">
```
dasgarner: it would be better to use urlFor directly here, rather than passing it from the controller

PR #593, opened 2019-06-26
Distribution report

ifarzana:


ifarzana:
retest this please

db/migrations/20190626110359_create_stat_table_migration.php
```diff
@@ -0,0 +1,54 @@
...
+        ->addColumn('scheduleId', 'integer')
+        ->addColumn('displayId', 'integer')
+        ->addColumn('campaignId', 'integer', ['default' => null, 'null' => true])
+        ->addColumn('layoutId', 'integer', ['default' => null, 'null' => true])
+        ->addColumn('mediaId', 'integer', ['default' => null, 'null' => true])
+        ->addColumn('widgetId', 'integer', ['default' => null, 'null' => true])
+        ->addColumn('start', 'integer')
+        ->addColumn('end', 'integer')
+        ->addColumn('tag', 'string', ['limit' => 254, 'default' => null, 'null' => true])
+        ->addColumn('duration', 'integer', ['null' => true])
```
dasgarner: I think these don't need to be NULL in this new table - we will always calculate and insert duration and count ?

lib/Controller/Report.php
```diff
@@ -104,7 +110,7 @@ class Report extends Base
      * @param LayoutFactory $layoutFactory
      * @param UserFactory $userFactory
      */
-    public function __construct($log, $sanitizerService, $state, $user, $help, $date, $config, $store, $timeSeriesStore, $reportService, $reportScheduleFactory, $savedReportFactory, $mediaFactory, $layoutFactory, $userFactory)
+    public function __construct($log, $sanitizerService, $state, $user, $help, $date, $config, $store, $timeSeriesStore, $reportService, $reportScheduleFactory, $savedReportFactory, $mediaFactory, $layoutFactory, $userFactory, $campaignFactory)
```
dasgarner: I think our recent refactor means we don't need $mediaFactory, $layoutFactory or $campaignFactory in this controller?

lib/XTR/StatsMigrationTask.php
```diff
@@ -0,0 +1,481 @@
...
+        $this->migrateStats();
+    }
+
+    public function migrateStats()
+    {
+        $this->runMessage = '# ' . __('Stats Migration') . PHP_EOL . PHP_EOL;
+
+        // read configOverride
+        $configOverrideFile = $this->getOption('configOverride', '');
+        if (file_exists($configOverrideFile)) {
```
dasgarner: can we use an array_merge here instead, that way we wouldn't have to provide all of the settings in the override.

something like:

```
if (!empty($configOverrideFile) && file_exists($configOverrideFile)) {
    $options = array_merge( json_decode(file_get_contents($configOverrideFile), true), $options);
}
```

lib/XTR/StatsMigrationTask.php
```diff
@@ -0,0 +1,481 @@
...
+            $options = [
+                'killSwitch' => $this->getOption('killSwitch', 0),
+                'numberOfRecords' => $this->getOption('numberOfRecords', 10000),
+                'numberOfLoops' => $this->getOption('numberOfLoops', 1000),
+                'pauseBetweenLoops' => $this->getOption('pauseBetweenLoops', 10),
+                'optimiseOnComplete' => $this->getOption('optimiseOnComplete', 1),
+            ];
+        }
+
+        if ($options['killSwitch'] == 1) {
```
dasgarner: reversed logic I think - if the killSwitch is ON then we shouldn't do anything

lib/XTR/StatsMigrationTask.php
```diff
@@ -0,0 +1,481 @@
...
+            'numberOfLoops' => 1000,
+            'pauseBetweenLoops' => 10,
+        ], $options);
+
+        // Get low watermark from file
+        $watermark = $this->getWatermarkFromFile($fileName, 'stat_archive');
+
+        $numberOfLoops = 0;
+        $count = 0;
+        while ($watermark > 0) {
```
dasgarner: will the watermark ever be 0?

lib/XTR/StatsMigrationTask.php
```diff
@@ -0,0 +1,481 @@
...
+
+        $sql = $this->store->getConnection()->prepare('SELECT statId FROM stat WHERE statId < :watermark ORDER BY statId DESC LIMIT 1');
+        $sql->bindParam(':watermark', $watermark, \PDO::PARAM_INT);
+        $sql->execute();
+
+        // Mark the Stats Archiver as disabled if there are records in stat table and set option archiveStats off
+        if ($sql->rowCount() > 0) {
+
+            // Quit the StatsArchiveTask if it is running
+            if ($archiveTask->runNow == 1) {
```
dasgarner: runNow flag doesn't mean the task is running - it means that it will run next time the task scheduler runs, even if its not scheduled to run.

I think you want to check the `status` instead

lib/XTR/StatsMigrationTask.php
```diff
@@ -0,0 +1,481 @@
...
+            // Keep count how many stats we've inserted
+            $recordCount = $stats->rowCount();
+            $count+= $recordCount;
+
+            // End of records
+            if ($this->checkEndOfRecords($recordCount, $fileName) === true) {
+
+                // Enable the StatsArchiver task and set option archiveStats on
+                $archiveTask->isActive = 1;
+                $archiveTask->options['archiveStats'] = 'On';
```
dasgarner: we shouldn't set the archiveStats option as the user might not want to archive stats - it should be the value it was when we started

lib/XTR/StatsMigrationTask.php
ifarzana: its the other way around I guess

>  // read configOverride
        $configOverrideFile = $this->getOption('configOverride', '');
        if (!empty($configOverrideFile) && file_exists($configOverrideFile)) {
            $options = array_merge($options, json_decode(file_get_contents($configOverrideFile), true));
        }

lib/XTR/StatsMigrationTask.php
ifarzana: no it will not. watermark is 1 if there are no records in stat/stat_archive.

dasgarner: so that means this loop wont ever finish?

the lowest statId in the table might be 8938549382 - do we handle that?

ifarzana: we do make it -1 when end of records is reached. will look into it in the morning again.

ifarzana: in checkEndOfRecords()

PR #610, opened 2019-07-10
OSX-702 Improve the way we handle Tags with values on various forms

PeterMis:


lib/Entity/DisplayGroup.php
```diff
@@ -527,18 +533,22 @@ public function replaceTags($tags = [])
...
-        });
+        if ($this->tags != $tags) {
+            $this->unassignTags = array_udiff($this->tags, $tags, function ($a, $b) {
+                /* @var Tag $a */
+                /* @var Tag $b */
+                return $a->tagId - $b->tagId;
+            });
 
-        $this->getLog()->debug('Tags to be removed: ' . json_encode($this->unassignTags));
+            $this->getLog()->debug('Tags to be removed: %s', json_encode($this->unassignTags));
```
dasgarner: using %s actually breaks the PSR LoggerInterface and only works because we dont obey it properly - just a point to note.

PR #611, opened 2019-07-10
Report with tags

ifarzana:


lib/XTR/StatsMigrationTask.php
```diff
@@ -104,12 +107,14 @@ public function migrateStats()
 
                 if( ($statSql->rowCount() == 0) && ($statArchiveSqlCount == 0) ) {
 
+                    $this->runMessage = '## Stat migration to Mongo' . PHP_EOL ;
+                    $this->runMessage = '- Both stat_archive and stat is empty. '. PHP_EOL . PHP_EOL;
```
dasgarner: this immediately overwrites the previous message - you could use appendMessage() to make this easier (and then you don't need to worry about the EOL either)

ifarzana: Can you check the modified StatsMigrationTask.php file please?

PR #614, opened 2019-07-18
Proof of play in report framework and improvements

ifarzana:


lib/Factory/ReportScheduleFactory.php
```diff
@@ -170,6 +171,12 @@ public function query($sortOrder = null, $filterBy = [])
             $params['reportName'] = $this->getSanitizer()->getString('reportName',  $filterBy);
         }
 
+        // isActive
+        if ($this->getSanitizer()->getInt('isActive', 0, $filterBy) != 0) {
```
dasgarner: I think this should be:

```
$this->getSanitizer()->getInt('isActive', $filterBy) != null
```

so that you can provide 1 or 0 to get active or not, and omit to provide anything to get both?

lib/Factory/ReportScheduleFactory.php
```diff
@@ -186,7 +193,7 @@ public function query($sortOrder = null, $filterBy = [])
         foreach ($this->getStore()->select($sql, $params) as $row) {
             $entries[] = $this->createEmpty()->hydrate($row, [
                 'intProperties' => [
-                    'reportScheduleId, lastRunDt, previousRunDt, lastSavedReportId'
+                    'reportScheduleId, lastRunDt, previousRunDt, lastSavedReportId, isActive'
```
dasgarner: this takes an array, like:

```
'reportScheduleId', 'lastRunDt', 'previousRunDt', 'lastSavedReportId', 'isActive'
```

lib/Report/DistributionReport.php
```diff
@@ -29,6 +30,11 @@ class DistributionReport implements ReportInterface
 
     use ReportTrait;
 
+    /**
+     * @var \Slim\Slim
+     */
+    public $app;
+
```
dasgarner: Do you need this anymore?

lib/Report/ProofOfPlay.php
```diff
@@ -0,0 +1,1152 @@
...
+        return $filter;
+    }
+
+    /**
+     * Set the sort order
+     * @return array
+     */
+    protected function gridRenderSort()
+    {
+        $columns = $this->app->request()->get('columns');
```
dasgarner: You can get this from the sanitizer so that you don't need $app in your class.

lib/Report/SummaryReport.php
```diff
@@ -27,6 +28,11 @@ class SummaryReport implements ReportInterface
 
     use ReportTrait;
 
+    /**
```
dasgarner: Don't think you need this.

lib/Factory/ReportScheduleFactory.php
ifarzana: Thanks!

lib/Factory/ReportScheduleFactory.php
ifarzana: :) missed the difference

PR #615, opened 2019-07-19
Bugfix/1.8.14 

PeterMis:


lib/Factory/PlayerVersionFactory.php
```diff
@@ -218,6 +218,31 @@ public function query($sortOrder = null, $filterBy = [])
             $params['playerCode'] = $this->getSanitizer()->getInt('playerCode', $filterBy);
         }
 
+        if ($this->getSanitizer()->getString('playerShowVersion') != null) {
```
dasgarner: Shouldn't this be `!==` ?

PR #617, opened 2019-07-19
Swift 2.1.0 rc1

PeterMis:


db/migrations/20190717101342_nullable_command_validation_string_migration.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+use Phinx\Migration\AbstractMigration;
+
+class NullableCommandValidationStringMigration extends AbstractMigration
+{
+    /** @inheritdoc */
+    public function change()
+    {
+        $this->execute('ALTER TABLE lkcommanddisplayprofile MODIFY validationString varchar(1000) NULL;');
```
dasgarner: Any reason we didn't use Phinx to change these columns? (execute directly like this means we cant rollback)

db/migrations/20190719074601_missing_default_value_migration.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+use Phinx\Migration\AbstractMigration;
+
+class MissingDefaultValueMigration extends AbstractMigration
+{
+    /** @inheritdoc */
+    public function change()
+    {
+        $this->execute('ALTER TABLE dataset MODIFY `lastDataEdit` int(11) DEFAULT 0');
```
dasgarner: Same question as other migration

PR #640, opened 2019-08-02
OSX-648 Schedule End Reminders

ifarzana:


lib/XTR/ScheduleReminderTask.php
```diff
@@ -0,0 +1,192 @@
...
+                        case ScheduleReminder::$TYPE_WEEK:
+                            $type = ScheduleReminder::$WEEK;
+                            $typeText = 'Week(s)';
+                            break;
+                        case ScheduleReminder::$TYPE_MONTH:
+                            $type = ScheduleReminder::$MONTH;
+                            $typeText = 'Month(s)';
+                            break;
+                        default:
+                            throw new \Xibo\Exception\NotFoundException('Unknown type');
```
dasgarner: We might want to log and `continue` here, otherwise an incorrect reminder would stop all reminders from processing until it was fixed (and there isn't enough info here to try and fix)

lib/XTR/ScheduleReminderTask.php
```diff
@@ -0,0 +1,192 @@
...
+                        $nextReminderDate = $schedule->getNextReminderDate($now, $reminder, $remindSeconds);
+                        if($nextReminderDate != null) {
+
+                            $reminder->reminderDt = $nextReminderDate;
+                            $reminder->save();
+                            $this->appendRunMessage(__('Reminder added for  '. $title));
+                        } else {
+
+                            // Remove the reminder
+                            $reminder->delete();
```
dasgarner: Should we be removing reminders that have passed? Seems like their reminder dates have gone by and we can safely just ignore them now?

views/schedule-form-add.twig
```diff
@@ -43,6 +43,7 @@
             <ul class="nav nav-tabs" role="tablist">
                 <li class="active"><a href="#general" role="tab" data-toggle="tab"><span>{% trans "General" %}</span></a></li>
                 <li class="repeats"><a href="#repeats" role="tab" data-toggle="tab"><span>{% trans "Repeats" %}</span></a></li>
+                <li class="reminder"><a href="#reminder" role="tab" data-toggle="tab"><span>{% trans "Reminder" %}</span></a></li>
```
dasgarner: Perhaps "reminders"? because there can be more than one

lib/Entity/Schedule.php
```diff
@@ -1148,4 +1170,133 @@ public function isCustomDayPart()
...
+     * @param Date $now
+     * @param ScheduleReminder $reminder
+     * @param int $remindSeconds
+     * @param int $recurrenceDetail
+     * @return int|null
+     * @throws XiboException
+     */
+    public function getNextReminderDate($now, $reminder, $remindSeconds, $recurrenceDetail = null) {
+
+        if ($recurrenceDetail == null) {
```
dasgarner: We probably don't need to pass this in?

lib/Controller/Schedule.php
```diff
@@ -1246,4 +1379,63 @@ public function scheduleNowForm($from, $id)
             'help' => $this->getHelp()->link('Schedule', 'ScheduleNow')
         ]);
     }
+
+    public function saveReminder($schedule, $scheduleReminder, $oldReminderDt = null)
```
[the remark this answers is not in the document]
ifarzana: No, it is not a public function, **public** keyword is removed now.

lib/Controller/Schedule.php
```diff
@@ -1246,4 +1379,63 @@ public function scheduleNowForm($from, $id)
...
+            if ($now->format('U') > $scheduleReminder->reminderDt) {
+                // find the next event from now where the remindetDt is greater or equal now
+
+                $nextReminderDt = $schedule->getNextReminderDate($now, $scheduleReminder, $remindSeconds, $schedule->recurrenceDetail);
+                $scheduleReminder->reminderDt = $nextReminderDt;
+            }
+        }
+
+        // Save only when reminderDt is not less than the current time
+        if($scheduleReminder->reminderDt >= $now->format('U') ) {
```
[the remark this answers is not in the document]
ifarzana: OK, I get it.

ifarzana:
retest this please

lib/XTR/ScheduleReminderTask.php
```diff
@@ -0,0 +1,228 @@
...
+                // Current reminderDt will be used as lastReminderDt
+                $reminder->lastReminderDt = $reminder->reminderDt;
+            }
+
+            // Save
+            $reminder->save();
+        }
+    }
+
+    function createNotification($subject, $body, $reminder, $schedule, $releaseDt = null) {
```
dasgarner: private/public/protected?

lib/XTR/StatsArchiveTask.php
```diff
@@ -205,11 +214,22 @@ private function tidyStats()
             $maxAttempts = $this->getOption('statsDeleteMaxAttempts', 10);
             $statsDeleteSleep = $this->getOption('statsDeleteSleep', 3);
 
-            $options = [
-                'maxAttempts' => $maxAttempts,
-                'statsDeleteSleep' => $statsDeleteSleep,
-                'limit' => 10000
-            ];
+            if ($this->timeSeriesStore->getEngine() == 'mongodb') {
```
dasgarner: we've got this handled here, but it seems like we should handle that inside the mongotimeseriesstore?

If we have a restriction in Mongo, then we ought to ignore the limit option internally i think, rather than handling it out here

lib/XTR/StatsMigrationTask.php
```diff
@@ -504,8 +503,11 @@ function migrationStatArchiveToMongo($options) {
 
             // Give Mongo time to recover
             if ($watermark > 0) {
-                $this->appendRunMessage('- '. $count. ' rows migrated.');
-                $this->log->debug('Mongo stats migration from stat_archive. '.$count.' rows effected, sleeping.');
+
+                if(count($statDataMongo) > 0 ) { // TODO FIX this
```
dasgarner: TODO ?

lib/Entity/Schedule.php
```diff
@@ -1148,4 +1206,132 @@ public function isCustomDayPart()
...
+     * @throws XiboException
+     */
+    public function getEventTitle() {
+
+        // Setting for whether we show Layouts with out permissions
+        $showLayoutName = ($this->config->getSetting('SCHEDULE_SHOW_LAYOUT_NAME') == 1);
+
+        // Load the display groups
+        $this->load();
+        $this->setCampaignFactory($this->campaignFactory);
```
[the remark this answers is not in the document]
ifarzana: That was a mistake. We are in fact calling that from outside i.e.. from ScheduleReminderTask.

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -455,9 +455,12 @@ public function deleteStats($maxage, $fromDt = null, $options = [])
         $options = array_merge([
             'maxAttempts' => 10,
             'statsDeleteSleep' => 3,
-            'limit' => 10000,
+            'limit' => 1000,
         ], $options);
 
+        // https://stackoverflow.com/questions/10014181/how-to-delete-documents-by-query-efficiently-in-mongo
```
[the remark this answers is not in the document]
dasgarner: did this get resolved in the latest changes?

PR #643, opened 2019-08-06
Swift 2.2.0 alpha

PeterMis:


PeterMis:
jenkins: retest this please

lib/Entity/User.php
```diff
@@ -261,6 +261,8 @@ class User implements \JsonSerializable
      */
     public $twoFactorRecoveryCodes = [];
 
+    public $showContentFrom = 1;
```
dasgarner: Can we have a description of this property here as its not immediately obvious what it means from the var name?

lib/routes.php
```diff
@@ -218,6 +218,7 @@
 $app->get('/library/usage/layouts/:id', '\Xibo\Controller\Library:usageLayouts')->name('library.usage.layouts');
 $app->get('/library/download/:id(/:type)', '\Xibo\Controller\Library:download')->name('library.download');
 $app->post('/library', '\Xibo\Controller\Library:add')->name('library.add');
+$app->post('/library/uploadURL', '\Xibo\Controller\Library:uploadFromURL')->name('library.uploadFromURL');
```
dasgarner: I think this should be camel case to fit without convention elsewhere? `uploadUrl`

PR #663, opened 2019-08-28
swift-2.2.0-alpha - Image compression

ifarzana:


lib/Entity/Media.php
```diff
@@ -858,12 +858,39 @@ public function saveFile()
         // Set to valid
         $this->valid = 1;
 
+        $filePath = $libraryFolder . $this->storedAs;
+        list($img_width, $img_height) = @getimagesize($filePath);
+
+        $resize_threshold = $this->config->getSetting('DEFAULT_RESIZE_THRESHOLD');
```
dasgarner: `_` in var names should be camel case ;)

lib/Entity/Media.php
```diff
@@ -858,12 +858,39 @@ public function saveFile()
...
+        $resize_limit = $this->config->getSetting('DEFAULT_RESIZE_LIMIT');
+
+        // Media released set to 0 for large size images
+        // if image size is greater than 8000 X 8000 then we flag that image as too big
+        if ($img_width > $resize_limit || $img_height > $resize_limit) {
+            $this->released = 2;
+        } elseif ($img_width > $img_height) { // 'landscape';
+
+            if ($img_width <= $resize_threshold && $img_height <= 1080 ) {
```
dasgarner: Why are we also doing the image height here? `$img_height <= 1080`

lib/Entity/Media.php
```diff
@@ -858,12 +858,39 @@ public function saveFile()
...
+                $this->released = 1;
+            } elseif ($img_height > 1080 || $img_width > $resize_threshold) {
+                $this->released = 0;
+            }
+        } else { // 'portrait';
+
+            if ($img_width <= 1080 && $img_height <= $resize_threshold ) {
```
dasgarner: again, not sure why we test width here?

lib/Factory/MediaFactory.php
```diff
@@ -633,6 +633,10 @@ public function query($sortOrder = null, $filterBy = [])
             $params['type'] = $this->getSanitizer()->getString('type', $filterBy);
         }
 
+        if ($this->getSanitizer()->getInt('imageProcessing', $filterBy) !== null) {
+            $body .= 'AND ( media.type = \'image\' OR media.type = \'module\' ) ';
```
dasgarner: we should also say where moduleSystemFile = 0 so that we don't process module files that are images (like the Flags in the stocks module for example)

something like:
```
$body .= 'AND ( media.type = \'image\' OR (media.type = \'module\' and `media`.moduleSystemFile =0) ) ';
```

lib/Helper/XiboUploadHandler.php
```diff
@@ -286,6 +286,29 @@ protected function handle_form_data($file, $index)
                     $media->enableStat = $controller->getConfig()->getSetting('MEDIA_STATS_ENABLED_DEFAULT');
                 }
 
+                $resize_threshold = $controller->getConfig()->getSetting('DEFAULT_RESIZE_THRESHOLD');
```
dasgarner: the same comments apply to this section

lib/Middleware/Storage.php
```diff
@@ -25,6 +25,7 @@
 
 use Slim\Helper\Set;
 use Slim\Middleware;
+use Xibo\Service\ImageProcessingService;
```
dasgarner: Left over use statement here

lib/Service/ImageProcessingService.php
```diff
@@ -0,0 +1,83 @@
...
+
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setDependencies($log, $date)
+    {
+        $this->log = $log;
+        $this->dateService = $date;
```
dasgarner: Why do we need the date service? its pretty light weight, im just curious

lib/Widget/Image.php
```diff
@@ -257,6 +257,12 @@ public function getResource($displayId = 0)
     /** @inheritdoc */
     public function isValid()
     {
+        if ($this->getMedia()->released == 0) {
+            return self::$STATUS_INVALID;
+        } elseif ($this->getMedia()->released == 2) {
+            return self::$STATUS_INVALID;
+        }
+
```
dasgarner: Didn't we have some code in here to set the Valid Message?

lib/XTR/ImageProcessingTask.php
```diff
@@ -0,0 +1,105 @@
...
+        $images = $this->mediaFactory->query(null, ['released' => 0, 'allModules' => 1, 'imageProcessing' => 1]);
+
+        $libraryLocation = $this->config->getSetting('LIBRARY_LOCATION');
+        $resize_threshold = $this->config->getSetting('DEFAULT_RESIZE_THRESHOLD');
+
+        // Get list of Images
+        foreach ($images as $media) {
+
+            $filePath = $libraryLocation . $media->storedAs;
+            list($img_width, $img_height) = @getimagesize($filePath);
```
dasgarner: `_`'s

lib/Factory/MediaFactory.php
```diff
@@ -634,7 +634,7 @@ public function query($sortOrder = null, $filterBy = [])
         }
 
         if ($this->getSanitizer()->getInt('imageProcessing', $filterBy) !== null) {
-            $body .= 'AND ( media.type = \'image\' OR media.type = \'module\' ) ';
+            $body .= 'AND ( media.type = \'image\' OR media.type = \'module\' AND media.moduleSystemFile = 0 ) ';
```
dasgarner: sorry but this still isn't right - you need extra `()` around the 2nd and 3rd criteria

```
$body .= 'AND ( media.type = \'image\' OR (media.type = \'module\' AND media.moduleSystemFile = 0) ) ';
```

lib/XTR/ImageProcessingTask.php
```diff
@@ -75,19 +70,19 @@ private function runImageProcessing()
...
 
             $filePath = $libraryLocation . $media->storedAs;
-            list($img_width, $img_height) = @getimagesize($filePath);
+            list($imgWidth, $imgHeight) = @getimagesize($filePath);
 
             // Orientation of the image
-            if ($img_width > $img_height) { // 'landscape';
-                $this->imageProcessingService->resizeImage($filePath, $resize_threshold, 1080);
+            if ($imgWidth > $imgHeight) { // 'landscape';
+                $this->imageProcessingService->resizeImage($filePath, $resizeThreshold, 1080);
```
dasgarner: I didn't notice this during the last review, but it looks odd to be passing 1080 into here? if its necessary can we add a comment to explain why we do that?

It feels more natural to pass `null` into the parameter which we don't want to change?

ifarzana:
retest this please

PR #664, opened 2019-08-29
Swift 2.2.0 alpha

PeterMis:


lib/Controller/Layout.php
```diff
@@ -352,6 +352,9 @@ function add()
 
         $this->getLog()->debug('Layout Added');
 
+        // automatically checkout the new layout for edit
+        $this->checkout($layout->layoutId);
```
dasgarner: This is an interesting use of the controller - which I think is OK

lib/Entity/Playlist.php
```diff
@@ -928,4 +931,56 @@ public function cloneClosureTable($newParentId)
             'parentId' => $this->playlistId
         ]);
     }
+
+    public function generatePlaylistMapping($widgets, $parentId, &$playlistMappings, &$count, &$nestedPlaylistDefinitions, &$dataSetIds, &$dataSets, $dataSetFactory, $includeData)
```
dasgarner: PHP Doc would be good here to explain what this does and why we're passing by reference

lib/Entity/Playlist.php
```diff
@@ -928,4 +931,56 @@ public function cloneClosureTable($newParentId)
...
+
+                        $playlistMappings[$parentId][$nestedPlaylist->playlistId] = [
+                            'parentId' => $parentId,
+                            'playlist' => $nestedPlaylist->name,
+                            'playlistId' => $nestedPlaylist->playlistId
+                        ];
+
+                        $count++;
+
+                        $this->generatePlaylistMapping($nestedPlaylist->widgets, $nestedPlaylist->playlistId, $playlistMappings, $count, $nestedPlaylistDefinitions,$dataSetIds, $dataSets, $dataSetFactory, $includeData);
```
dasgarner: Good to put a comment above this pointing out that this is a recursive function

lib/Entity/Layout.php
```diff
@@ -1436,11 +1447,41 @@ public function toZip($dataSetFactory, $fileName, $options = [])
...
+
+                foreach ($playlistIds as $playlistId) {
+                    $count = 1;
+                    $playlist = $this->playlistFactory->getById($playlistId);
+                    $playlist->load();
+                    $playlist->expandWidgets(0, false);
+                    $playlistDefinitions[$playlist->playlistId] = $playlist;
+
+                    // we are adding Playlist definitions, Playlist mappings and DataSets existing on the nested Playlist.
+                    $playlistMappings = $playlist->generatePlaylistMapping($playlist->widgets, $playlist->playlistId,$playlistMappings, $count, $nestedPlaylistDefinitions, $dataSetIds, $dataSets, $dataSetFactory, $options['includeData']);
```
dasgarner: We also pass `$playlistMappings` by reference, so we do need to assign it as well?

lib/Factory/LayoutFactory.php
```diff
@@ -818,10 +1123,10 @@ public function createFromZip($zipFile, $layoutName, $userId, $template, $replac
                 $layout->backgroundImageId = $newMediaId;
             } else if ($isFont) {
                 // Just raise a flag to say that we've added some fonts to the library
-                if ($newMedia)
+                if ($newMedia) {
```
dasgarner: this has been bothering me for a while - pleased you fixed it!

PR #678, opened 2019-09-18
OSX-830 Interrupt Layout CMS changes

PeterMis:


PeterMis:
Retest this please

db/migrations/20190917093141_interrupt_layout_migration.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /** @inheritdoc */
+    public function change()
+    {
+        $scheduleTable = $this->table('schedule');
+
+        // Add a two new columns to Display table, newCmsAddress and newCmsKey
+        if (!$scheduleTable->hasColumn('shareOfVoice')) {
+            $scheduleTable
+                ->addColumn('shareOfVoice', 'integer', ['limit' => 100, 'default' => null, 'null' => true])
```
dasgarner: that is a large integer! i think a TINY_INT would do the trick? also the comment might want updating ;)

lib/Controller/Schedule.php
```diff
@@ -1224,6 +1278,57 @@ public function edit($eventId)
             $this->getLog()->debug('Processed start is: FromDt=' . $fromDt->toRssString());
         }
 
+        // interrupt layout
+        if ($schedule->eventTypeId == 4) {
```
dasgarner: Do we have a static variable for this ID? Might be better to use it here instead of `4`

lib/Controller/Schedule.php
```diff
@@ -1224,6 +1278,57 @@ public function edit($eventId)
...
+
+            $campaign = $this->campaignFactory->getById($schedule->campaignId);
+
+            if ($campaign->isLayoutSpecific == 0) {
+                throw new InvalidArgumentException(__('Cannot select a campaign for this event type.'), 'campaignId');
```
dasgarner: Campaign is a Xibo object and should be a capital `C`

lib/Controller/Schedule.php
```diff
@@ -1224,6 +1278,57 @@ public function edit($eventId)
...
+                $layoutDuration = $layout->duration;
+                $shareInSeconds = ($schedule->shareOfVoice * 3600) / 100;
+
+                if ($layoutDuration > $shareInSeconds) {
+                    throw new InvalidArgumentException(__('Layout duration is longer than share of voice in the hour period'), 'shareOfVoice');
+                }
+            }
+
+            foreach($schedule->displayGroups as $displayGroup) {
+                $events = $this->scheduleFactory->getByDisplayGroupId($displayGroup->displayGroupId);
```
dasgarner: I see what you're doing here, but I don't think this is sustainable - over time we will built up more and more events on the DisplayGroup which need to be assessed. Imagine an only system with 10000 events - we go through them all here, one by one.

If we passed this events from/to date into the scheduleFactory, we could only return events here that were overlapping. It may be easier to call the `query` method directly to do that.

lib/Entity/Schedule.php
```diff
@@ -52,6 +52,7 @@ class Schedule implements \JsonSerializable
     public static $LAYOUT_EVENT = 1;
     public static $COMMAND_EVENT = 2;
     public static $OVERLAY_EVENT = 3;
+    public static $INTERRUPT_EVENT = 4;
```
dasgarner: Great - we should use this in other places where you've got 4 hard coded please

PR #701, opened 2019-10-09
Change required file download order

inverted-logic:
Download layout resources before downloading layout to fix bug where expired layout was using old resources.

### DIFF
--- a/lib/Xmds/Soap.php
+++ b/lib/Xmds/Soap.php
@@ -693,14 +693,14 @@ class Soap
                             $updatedDt = ($updatedDt->greaterThan($cachedDt)) ? $updatedDt : $cachedDt;
 
                             // Append this item to required files
-                            $resource_file = $requiredFilesXml->createElement("file");
-                            $resource_file->setAttribute('type', 'resource');
-                            $resource_file->setAttribute('id', $widget->widgetId);
-                            $resource_file->setAttribute('layoutid', $layoutId);
-                            $resource_file->setAttribute('regionid', $region->regionId);
-                            $resource_file->setAttribute('mediaid', $widget->widgetId);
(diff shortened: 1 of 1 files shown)


lib/Xmds/Soap.php
```diff
@@ -692,18 +693,21 @@ protected function doRequiredFiles($serverKey, $hardwareKey, $httpDownloads)
...
                             // Append this item to required files
-                            $file = $requiredFilesXml->createElement("file");
-                            $file->setAttribute('type', 'resource');
-                            $file->setAttribute('id', $widget->widgetId);
-                            $file->setAttribute('layoutid', $layoutId);
-                            $file->setAttribute('regionid', $region->regionId);
-                            $file->setAttribute('mediaid', $widget->widgetId);
-                            $file->setAttribute('updated', $updatedDt->format('U'));
-                            $fileElements->appendChild($file);
+                            $resource_file = $requiredFilesXml->createElement("file");
```
dasgarner: The project uses lower camel case please, so this would be `$resourceFile`

inverted-logic: Variable names updated to match project format.

I am not familiar with the player code, but I have witnessed players concurrently download required files saying to me that download order isn't guaranteed as you say. However, this change did resolve a problem I was experiencing.

I was inspired to alter this by the comments starting on line 450.

dasgarner: As you say, it won't hurt to start off the downloads in a better order - and you're also right when you say that a Player won't check whether it has all new resources for the current Layout before it interrupts it (we do that so that we don't go to the splash screen in the above situation)

I think its well worth having the order changed, i'll merge.

Thanks

PR #702, opened 2019-10-14
2.1.2

ifarzana:


lib/Entity/Schedule.php
```diff
@@ -474,6 +474,37 @@ public function validate()
...
         if ($this->recurrenceType != '' && ($this->recurrenceDetail === null || $this->recurrenceDetail <= 0))
             throw new InvalidArgumentException(__('Repeat every must be a positive number'), 'recurrenceDetail');
+        
+        // Check and disallow Minute/Hourly repeats to be indefinite (or too long)
+        $twelveHoursInSeconds = 12 * 60 * 60;
+        $oneWeekInSeconds = 24 * 7 * 60 * 60;
+        if($this->recurrenceType == 'Minute') {
+
+            if (empty($this->recurrenceRange)) {
+                throw new InvalidArgumentException(__('You must provide a date and time to end the Repeat'), 'recurrenceRange');
```
dasgarner: Can we make this a little friendlier? Something like:

"An end time is needed for an event that repeats every minute."

lib/Entity/Schedule.php
```diff
@@ -474,6 +474,37 @@ public function validate()
...
+            }
+
+            $distance = ($this->getDate()->parse($this->recurrenceRange, 'U')->diffInSeconds($this->getDate()->parse($this->fromDt, 'U'))) / $this->recurrenceDetail;
+
+            if ($distance > $twelveHoursInSeconds) {
+                // Recurrence range cannot be more than 12 hours
+                $exceedLimit = $this->getDate()->parse($this->fromDt, 'U')->addSeconds($twelveHoursInSeconds * $this->recurrenceDetail );
+                throw new InvalidArgumentException(__('The date and time to end the Repeat for this Event cannot exceed '.$exceedLimit), 'recurrenceRange');
```
dasgarner: For translated strings that have a variable, we should use sprintf so that the translator can put the variable in the appropriate place. E.g.

```
sprintf(__('The end time for this event can only be %d seconds in the future because of the repeating interval being minute or hour.', $exceedLimit))
```

lib/Report/ProofOfPlay.php
```diff
@@ -541,7 +541,7 @@ private function getProofOfPlayReportMySql($fromDt, $toDt, $displayIds, $layoutI
     {
 
         $fromDt = $fromDt->format('U');
-        $toDt = $toDt->startOfDay()->addDay()->format('U'); // added a day
+        $toDt = $toDt->format('U'); // added a day
```
dasgarner: The comment here and the code don't match up?

lib/Entity/Schedule.php
```diff
@@ -810,6 +845,11 @@ private function generateMonth($generateFromDt, $start, $end)
             /** @var Date $start */
             $start = $lastWatermark->copy();
             $end = $start->copy()->addSeconds($eventDuration);
+
+            if (($start <= $generateToDt) && ($end >= $generateFromDt)) {
```
dasgarner: This might be clearer and more consistent with the rest of the code if it had fewer brackets?

```
if ($start <= $generateToDt && $end >= $generateFromDt) {
```

PR #703, opened 2019-10-14
Check if timezone exists in moment then use it

mhmoosazadeh:
Hi
In 'bundle_vendor.js' file, We check for the 'calendar_type' and if it is equal to the 'Jalali', Then we set 'window.moment' as 'moment-jalaali' package. But this package doesn't have timezone function (.tz), So whenever we use .tz an error acquired.

In this PR we check if '.tz' available or undefined and then use it.


dasgarner:
jenkins: test this please

dasgarner:
xibosignage/xibo#1902

dasgarner:
Great, I will retarget this PR and associated bug back to 2.1.2 release (release21 branch) and get it in as a bugfix.

Thanks @mhmoosazadeh

dasgarner:
jenkins: retest this please

PR #706, opened 2019-10-17
OSX-573 Player-auth CMS side

PeterMis:


lib/Controller/Display.php
```diff
@@ -1836,4 +1838,46 @@ public function moveCms($displayId)
...
+        $cmsKey = $this->getConfig()->getSetting('SERVER_KEY');
+
+        if ($user_code == '') {
+            throw new InvalidArgumentException('Code cannot be empty', 'code');
+        }
+
+        $guzzle = new Client();
+
+        try {
+            $request = $guzzle->request('POST', 'https://auth.signlicence.co.uk/addDetails',
```
dasgarner: I think it would be sensible to add some comments here explaining what this URL is and what this does in general, so that people are not concerned about what is happening to their data.

views/display-form-addViaCode.twig
```diff
@@ -0,0 +1,49 @@
...
+
+{% block formHtml %}
+    <div class="row">
+        <div class="col-md-12">
+            <form id="displayAddViaCodeForm" class="XiboForm form-horizontal" method="post" action="{{ urlFor("display.addViaCode") }}">
+
+                {% set title %}{% trans "Code" %}{% endset %}
+                {% set helpText %}{% trans "Please provide the code displayed on the Player screen" %}{% endset %}
+                {{ forms.input("user_code", title, "", helpText) }}
+
```
dasgarner: Would it be sensible here to show an info box explaining what happens when they submit the form, and that their CMS will need access to the Authentication Service for this to work?

PR #715, opened 2019-10-28
Wolf2.3.0alpha

ifarzana:


lib/routes-web.php
```diff
@@ -66,6 +66,7 @@
 $app->get('/schedule/form/add', '\Xibo\Controller\Schedule:addForm')->name('schedule.add.form');
 $app->get('/schedule/form/edit/:id', '\Xibo\Controller\Schedule:editForm')->name('schedule.edit.form');
 $app->get('/schedule/form/delete/:id', '\Xibo\Controller\Schedule:deleteForm')->name('schedule.delete.form');
+$app->get('/schedule/form/deleterecurrence/:id', '\Xibo\Controller\Schedule:deleteRecurrenceForm')->name('schedule.recurrence.delete.form');
```
maurofmferrao: I would probably use '/schedulerecurrence/form/delete/:id' path instead, to match the other routes :)

ui/src/vendor/calendar/js/calendar.js
```diff
@@ -665,6 +665,8 @@ if(!String.prototype.formatNum) {
 	Calendar.prototype._month = function(month) {
 		this._loadTemplate('year-month');
 
+		console.log('yeah-month');
```
maurofmferrao: Do we need this log here? :P

lib/routes-web.php
maurofmferrao: I would probably use '/schedulerecurrence/form/delete/:id' path instead, to match the other routes :)

docker-compose.yml
```diff
@@ -43,4 +43,6 @@ services:
     image: ianw/quickchart
 
 networks:
-  default:
\ No newline at end of file
+  default:
```
dasgarner: I suspect this is a whoops?

lib/Controller/Schedule.php
```diff
@@ -978,6 +988,10 @@ public function add()
      */
     function editForm($eventId)
     {
+        // Recurring event start/end
+        $eventStart = $this->getSanitizer()->getString('eventStart', 1000) / 1000;
```
dasgarner: Are these provided to you as integers? if so should we use getInt instead?

lib/Controller/Schedule.php
```diff
@@ -1011,7 +1025,81 @@ function editForm($eventId)
...
+    function deleteRecurrenceForm($eventId)
+    {
+        // Recurring event start/end
+        $eventStart = $this->getSanitizer()->getString('eventStart', 1000);
+        $eventEnd = $this->getSanitizer()->getString('eventEnd', 1000);
+
+        $schedule = $this->scheduleFactory->getById($eventId);
+        $schedule->load();
+
+        if (!$this->isEventEditable($schedule->displayGroups))
```
dasgarner: Can we add `{}` to these types of statements, just so we're all doing the same thing going forward

lib/Entity/Schedule.php
```diff
@@ -496,6 +503,7 @@ public function validate()
             || $this->eventTypeId == Schedule::$OVERLAY_EVENT
             || $this->eventTypeId == Schedule::$INTERRUPT_EVENT
         ) {
+        //if ($this->eventTypeId == Schedule::$LAYOUT_EVENT || $this->eventTypeId == Schedule::$CAMPAIGN_EVENT || $this->eventTypeId == Schedule::$OVERLAY_EVENT || $this->eventTypeId == Schedule::$INTERRUPT_EVENT) {
```
dasgarner: is this comment here for a reason?

lib/Entity/Schedule.php
```diff
@@ -839,6 +852,20 @@ public function getEvents($fromDt, $toDt)
...
+
+                $exclude = false;
+                foreach ($scheduleExclusions as $k => $exclusion) {
+                    if ($scheduleEvent->fromDt == $exclusion->fromDt &&
+                        $scheduleEvent->toDt == $exclusion->toDt) {
+                        $exclude = true;
+                    }
+                }
+
+                if ($exclude)
```
dasgarner: `{}` :-)

lib/Entity/Schedule.php
```diff
@@ -659,6 +667,11 @@ public function delete()
         $this->displayGroups = [];
         $this->unlinkDisplayGroups();
 
+        // Delete schedule exclusions
+        $scheduleExclusions = $this->scheduleExclusionFactory->query(null, ['eventId' => $this->eventId]);
```
dasgarner: Sorry i'm not clear on something - the $exclusion->delete method deletes by `eventId` which I think will delete all of the events returned by the query in one go?

So the first delete will do something, and then the others wont effect any records, because they are already gone.

lib/Entity/ScheduleExclusion.php
```diff
@@ -0,0 +1,84 @@
...
+use Xibo\Storage\StorageServiceInterface;
+
+
+/**
+ * Class ScheduleExclusion
+ * @package Xibo\Entity
+ *
+ * @SWG\Definition()
+ */
+class ScheduleExclusion implements \JsonSerializable
```
dasgarner: Query whether we need the actual primary key in here so we can delete individuals

views/schedule-recurrence-form-delete.twig
```diff
@@ -0,0 +1,37 @@
+{#
```
dasgarner: Could you update this copyright please?

PR #718, opened 2019-11-14
Removed recursive chown

pascallieverse:
This commit replaces the recursive chown in the entrypoint.sh for the cms folder to a non-recursive chown on some files and set the correct permissions when copying the files to the docker container.
See issue's https://github.com/xibosignage/xibo/issues/2010 and https://github.com/docker/for-linux/issues/388

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -151,7 +151,7 @@ COPY --from=composer /app /var/www/cms
 COPY --from=webpack /app/web/dist /var/www/cms/web/dist
 
 # All other files (.dockerignore excludes many things, but we tidy up the rest below)
-COPY . /var/www/cms
+COPY --chown=apache:apache . /var/www/cms
 
 # Tidy up
 RUN rm /var/www/cms/composer.* && \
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -203,7 +203,11 @@ then
     /bin/chmod g+s /usr/sbin/ssmtp
 
     mkdir -p /var/www/cms/library/temp
-    chown apache.apache -R /var/www/cms
```
alexharrington: Unfortunately these don't do the same job as before.

The issue is that users may modify files inside those directories themselves, and this step when run ensures that the folders and their contents have correct permissions.

Making each of those recursive would probably do the same job, but will probably be just as slow unfortunately.

Dockerfile
```diff
@@ -151,7 +151,7 @@ COPY --from=composer /app /var/www/cms
 COPY --from=webpack /app/web/dist /var/www/cms/web/dist
```
alexharrington: These files (and others) potentially have incorrect ownership now?

pascallieverse: No, the file permission are looking fine inside the dist folder

docker/entrypoint.sh
pascallieverse: Understood, I tested the recursion on the mounted volumes only and it seems to be working just as fast as my last commit. (I don't have any custom files in the mounted volumes).

I have commited a new version.

dasgarner:
Merged with thanks! :tada:

PR #719, opened 2019-11-19
Patch to allow server timeout for MongoDB

ifarzana:


lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -533,6 +543,16 @@ public function executeQuery($options = [])
 
         } catch (\MongoDB\Exception\RuntimeException $e) {
             $this->log->error($e->getMessage());
+            throw new NotFoundException($e->getMessage());
```
dasgarner: We're handling all of these in the same way, so should we try and capture a higher level exception? I think also that GeneralException might be more applicable that not found? NotFound will give a 404 status code which isn't quite right here.

PR #730, opened 2019-11-26
Wolf OSX-887

PeterMis:
CSV as source for remote Datasets


lib/Factory/DataSetFactory.php
```diff
@@ -437,9 +439,20 @@ public function callRemoteService(DataSet $dataSet, DataSet $dependant = null, $
                 }
 
                 // TODO: we should probably do some checking to ensure we have JSON back
-                $result->entries[] = json_decode($request->getBody());
-                $result->number = $result->number + 1;
+                if ($dataSet->sourceId === 1) {
+                    $result->entries[] = json_decode($request->getBody());
+                    $result->number = $result->number + 1;
+                } else {
+                    $csv = file_get_contents($resolvedUri);
```
dasgarner: We're double tapping the url here? We already have a $request object which should contain this data - getBody() on that object should give the same results.

file_get_contents will go and get the same URL again and might not work on all systems (Guzzle is better to use).

PR #732, opened 2019-11-27
Patch - Print Style Sheet for pop reports

ifarzana:


views/user-welcome.twig
```diff
@@ -17,6 +17,55 @@
...
+        }
+
+        /* Reports of summary, distribution */
+        #reportChart {
+            max-width: 25cm;
+            max-height: 7cm;
+        }
+
+        /* Custom reports*/
+        #campaignChart, #sensor1Chart, #sensor2Chart, #sensor3Chart {
```
dasgarner: We're going to need to handle this differently as we can't name all of the reports individually.

Can you set some generic class names for this function, and then apply those to the reports?

views/user-welcome.twig
```diff
@@ -17,6 +17,55 @@
     .tour .popover-title{
         background-color: #b2bfdc;
     }
+
+    /* Print Style Sheet - for POP reports */
+    @media print {
```
dasgarner: Unfortunately I don't think this is the file we need to put it in - can we create a `print.css` file in the default theme, and then add that to the base.twig file as a CSS reference?

PR #736, opened 2019-12-02
Bugfix2.2.1

ifarzana:


lib/Entity/Media.php
```diff
@@ -872,41 +872,77 @@ public function saveFile()
         // Set to valid
         $this->valid = 1;
 
-        $filePath = $libraryFolder . $this->storedAs;
-        list($imgWidth, $imgHeight) = @getimagesize($filePath);
+        // Resize image dimensions if threshold exceeds
+        Media::assessDimensions();
```
dasgarner: this isn't a static call? should be `$this->` ?

lib/XTR/ImageProcessingTask.php
```diff
@@ -85,16 +86,23 @@ private function runImageProcessing()
...
             }
 
+            // Clears file status cache
+            clearstatcache(true, $filePath);
+
             // Release image and save
-            // Work out the MD5
-            $media->md5 = md5_file($libraryLocation . $media->storedAs);
+            // Work out the MD5 and Filesize
+            $media->md5 = md5_file($filePath);
```
dasgarner: Maybe it would make more sense to set these inside the release() method?

so we call: ```$media->release(md5_file($filePath), filesize($filePath));```

and then inside release, set the properties (including released = 1)

PR #738, opened 2019-12-02
OSX-539 Playlist Designer inside Layout Designer

maurofmferrao:


views/common.twig
```diff
@@ -409,7 +410,7 @@
             var errorMessagesTrans = {
                 error: "{% trans "ERROR" %}",
                 loadingLayout: "{% trans "There was a problem loading the layout!" %}",
```
dasgarner: Layout and Playlist should be capitals here (no need to amend this PR, but you could modify for the next one).

maurofmferrao: I will create an issue for those changes and apply them on the next PR

PR #744, opened 2019-12-06
Feature/playlist dashboard

ifarzana:


lib/Controller/Module.php
```diff
@@ -622,6 +622,29 @@ public function deleteWidgetForm($widgetId)
         ]);
     }
 
+    /**
+     * Delete Playlist Widget Form
+     * @param int $widgetId
+     * @throws XiboException
+     */
+    public function deletePlaylistWidgetForm($widgetId)
```
dasgarner: Would it make more sense for this function to be in the Playlist Dashboard Controller?

lib/routes.php
```diff
@@ -167,6 +167,7 @@
 $app->post('/playlist/widget/:type/:id', '\Xibo\Controller\Module:addWidget')->name('module.widget.add');
 $app->put('/playlist/widget/:id', '\Xibo\Controller\Module:editWidget')->name('module.widget.edit');
 $app->delete('/playlist/widget/:id', '\Xibo\Controller\Module:deleteWidget')->name('module.widget.delete');
+$app->delete('/playlist/dashboard/widget/:id', '\Xibo\Controller\Module:deletePlaylistWidget')->name('playlist.module.widget.delete');
```
dasgarner: I dont see this method added to the Module controller?
Even so - perhaps it would be better to have this in the Playlist Dashboard controller, and for the URL to be playlistdashboard?

db/migrations/20191205180116_add_playlist_dashboard_page_user_migration.php
```diff
@@ -30,26 +30,25 @@ class AddPlaylistDashboardPageUserMigration extends AbstractMigration
...
+                   ');
+        $pageId = $this->getAdapter()->getConnection()->lastInsertId();
+
+        $group = $this->execute(
+            'INSERT INTO `group` SET `group`=\'Playlist Dashboard User\', `isUserSpecific`= 0, `isEveryone`= 0, `isSystemNotification`= 0;
+                   ');
+        $groupId = $this->getAdapter()->getConnection()->lastInsertId();
+
+        // Set Playlist Dashboard Page Permission
+        $this->execute('INSERT INTO `permission` (`entityId`, `groupId`, `objectId`, `view`) SELECT  1, '.$groupId.', '.$pageId.', 1');
```
dasgarner: @PeterMis - I can't remember whether we can rely on these IDs or if we should look them up - I think look them up?

PeterMis: entityId for Page entity in permissionentity is set to 1 on install, I think that should be fine.
However it might be better to be safe than sorry and look it up, just in case :).

PR #746, opened 2019-12-11
Bugfix xibosignage/xibo#2049

ifarzana:
xibosignage/xibo#2049
xibosignage/xibo#2051

### DIFF
--- a/lib/Entity/Media.php
+++ b/lib/Entity/Media.php
@@ -901,28 +901,33 @@ class Media implements \JsonSerializable
             // Media released set to 0 for large size images
             // if image size is greater than 8000 X 8000 then we flag that image as too big
             if ($imgWidth > $resizeLimit || $imgHeight > $resizeLimit) {
-                $this->released = 2;
-                $this->getLog()->debug('Image size is too big. MediaId '. $this->mediaId);
-
+                if ($resizeLimit > 0) {
+                    $this->released = 2;
+                    $this->getLog()->debug('Image size is too big. MediaId '. $this->mediaId);
+                }
             } elseif ($imgWidth > $imgHeight) { // 'landscape';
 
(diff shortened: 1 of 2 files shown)


lib/Entity/Media.php
```diff
@@ -901,28 +901,33 @@ private function assessDimensions()
             // Media released set to 0 for large size images
             // if image size is greater than 8000 X 8000 then we flag that image as too big
             if ($imgWidth > $resizeLimit || $imgHeight > $resizeLimit) {
```
dasgarner: What you've got here will definitely work - but I think we might be able to make it more DRY

```
if ($resizeLimit > 0 && ($imgWidth > $resizeLimit || $imgHeight > $resizeLimit)) {
```

Then have another elseif:

```
elseif ($resizeLimit <= 0) {
   // the current imgWidth > $imgHeight clauses...
}
```
```

File: lib/Entity/Media.php [hunk 2]
```diff
@@ -899,30 +899,35 @@ private function assessDimensions()
...
-                }
-            } else { // 'portrait';
-
-                if ($imgHeight <= $resizeThreshold) {
-                    $this->released = 1;
-                } else {
-                    $this->released = 0;
-                    $this->getLog()->debug('Image exceeded threshold, released set to 0. MediaId '. $this->mediaId);
-
+            } elseif ($resizeLimit <= 0) {
```
dasgarner: Should this be `$resizeThreshold` ?

dasgarner: And actually `>0`

So we're trying to say, only go into the if block when we need to do some resizing. If the resizeThreshold is `>0` then we should do resizing, otherwise not.

ifarzana: That makes sense

PR #749, opened 2019-12-13
Bugfix/release22

ifarzana:
xibosignage/xibo#2052

### DIFF
--- a/lib/Controller/Report.php
+++ b/lib/Controller/Report.php
@@ -745,7 +745,7 @@ class Report extends Base
 
             try {
                 $mpdf = new \Mpdf\Mpdf([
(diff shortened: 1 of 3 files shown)


lib/Controller/Report.php
```diff
@@ -745,6 +745,7 @@ public function savedReportExport($savedreportId, $reportName)
 
             try {
                 $mpdf = new \Mpdf\Mpdf([
+                    'tempDir' => realpath(__DIR__ . '/../../library/temp'),
```
dasgarner: You will need to look up the LIBRARY_LOCATION setting here instead, because a relative path won't always work

PR #750, opened 2019-12-13
Feature/2.3.0 alpha layout editor upgrade

maurofmferrao:


views/layout-designer-page.twig
```diff
@@ -74,12 +74,12 @@
                     edit: "{% trans "Edit layout regions" %}",
                     addRegion: "{% trans "Add" %}",
                     addRegionDesc: "{% trans "Add a new region" %}",
-                    deleteRegion: "{% trans "Delete" %}",
+                    deleteRegion: "{% trans "Delete region" %}",
```
dasgarner: I think we ought to have a capital R for region where and on the prior lines. This would be consistent with our use of capital letters for naming "Xibo objects".

PR #754, opened 2019-12-18
Layout conversion task to only select schemaVersions < 2.

dasgarner:
xibosignage/xibo#2056

### DIFF
--- a/lib/Factory/LayoutFactory.php
+++ b/lib/Factory/LayoutFactory.php
@@ -636,6 +636,9 @@ class LayoutFactory extends BaseFactory
                     }
                 }
 
+                // Calculate the duration
+                $widget->calculateDuration($module, true);
+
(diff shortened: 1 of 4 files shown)


lib/Factory/LayoutFactory.php
```diff
@@ -636,6 +636,9 @@ public function loadByXlf($layoutXlf, $layout = null)
                     }
                 }
 
+                // Calculate the duration
+                $widget->calculateDuration($module, true);
```
dasgarner: @PeterMis I could do with running over this with you when you return. Thanks.

dasgarner:
jenkins: retest this please

lib/Factory/LayoutFactory.php
dasgarner: I think this is where the test failed - need to review.

PR #757, opened 2019-12-19
Revisit: Scheduling

ifarzana:


lib/Entity/Schedule.php
```diff
@@ -567,7 +567,7 @@ public function validate()
             throw new InvalidArgumentException(__('Repeat every must be a positive number'), 'recurrenceDetail');
         
         // Check and disallow Minute/Hourly repeats to be indefinite (or too long)
-        $twelveHoursInSeconds = 12 * 60 * 60;
+        /*$twelveHoursInSeconds = 12 * 60 * 60;
```
dasgarner: We can delete this validation completely I think?

lib/Controller/Schedule.php
```diff
@@ -331,14 +331,23 @@ function eventData()
                 $title = __('%s scheduled on %s', $row->campaign, $displayGroupList);
             }
 
+            if ($row->recurrenceType == 'Minute' || $row->recurrenceType == 'Hour') {
+                $title .= ', Repeats every ' . $row->recurrenceDetail . ' ' . $row->recurrenceType;
```
dasgarner: We will need this to be translatable I think, so:

`__(', Repeats every %s %s', $row->recurrenceDetail, $row->recurrenceType)` 

I think?

PR #769, opened 2020-01-09
Pagination support in stat search xibosignage/xibo#2059

ifarzana:


lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -258,8 +271,11 @@ public function getStats($filterBy = [])
 
         $result = new TimeSeriesMySQLResults($statement);
 
-        return $result;
-
+        return
```
dasgarner: Would it be better to add the `totalCount` as a property of the TimeSeriesMySQLResults class (and the Mongo one) rather than returning an array? It would make the interface concrete.

We could then have a getTotalCount() method on the TimeSeriesMySQLResults - don't forget to change the interface too.

lib/Controller/Stats.php
```diff
@@ -299,8 +299,8 @@ function displayProofOfPlayPage()
      */
     public function grid()
     {
-        $fromDt = $this->getSanitizer()->getDate('fromDt', $this->getSanitizer()->getDate('statsFromDt', $this->getDate()->parse()->addDay(-1)));
-        $toDt = $this->getSanitizer()->getDate('toDt', $this->getSanitizer()->getDate('statsToDt', $this->getDate()->parse()));
+        $fromDt = $this->getSanitizer()->getDate('fromDt');
```
dasgarner: If memory serves me correctly the form fields in the UI are called `statsFromDt` and `statsToDt`, meaning the "select a range" filter won't work after this change?

ifarzana: This endpoint is never used by any UI form. if we do above, we will always have value for fromDt and ToDt, which we do not want if there is any statDate. Our logic is 'Either fromDt/toDt or statDate should be provided' .

I might do like this way that if there is any statDate provided ignore fromDt/toDt?

dasgarner: If the UI does not call this method ever, then what you have done is great.
Please can you just make a comment on it and say that it is only ever access by the API?

ifarzana: I will do that thanks

PR #770, opened 2020-01-10
Brooks 2.1.3

PeterMis:


lib/Controller/DisplayGroup.php
```diff
@@ -581,6 +582,7 @@ public function add()
     public function edit($displayGroupId)
     {
         $displayGroup = $this->displayGroupFactory->getById($displayGroupId);
+        $preEditIsDynamic = $displayGroup->isDynamic;
```
dasgarner: can you use `->getOriginalProperty` for this?

lib/Controller/Region.php
```diff
@@ -666,6 +666,9 @@ public function preview($regionId)
             // No media to preview
             $this->getState()->extra['empty'] = true;
             $this->getState()->extra['text'] = __('Empty Region');
+        } catch (InvalidArgumentException $e) {
+            $this->getState()->extra['empty'] = true;
+            $this->getState()->extra['text'] = __('Please select a Playlist');
```
dasgarner: Because this might be triggered by other things in the future (potentially) can we say "Please correct the error with this Widget" ?

lib/Controller/DisplayGroup.php
```diff
@@ -582,10 +582,11 @@ public function add()
     public function edit($displayGroupId)
     {
         $displayGroup = $this->displayGroupFactory->getById($displayGroupId);
-        $preEditIsDynamic = $displayGroup->isDynamic;
+        $preEditIsDynamic = $displayGroup->getOriginalValue('isDynamic');
```
dasgarner: haha, i meant you didn't have to set it up here and that you could just use the getOriginalValue below (i.e. you don't need to store it)

NM though :-)

PR #776, opened 2020-01-14
Updated api/stats

ifarzana:
Pagination support in stat search
Search by >= statDate
Search by > statId

### DIFF
--- a/lib/XTR/ReportScheduleTask.php
+++ b/lib/XTR/ReportScheduleTask.php
@@ -126,8 +126,6 @@ class ReportScheduleTask implements TaskInterface
                 $rs = $this->reportScheduleFactory->getById($reportSchedule->reportScheduleId);
                 $rs->previousRunDt = $rs->lastRunDt;
                 $rs->lastRunDt = time();
-                $rs->isActive = 0;
-                $rs->save();
 
                 $this->log->debug('Last run date is updated to '. $rs->lastRunDt);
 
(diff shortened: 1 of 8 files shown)


lib/XTR/ReportScheduleTask.php
```diff
@@ -126,51 +126,63 @@ private function runReportSchedule()
...
+                    // Get the generated saved as report name
+                    $saveAs = $this->reportService->generateSavedReportName($reportSchedule->reportName, $reportSchedule->filterCriteria);
+
+
+                    // Run the report to get results
+                    $result =  $this->reportService->runReport($reportSchedule->reportName, $reportSchedule->filterCriteria, $reportSchedule->userId);
+                    $this->log->debug(__('Run report results: %s.', json_encode($result, JSON_PRETTY_PRINT)));
+
+                    //  Save the result in a json file
+                    $fileName = tempnam(sys_get_temp_dir(), 'reportschedule');
```
dasgarner: I know this hasn't been changed here, but I didn't notice this before.
We need to always write temporary files into the library temp folder and not the FS folder please.

lib/XTR/ReportScheduleTask.php
```diff
@@ -126,51 +126,63 @@ private function runReportSchedule()
...
+
+                } catch (\Exception $error) {
+                    $this->log->error('Error: ' . $error->getMessage());
+                    continue;
+                }
 
                 // Add the last savedreport in Report Schedule
                 $this->log->debug('Last savedReportId in Report Schedule: '. $savedReport->savedReportId);
                 $rs->lastSavedReportId = $savedReport->savedReportId;
+                $rs->isActive = 1;
```
dasgarner: We set it to 0 at the beginning and set it to 1 at the end if its failed, but doesn't the `continue` in the catch skip this step?

I think we need to remove the continue from the catch?

It also makes more logical sense to me to leave isActive at the top, and to set it to 0 in the catch (so that if we fail, we set it to 0)?

dasgarner: Sorry, I see the `continue` is necessary - but it still seems strange to me that we're saving it before we run it, and again after.

Can't we save one time at the end instead?

PR #783, opened 2020-01-17
Optimise XMDS submit stats xibosignage/xibo#2065

ifarzana:


lib/Xmds/Soap.php
```diff
@@ -1480,50 +1480,30 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
...
             }
 
-            $campaignId = 0;
             $layoutId = $node->getAttribute('layoutid');
 
-            // For a type "event" we have layoutid 0
-            // otherwise we should try and resolve the campaignId
             if ($type != 'event') {
-                try {
-                    // Handle the splash screen
```
dasgarner: Do we still handle only logging this message one time?

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -73,27 +77,65 @@ public function setDependencies($log, $date, $layoutFactory = null, $campaignFac
...
+                    $this->layoutCampaignIds[$statData['layoutId']] = $campaignId;
 
-        // Flatten the array
-        $data = [];
-        foreach ($statData as $stat) {
-            foreach ($stat as $field) {
-                $data[] = $field;
+                } catch (XiboException $error) {
+
+                    $this->log->error('Layout not found. Layout Id: '. $statData['layoutId']);
```
dasgarner: We used to only log this one time per layoutId - does this new code log it each time?

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -120,137 +127,231 @@ public function setDependencies($log, $date, $layoutFactory = null, $campaignFac
...
+        $this->stats[] = $statData;
+
+    }
+
+    /** @inheritdoc */
+    public function addStatFinalize()
+    {
+
         // Insert statistics
         $collection = $this->client->selectCollection($this->config['database'], $this->table);
```
dasgarner: I didn't notice this before, but I think we should throw an error out after we log here?

If the SubmitStats request returns a 200 response, the Player will delete the stats locally - but if the insert fails here, we should make the player try again I think? we do that be throwing an exception

I think the MySQL one does throw an exception, but worth checking.

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -63,6 +64,12 @@ class MongoDbTimeSeriesStore implements TimeSeriesStoreInterface
 
     private $periodTable = 'period';
 
+    private $stats = [];
```
dasgarner: A quick comment here would be useful, so that we know what these variables are intended for.
Also (nitpick) but I don't think the naming is quite right - for example `$widgetIds` actually contains widget names, right?

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -120,137 +127,231 @@ public function setDependencies($log, $date, $layoutFactory = null, $campaignFac
     /** @inheritdoc */
     public function addStat($statData)
     {
-        foreach ($statData as $k => $stat) {
 
-            $statData[$k]['statDate'] = new UTCDateTime($statData[$k]['statDate']->format('U') * 1000);
-            $statData[$k]['start'] = new UTCDateTime($statData[$k]['fromDt']->format('U') * 1000);
-            unset($statData[$k]['fromDt']);
+        $statData['statDate'] = new UTCDateTime($statData['statDate']->format('U') * 1000);
```
dasgarner: we could take this opportunity to comment why we're doing this, and why we're doing the unset

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -120,137 +127,231 @@ public function setDependencies($log, $date, $layoutFactory = null, $campaignFac
...
 
                 try {
-                    $layout = $this->layoutFactory->getByParentCampaignId($stat['campaignId']);
 
-                    $this->log->debug('Found layout : '. $stat['layoutId']);
+                    // Get the layout campaignId - this can give us a campaignId of a layoutId which id was replaced when draft to published
+                    $campaignId = $this->layoutFactory->getCampaignIdFromLayoutHistory($statData['layoutId']);
+
+                    // Get the layout by campaignId
+                    $layout = $this->layoutFactory->getByParentCampaignId($campaignId);
```
dasgarner: I think you could have a `layoutFactory->getByLayoutHistory($layoutId)` method which combined these two queries into 1?

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -120,137 +127,231 @@ public function setDependencies($log, $date, $layoutFactory = null, $campaignFac
...
-                continue;
-            }
+            // SET widgetName
+            $statData['widgetName'] = $widgetName;
 
-            $statData[$k]['displayName'] = $display->display;
+        }
+
+        // Display name
+        try {
```
dasgarner: I think we pass this in from the Soap.php using `$this->display->displayId` - which we could change to `$this-display` and completely remove this lookup?

lib/Storage/MongoDbTimeSeriesStore.php
ifarzana: $this->store->isolated(..)  does throw PDOException

dasgarner: I think it does actually - if its not a reconnect error it catches and rethrows

PR #784, opened 2020-01-17
Scheduled Report Bugfix

ifarzana:


lib/XTR/ReportScheduleTask.php
```diff
@@ -122,51 +122,62 @@ private function runReportSchedule()
...
+                } catch (\Exception $error) {
+                    $rs->isActive = 0;
+                    $rs->message = $error->getMessage();
+                    $this->log->error('Error: ' . $error->getMessage());
+                }
 
-                // Add the last savedreport in Report Schedule
-                $this->log->debug('Last savedReportId in Report Schedule: '. $savedReport->savedReportId);
-                $rs->lastSavedReportId = $savedReport->savedReportId;
+                // Finally save schedule report
```
dasgarner: Should setting the message to full be inside the try?

PR #787, opened 2020-01-20
2.2.3 Optimise XMDS

ifarzana:


lib/Xmds/Soap.php
```diff
@@ -1476,48 +1476,32 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
             $duration = $node->getAttribute('duration');
             $count = $node->getAttribute('count');
 
-            // if fromdt and to dt are same then ignore them
-            if ($fromdt == $todt) {
-                $this->getLog()->debug('Fromdt (' . $fromdt. ') and ToDt (' . $todt. ') are same. ');
+            if ($fromdt == '' || $todt == '' || $type == '') {
+                $this->getLog()->error('Stat submitted without the fromdt, todt or type attributes.');
```
dasgarner: Because there is nothing we can do about this, I think we should downgrade these log levels to `info` - it saves filling the log with errors we cannot do anything about.

lib/Factory/LayoutFactory.php
```diff
@@ -283,6 +283,41 @@ public function getCampaignIdFromLayoutHistory($layoutId)
...
+        $row = $this->getStore()->select('SELECT campaignId FROM `layouthistory` WHERE layoutId = :layoutId LIMIT 1', ['layoutId' => $layoutId]);
+
+        if (count($row) <= 0) {
+            throw new NotFoundException(__('Layout does not exist'));
+        }
+
+        $campaignId = intval($row[0]['campaignId']);
+
+        // Get a Layout by its Layout Specific Campaign OwnerId
+        $layouts = $this->query(null, array('disableUserCheck' => 1, 'ownerCampaignId' => $campaignId, 'excludeTemplates' => -1, 'retired' => -1));
```
dasgarner: If we add a new parameter to the `query` method here called `layoutHistoryId`, then we could eliminate the first query in this method and do it all in one?

lib/Xmds/Soap.php
```diff
@@ -1552,21 +1536,15 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
...
                 }
             }
-            
-            $tag = $node->getAttribute('tag');
 
+            $tag = $node->getAttribute('tag');
             if ($tag == 'null')
                 $tag = null;
 
-            if ($fromdt > $todt) {
```
dasgarner: Seems like this logic has been lost during the changes? It makes sense to have this happen when we are checking the other date/time bits above?

lib/Xmds/Soap.php
```diff
@@ -1994,7 +1976,7 @@ protected function checkBandwidth($displayId)
         // Uncomment to enable auditing.
         //$this->logProcessor->setDisplay(0, true);
 
-        $this->display = $this->displayFactory->getById($displayId);
+        $this->display = $this->displayFactory->getById($displayId, true);
```
dasgarner: So I can see why you did this and I understand that it makes the tag lookup more elegant for submitting stats - however `checkBandwidth` is called for every XMDS method, so adding this here means we have made **all** the XMDS calls more complicated to save a lookup in the stats call.

This probably isn't a good trade-off in this case.

PR #794, opened 2020-01-24
Moved screenshot label to the right-bottom

peppelauro:
Usually the right-bottom corner is a better place to set the label's screenshot.

### DIFF
--- a/lib/Controller/Display.php
+++ b/lib/Controller/Display.php
@@ -1366,11 +1366,13 @@ class Display extends Base
         $date = $display->getCurrentScreenShotTime($this->pool);
 
         if ($date != '') {
+            $w = $img->getWidth();
+            $h = $img->getHeight();
             $img
-                ->rectangle(0, 0, 110, 15, function($draw) {
+                ->rectangle($w - 110, $h - 15, 110, 15, function($draw) {
                     $draw->background('#ffffff');
                 })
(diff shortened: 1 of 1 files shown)


dasgarner:
Thank you for your contribution - welcome to the project.

The feature request for the timestamp specified that it should be in the top left corner, but it seems to us that it doesn't matter where you put it, there is a chance it will be in the way. :-)

When you say "usually" do you mean there is some standard for doing this? If so can you link us to it?

Thanks,
Dan

peppelauro:
Hi @dasgarner 
thank you for your great job on Xibo.
I don't know if there is some standard but thinking as a "western language user" (reading from top to bottom and from left to right) and working with many graphic designer, I've learned some rules about the weight and the position of the info into an image (label, text and other...).
So, talking about the timestamp into the screenshot, It's not the main info in the most of the case, the main info is the image self. The timestamp is a secondary level info, and placing it into the right-bottom corner has more sense. (All the reasoning is very opinionated and not the holy thruth).

peppelauro:
Obviously my final target is to give the possibility to the user to choose the position of the label.

I will try to open a new pull request about that :)

dasgarner:
I'm going to close this out as it seems wherever we put the information someone will not like it :-)

PR #796, opened 2020-01-27
Playlist Usage Report xibosignage/xibo#2037

ifarzana:


views/playlist-form-usage.twig
```diff
@@ -0,0 +1,100 @@
+{#
+/*
+ * Spring Signage Ltd - http://www.springsignage.com
```
dasgarner: New file should have new copyright (Xibo Signage and 2020)

PR #798, opened 2020-01-29
Proof of Play: Track engagements  xibosignage/xibo#2094

ifarzana:


lib/Xmds/Soap.php
```diff
@@ -1487,6 +1487,24 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
             $type = $node->getAttribute('type');
             $duration = $node->getAttribute('duration');
             $count = $node->getAttribute('count');
+            $engagements = [];
+
+            foreach ($node->childNodes as $nodeElements) {
+                /* @var \DOMElement $nodeElements */
+                if ($nodeElements->nodeName == "engagements") {
```
dasgarner: `""`'s :cry:

lib/Xmds/Soap.php
```diff
@@ -1623,6 +1641,7 @@ protected function doSubmitStats($serverKey, $hardwareKey, $statXml)
                 'widgetId' => (int) $widgetId,
                 'duration' => (int) $duration,
                 'count' => ($count != '') ? (int) $count : 1,
+                'engagements' => (count($engagements) > 0) ? json_encode($engagements) : '[]',
```
dasgarner: Does this mean we are `json_encoding` and then on the other side in the case of mongo `json_decoding` again? 
It would make more sense (I think) to pass the array here, and then `json_encode` only for MySQL?

lib/Xmds/Soap.php
dasgarner: I know its pedantic, but we use `'` quotes not double quotes `"` (there are even some above and below these new lines) :smile:

PR #810, opened 2020-02-06
Bugfix Release2.2.3 RegEx

ifarzana:


lib/Factory/BaseFactory.php
```diff
@@ -290,24 +291,24 @@ public function nameFilter($tableName, $tableColumn, $terms, &$body, &$params)
                 if (substr($searchName, 0, 1) == '-') {
                     if ($i == 1) {
                         $body .= " AND ( $tableAndColumn NOT RLIKE (:search$i) ";
-                        $params['search' . $i] = preg_quote(ltrim(($searchName), '-'));
+                        $params['search' . $i] = empty($regex) ? preg_quote(ltrim(($searchName), '-')) : ltrim(($searchName), '-');
```
dasgarner: Do we need to use `empty` for a true/false parameter?

lib/Factory/BaseFactory.php
```diff
@@ -320,8 +321,9 @@ public function nameFilter($tableName, $tableColumn, $terms, &$body, &$params)
      * @param string $operator exactTags passed from factory, determines if the search is LIKE or =
      * @param string $body Current SQL body passed by reference
      * @param array $params Array of parameters passed by reference
+     * @param bool $regex flag to match against a regex pattern
      */
-    public function tagFilter($tags, $operator, &$body, &$params)
+    public function tagFilter($tags, $operator, &$body, &$params, $regex = false)
```
dasgarner: No regex parameter on tagFilter

lib/Factory/DisplayFactory.php
```diff
@@ -314,7 +314,12 @@ public function query($sortOrder = null, $filterBy = [])
         // Filter by Display Name?
         if ($this->getSanitizer()->getString('display', $filterBy) != null) {
             $terms = explode(',', $this->getSanitizer()->getString('display', $filterBy));
-            $this->nameFilter('display', 'display', $terms, $body, $params);
+
+            $useRegexForName = false;
+            if ($this->getSanitizer()->getCheckbox('useRegexForName') == 1) {
```
dasgarner: Can this be refactored to be shorter?

```$useRegexForName = ($this->getSanitizer()->getCheckbox('useRegexForName') == 1);```

File: lib/Factory/DisplayGroupFactory.php [hunk 4]
```diff
@@ -337,7 +337,12 @@ public function query($sortOrder = null, $filterBy = [])
         // Filter by DisplayGroup Name?
         if ($this->getSanitizer()->getString('displayGroup', $filterBy) != null) {
             $terms = explode(',', $this->getSanitizer()->getString('displayGroup', $filterBy));
-            $this->nameFilter('displaygroup', 'displayGroup', $terms, $body, $params);
+
+            $useRegexForName = false;
+            if ($this->getSanitizer()->getCheckbox('useRegexForName') == 1) {
```
dasgarner: Same question as in display factory

views/displaygroup-page.twig
```diff
@@ -57,6 +57,9 @@
 
                             {% set title %}{% trans "Exact match?" %}{% endset %}
                             {{ inline.checkbox("exactTags", title) }}
+
```
dasgarner: Can we make this change for 2.3 instead of 2.2? We want to avoid changing functionality in 2.2.3 (as it is a bug fix patch only)

lib/Factory/BaseFactory.php
```diff
@@ -264,8 +264,9 @@ protected function parseComparisonOperator($variable)
      * @param array $terms An Array exploded by "," of the search names
      * @param string $body Current SQL body passed by reference
      * @param array $params Array of parameters passed by reference
+     * @param bool $regex flag to match against a regex pattern
      */
-    public function nameFilter($tableName, $tableColumn, $terms, &$body, &$params)
+    public function nameFilter($tableName, $tableColumn, $terms, &$body, &$params, $regex = false)
```
dasgarner: Should we call this `useRegex` to make it clear what it does?

PR #813, opened 2020-02-07
BugFix2.2.3

ifarzana:


lib/Report/DistributionReport.php
```diff
@@ -257,6 +257,7 @@ public function setReportScheduleFormData()
     /** @inheritdoc */
     public function generateSavedReportName($filterCriteria)
     {
+        $saveAs = '';
```
dasgarner: I might be wrong, sorry if I am, but I think we can still get to the end of this method without setting `$saveAs` to be anything other than empty string?

lib/Report/SummaryReport.php
```diff
@@ -244,6 +244,7 @@ public function setReportScheduleFormData()
     public function generateSavedReportName($filterCriteria)
     {
 
+        $saveAs = '';
```
dasgarner: Same as other instance

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -533,46 +557,27 @@ public function getStats($filterBy = [])
...
-                    ]
-                ],
-            ];
-            $totalCursor = $collection->aggregate($totalQuery);
+            // Total
+            $result->totalCount = $total;
 
         } catch (\MongoDB\Exception\RuntimeException $e) {
             $this->log->error($e->getMessage());
+            throw new GeneralException($e->getMessage());
```
dasgarner: Do we really want to pass e->getMessage out to the user? Will that mean anything to them at this point?

We could pass them a user friendly message instead?

```
 throw new GeneralException(__('Sorry we encountered an error getting Proof of Play data, please consult your administrator'));
```

lib/Report/DistributionReport.php
ifarzana: yes, we can get to end and no error is thrown.

ifarzana: does that mean I can remove the line $saveAs = '';  ?

PR #818, opened 2020-02-10
Fix event 0 duration when DayPart start/end time are the same

dasgarner:
xibosignage/xibo#2102


dasgarner:
I am not sure how this test could have passed in 2.3, but it did!

PR #839, opened 2020-02-18
3.0 fixes

PeterMis:
Fixes for XTR, Tasks and Reports
Auto-reload Twig enabled when in test mode
Some code cleanup

### DIFF
--- a/lib/Report/SummaryReport.php
+++ b/lib/Report/SummaryReport.php
@@ -303,6 +303,7 @@ class SummaryReport implements ReportInterface
     /** @inheritDoc */
     public function getResults($filterCriteria, Request $request)
     {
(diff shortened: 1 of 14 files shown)


lib/Report/SummaryReport.php
```diff
@@ -303,6 +303,7 @@ public function getSavedReportResults($json, $savedReport)
     /** @inheritDoc */
     public function getResults($filterCriteria, Request $request)
     {
+        // This function is called by both web ui and xtr,  XTR will have filters in $filterCriteria array and web in $request params
```
dasgarner: Maybe we should handle this one step up? So whatever calls getResults makes the decision about which array of params to pass down?

PeterMis: Agreed, attempt to achieve that in the next commit.

PR #843, opened 2020-02-25
Moar Fixes

PeterMis:
- Added our ouath2 library back, now with updated oauth2 client
- Fixed SSSP Player software issues
- Added http cache Middleware and helper
- Fixed Display Profile config fields, updated Sanitizer with a new function (hasParam())
- Fixed interrupt Notification redirect and "Continue..." url


lib/Controller/PlayerSoftware.php
```diff
@@ -450,9 +465,12 @@ public function getSsspInstall(Request $request, Response $response)
...
             $versionInformation = $this->playerVersionFactory->getByMediaId($mediaId);
 
-            $this->outputSsspXml($versionInformation->version . '.' . $versionInformation->code, $media->fileSize);
+            $xml = $this->outputSsspXml($versionInformation->version . '.' . $versionInformation->code, $media->fileSize);
+            $response = $response
+                ->withHeader('Content-Type', 'application/xml')
+                ->write($xml);
         } else {
-            $response->withStatus(404);
+            return $response->withRedirect('/notFound');
```
dasgarner: So I think this will cause an issue because Tizen requires a completely empty document with a 404 error code.

lib/Controller/PlayerSoftware.php
```diff
@@ -484,10 +502,10 @@ public function getSsspInstallDownload(Request $request, Response $response)
 
             // Create a widget from media and call getResource on it
             $widget = $this->moduleFactory->createWithMedia($media);
-            $response = $widget->getResource($request, $response);
+            $response = $widget->download($request, $response);
 
         } else {
-            $response->withStatus(404);
+            return $response->withRedirect('/notFound');
```
dasgarner: Same as above

PR #850, opened 2020-03-03
2.3.1

ifarzana:


lib/Controller/Stats.php
```diff
@@ -541,6 +541,7 @@ public function export()
         $fromDt = $this->getSanitizer()->getDate('fromDt');
         $toDt = $this->getSanitizer()->getDate('toDt');
         $displayId = $this->getSanitizer()->getInt('displayId');
+        $isExport = $this->getSanitizer()->getInt('isExport', 1);
```
dasgarner: We would normally use the `getCheckbox()` method here as it returns for On/1/true

```suggestion
        $isExport = $this->getSanitizer()->getCheckbox('isExport');
```

lib/Controller/Stats.php
```diff
@@ -582,56 +583,70 @@ public function export()
...
+        $resultSet = $this->timeSeriesStore->getStats([
+            'fromDt' => $fromDt,
+            'toDt' => $toDt,
             'displayIds' => $displayIds,
         ]);
 
-        $out = fopen('php://output', 'w');
-        fputcsv($out, ['Stat Date', 'Type', 'FromDT', 'ToDT', 'Layout', 'Display', 'Media', 'Tag', 'Duration', 'Count', 'Engagements']);
+        // We only export if isExport is 1 else we return the count of stats
+        if (!empty($isExport)) {
```
dasgarner: To get a count of stats we ought to have a separate endpoint to call - probably a web only route in routes-web

Having it triggered by a special parameter means that this method now has two different purposes, which is not RESTful.

views/statistics-form-export.twig
```diff
@@ -42,7 +44,27 @@
...
     </div>
+{% endblock %}
+
+{% block javaScript %}
+<script type="text/javascript">
+
+    // Bind a click event to each table rows + button (span)
+    $("#fromDtLink").on("click", function() {
+
+        console.log( 'clicked ');
```
dasgarner: i'm pretty certain this shouldn't be in here? :-)

views/stats-proofofplay-page.twig
```diff
@@ -69,6 +69,60 @@
...
+            var fromDtLink;
+            var toDtLink;
+            var displayId;
+
+            $(dialog).on("change", function() {
+                fromDtLink =  $(dialog).find("#fromDt").val();
+                toDtLink =  $(dialog).find("#toDt").val();
+                displayId =  $(dialog).find("#displayId").val();
+
+                if (
```
dasgarner: you can remove the empty if block by negating this?

`if (!(....))`

views/stats-proofofplay-page.twig
```diff
@@ -69,6 +69,60 @@
...
+
+                } else {
+
+                    $(dialog).find(".total-stat").remove();
+                    $(dialog).find('.save-button').prop('disabled', true);
+                    $(dialog).find(".loading-overlay").show();
+
+                    $.ajax({
+                        type: "get",
+                        url: "{{ urlFor("stats.export") }}",
```
dasgarner: As with prior comment, I think this should be a separate route so that the request is "restful" and only does one thing

lib/Controller/Stats.php
```diff
@@ -582,56 +583,70 @@ public function export()
         }
 
         // Get result set
-        $resultSet =  $this->timeSeriesStore->getStats([
-            'fromDt'=> $fromDt,
-            'toDt'=> $toDt,
+        $resultSet = $this->timeSeriesStore->getStats([
```
dasgarner: If I am reading this correctly, we have to run the entire query in order to get the count?

We can certainly make that more elegant in MySQL by just running the `COUNT()` statement with filters - perhaps there is nothing we can do in Mongo though?

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -161,10 +163,12 @@ public function addStat($statData)
                     $this->mediaItems[$statData['mediaId']] = $media;
 
                 } catch (NotFoundException $error) {
-                    // Media not found ignore and log the stat
-                    $this->log->error('Media not found. Media Id: '. $statData['mediaId'] .',Layout Id: '. $statData['layoutId']
-                        .', FromDT: '.$statData['start'].', ToDt: '.$statData['end'].', Type: '.$statData['type']
-                        .', Duration: '.$statData['duration'] .', Count '.$statData['count']);
+
```
dasgarner: This is good - but can we go one step further do you think?

If we have `$this->mediaItemsNotFound`, we could not do the look up next time and save ourselves lots of processing? This also makes me wonder if we should cache the result as well - if we have 100 stats and they are all the same mediaId, we make this call 100 times and it could be 1 time?

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -354,150 +354,30 @@ public function getStats($filterBy = [])
...
-            LEFT OUTER JOIN display
-            ON stat.DisplayID = display.DisplayID
-            LEFT OUTER JOIN layout
-            ON layout.LayoutID = stat.LayoutID
-            LEFT OUTER JOIN media
-            ON media.mediaID = stat.mediaID
-            LEFT OUTER JOIN widget
-            ON widget.widgetId = stat.widgetId
-         WHERE 1 = 1 ';
+        $sql = ' SELECT stat.statId FROM stat WHERE 1 = 1 ';
```
dasgarner:
```suggestion
        $sql = ' SELECT COUNT(*) FROM `stat` WHERE 1 = 1 ';
```

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -354,150 +354,30 @@ public function getStats($filterBy = [])
...
-                // we wont get any match as we store layoutspecific campaignid in stat
-                $body .= ' AND stat.campaignId = '. $campaignId;
-            }
-        }
-
-        $body .= " ORDER BY stat.statId ";
-
-
         // Total count
         $resTotal = $this->store->select('
```
dasgarner:
```suggestion
        $resTotal = $this->store->select($sql, $params);
```

views/stats-proofofplay-page.twig
```diff
@@ -69,6 +69,59 @@
...
 {% block javaScript %}
     <script type="text/javascript">
+
+        function statsExportFormSelect(dialog) {
+
+            var fromDtLink;
+            var toDtLink;
+            var displayId;
+
+            $(dialog).on("change", function() {
```
maurofmferrao: Using the "on change" event for the main dialog object can lead to some problems, maybe we need to specify the elements we want to be triggered by this listener, something like:

`$(dialog).find("input, select").on("change", function() { ...`

PR #851, opened 2020-03-04
More fixed tests

PeterMis:


lib/Controller/Fault.php
```diff
@@ -209,21 +210,13 @@ public function collect(Request $request, Response $response)
...
             header("X-Sendfile: $tempFileName");
-            $response->withStatus(200);
         }
         // Send via Nginx X-Accel-Redirect?
         if ($this->getConfig()->getSetting('SENDFILE_MODE') == 'Nginx') {
             header("X-Accel-Redirect: /download/temp/" . basename($tempFileName));
-            $response->withStatus(200);
         }
 
-        // Return the file with PHP
```
dasgarner: If you delete this, then it won't return the file for a CMS with sendfile mode off?

I wonder if you might want something like:

```
use GuzzleHttp\Psr7\Stream;
$response = $response->withBody(new Stream(fopen($tempFileName, 'r')));
```

dasgarner: We most likely should use the `withHeader()` also, so that we don't modify the output directly

lib/Middleware/State.php
```diff
@@ -238,8 +238,8 @@ public static function setState(App $app, Request $request): Request
 
         // Configure logging
         if (Environment::isForceDebugging() || strtolower($mode) == 'test') {
-            error_reporting(E_ALL);
-            ini_set('display_errors', 1);
+           // error_reporting(E_ALL);
```
dasgarner: Did you mean for this to be commented out in the PR?

PeterMis: oops, no I didn't, will bring this back in the next PR, thanks.

PR #861, opened 2020-03-12
Fix single quote/double quotes issue

Inglebard:
This avoid issue with French translations in html title attribute. (Specially in display-page.twig)
I also try to change it for all class attribute.
Exception for class inside data- attribute.

### DIFF
--- a/views/dashboard-status-page.twig
+++ b/views/dashboard-status-page.twig
@@ -738,7 +738,7 @@
                         else
                             icon = "fa-cloud-download";
 
-                        return "<span class='fa " + icon + "' title='" + (row.statusDescription) + "'></span>";
+                        return '<span class="fa ' + icon + '" title="' + (row.statusDescription) + '"></span>';
                     }
                 },
                 {"data": "licensed", "render": dataTableTickCrossColumn},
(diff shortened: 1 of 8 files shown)


Inglebard:
Hi,
I understand.
I don't use git/github a lot, sorry for the bad PR.

I undo/force push the commit with the only necessary changes.
However, I think it should be great to normalize quote and double-quotes for all html attribute and/or escape necessary characters.

maurofmferrao:
> Hi,
> I understand.
> I don't use git/github a lot, sorry for the bad PR.
> 
> I undo/force push the commit with the only necessary changes.
> However, I think it should be great to normalize quote and double-quotes for all html attribute and/or escape necessary characters.

Thank you for the quick PR changes!

It wasn't really a bad PR per se, it just had some tweaks that had to be made. I will test your recent changes today ( that look really good at first sight ) and merge them if all passes :)

We also agree that the HTML attributes should all be normalized for JS/Twig files, we will keep that in mind for future developments/versions.

Inglebard:
Hi,

Do you know when this change will be integrated in a release ?

dasgarner:
You should see them in the 3.0.x set of alpha releases, with a beta release to follow along soon.

PR #865, opened 2020-03-17
2.3.2

ifarzana:


lib/Entity/Schedule.php
```diff
@@ -531,9 +531,21 @@ public function validate()
...
+                // Hack : If this is an interrupt, check that the column is a SMALLINT and if it isn't alter the table
+                $sql = 'SELECT DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE table_name = :table_name AND COLUMN_NAME = :column_name';
+                $params = ['table_name' => 'schedule', 'column_name' => 'shareOfVoice' ];
+                $results = $this->store->select($sql, $params);
+
+                if (count($results) > 0) {
+                    $dataType = $results[0]['DATA_TYPE'];
+
+                    if ($dataType !== 'smallint') {
+                        $this->store->update('ALTER TABLE `schedule` MODIFY `shareOfVoice` SMALLINT', []);
```
dasgarner: Looking at this I had an idea - could we use this detection as an opportunity to convert any existing interrupt schedules?

We could say ```UPDATE `schedule` SET shareOfVoice = 3600 * (shareOfVoice / 100) WHERE shareOfVoice > 0```

I think that would convert any existing percentages to their correct seconds/hour? For example 10% share of voice would be 3600 * (10/100) = 360 ? What do you think?

views/schedule-form-add.twig
```diff
@@ -123,9 +123,18 @@
                             </div>
                         </div>
 
-                        {% set title %}{% trans "Share of Voice" %}{% endset %}
-                        {% set helpText %}{% trans "The percentage of time this Layout should be shown." %}{% endset %}
-                        {{ forms.number("shareOfVoice", title, 0, helpText, "interrupt-control") }}
+                        <div class="form-group interrupt-control" style="display: block;">
+                            <label class="col-sm-2 control-label" for="shareOfVoice" accesskey="">Share of Voice</label>
```
dasgarner: These fields need to be translated please :-)

PR #868, opened 2020-03-18
Get stats by displayIds in stat search endpoint  xibosignage/xibo#2173

ifarzana:


lib/Controller/Stats.php
```diff
@@ -341,30 +352,47 @@ public function grid()
...
+           if (count($displays) <= 0) {
+               $displayIds = $displaysAccessible;
+           } else {
+               foreach ($displays as $key => $id) {
+
+                   if (!in_array($id, $displaysAccessible)) {
+
+                       unset($displays[$key]);
+
+                       if (count($displays) <= 0 ) {
```
dasgarner: Should this be outside the `foreach`?

lib/Controller/Stats.php
```diff
@@ -341,30 +352,47 @@ public function grid()
...
+                           $displayIds = [-1];
+                       }
+                   }
+
+                   else {
```
dasgarner: can we get this else formatted up under the closing `}`

dasgarner:
```suggestion
                   } else {
```

PR #871, opened 2020-03-19
Cherrypick2.3.2

ifarzana:


lib/Controller/Stats.php
```diff
@@ -346,6 +356,7 @@ public function grid(Request $request, Response $response)
         $type = strtolower($sanitizedQueryParams->getString('type'));
 
         $displayId = $sanitizedQueryParams->getInt('displayId');
+        $displays = $sanitizedQueryParams->getIntArray('displayIds');
```
PeterMis: I think we need to provide default value here - an empty array.
That's most likely the reason why the tests failed.

PR #873, opened 2020-03-20
Interactive signage

PeterMis:
Drawer - BE
Actions - BE & part of FE
Updated Sanitizer, targets our xibosignage support repo now
Added url_decode filter to Twig


modules/audio-form-edit.twig
```diff
@@ -31,43 +31,54 @@
 {% block formHtml %}
     <div class="row">
         <div class="col-md-12">
+            <ul class="nav nav-tabs" role="tablist">
+                <li class="active"><a href="#generalTab" role="tab" data-toggle="tab"><span>{% trans "General" %}</span></a></li>
+                <li><a href="#actionTab" role="tab" data-toggle="tab" class="layout-designer-only"><span>{% trans "Action" %}</span></a></li>
```
dasgarner: I don't think we want to add these to each Widget form manually if we can avoid it - there will be loads of custom widgets, etc which will need upgrading as a result.

I imagined we would dynamically add the tab to each form using JavaScript - shall we have a chat about this?

ui/src/designer/properties-panel.js
```diff
@@ -205,6 +205,11 @@ PropertiesPanel.prototype.render = function(element, step) {
         // Append layout html to the main div
         self.DOMObject.html(html);
 
+        // Hide any elements with layout-designer-only class, hides actions tab in Playlist editor
+        if(app.mainObjectType === 'playlist') {
```
dasgarner: Linked to my prior comment - I think we'd be dynamically adding the actions grid and therefore not doing this (this might even be where we add it?)

views/layout-form-background.twig
```diff
@@ -38,63 +38,75 @@
 
 {% block formHtml %}
     <div class="row">
+        <ul class="nav nav-tabs" role="tablist">
+            <li class="active"><a href="#general" role="tab" data-toggle="tab"><span>{% trans "General" %}</span></a></li>
+            <li><a href="#actionTab" role="tab" data-toggle="tab"><span>{% trans "Action" %}</span></a></li>
```
dasgarner: We can probably add this one through the back end as you have done here (and the Region one) as that does not have any knock on implications for Custom stuff - I think?

PR #874, opened 2020-03-23
2.3.4

ifarzana:


dasgarner:
Sorry but this isn't what was meant in the issue - the restriction on Hourly repeating schedules should remain on the Month/Week views, but be relaxed in the Day view.

This PR relaxes it also on Month/Week?

PR #878, opened 2020-03-31
3.0.0 alpha

PeterMis:
Removed dateService
Switched jenssegers to Carbon2 
Fixed headers on library fontCss and Library download
Updated Sanitizer
Fixed PHP warnings in Calendar and Ticker Widgets
Removed old Sanitizer service
General cleanup


lib/Controller/Clock.php
```diff
@@ -89,7 +88,7 @@ public function clock(Request $request, Response $response)
         $this->session->refreshExpiry = false;
 
         if ($request->isXhr() || $this->isApi($request)) {
-            $output = $this->getDate()->getLocalDate(null, 'H:i T');
+            $output = Carbon::createFromTimestamp(time())->format('H:i T');
```
dasgarner: I wonder if it would be good practice to use `Carbon::now()` instead?

lib/Controller/Layout.php
```diff
@@ -689,7 +688,7 @@ function editBackground(Request $request, Response $response, $id)
     function deleteForm(Request $request, Response $response, $id)
     {
         $layout = $this->layoutFactory->getById($id);
-
+$this->getLog()->debug('LAYOUT IN DELETE FORM ' . json_encode($layout));
```
dasgarner: Did you mean to leave these in here?

lib/Controller/Playlist.php
```diff
@@ -1064,6 +1064,7 @@ public function widgetGrid(Request $request, Response $response)
     {
         $this->getState()->template = 'grid';
         $sanitizedParams = $this->getSanitizer($request->getParams());
+        $dateHelper = new DateFormatHelper();
```
dasgarner: Just a note as I go - this could probably be `static`

lib/Controller/Task.php
```diff
@@ -490,13 +488,13 @@ public function run(Request $request, Response $response, $id)
             $task->lastRunStatus = \Xibo\Entity\Task::$STATUS_ERROR;
         }
 
-        $task->lastRunDt = $this->getDate()->getLocalDate(null, 'U');
+        $task->lastRunDt = Carbon::createFromTimestamp(time())->format('U');
```
dasgarner: The more I look through the more I think replacing this with:


```suggestion
        $task->lastRunDt = Carbon::now()->format('U');
```

lib/Entity/Layout.php
```diff
@@ -625,7 +618,7 @@ public function load($options = [])
         }
 
         $this->getLog()->debug(sprintf('Loading Layout %d with options %s', $this->layoutId, json_encode($options)));
-
+$this->getLog()->debug('LAYOUT LOAD IS ' . json_encode($this));
```
dasgarner: Did you mean to leave these here?

lib/Factory/ActionFactory.php
```diff
@@ -45,7 +45,7 @@ class ActionFactory  extends BaseFactory
      * Construct a factory
      * @param StorageServiceInterface $store
      * @param LogServiceInterface $log
-     * @param SanitizerServiceInterface $sanitizerService
+     * @param SanitizerService $sanitizerService
```
dasgarner: I'm not sure why we're changing this? The Interface is still there, right?

dasgarner: I get this now!

PR #883, opened 2020-04-02
Switch all HLS widgets to use Edge Browser.

dasgarner:


dasgarner:
This didn't go anywhere, closing.

PR #893, opened 2020-04-08
3.0.0 alpha

PeterMis:
Updated Interactive Actions BE
Implemented System User
Minor adjustments here and there :)


db/migrations/20200401121544_add_system_user_setting_migration.php
```diff
@@ -0,0 +1,48 @@
...
+    /** @inheritdoc */
+    public function change()
+    {
+        if (!$this->fetchRow('SELECT * FROM `setting` WHERE setting = \'SYSTEM_USER\'')) {
+
+            // get System User
+            $getSystemUserQuery = $this->query('SELECT userId FROM user WHERE userTypeId = 1 ORDER BY userId LIMIT 1');
+            $getSystemUserResult = $getSystemUserQuery->fetchAll(PDO::FETCH_ASSOC);
+
+            $userId = $getSystemUserResult[0]['userId'];
```
dasgarner, line 36: I know this would be odd (very odd) - but what if there aren't any users?
We probably shouldn't fail to upgrade, even though no-one can login.

lib/Controller/Action.php
```diff
@@ -248,17 +256,28 @@ public function addForm(Request $request, Response $response, string $source, in
...
+
         $this->getState()->template = 'action-form-add';
         $this->getState()->setData([
             'sourceObject' => $sourceObject,
             'source' => $source,
             'id' => $id,
-            'regions' => $layout->regions
+            'regions' => $layout->regions,
+            'widgets' => $widgets,
+            'layoutCodes' => $this->layoutFactory->getLayoutCodes()
```
dasgarner, line 280: This could potentially be 1000s of Layouts?
Would it be better for this to be a dynamic select2 with paging?

lib/Controller/User.php
```diff
@@ -925,6 +925,11 @@ public function delete(Request $request, Response $response, $id)
     {
         $user = $this->userFactory->getById($id);
 
+        // System User
+        if ($user->userId == $this->getConfig()->getSetting('SYSTEM_USER')) {
+            throw new InvalidArgumentException('This User is set as System User and cannot be deleted.', 'userId');
```
dasgarner, line 930: Translations

lib/Entity/User.php
```diff
@@ -805,6 +805,12 @@ public function reassignAllTo($user)
             'oldUserId' => $this->userId
         ]);
 
+        // Reassign oAuth Clients
```
dasgarner, line 808: This is interesting, although I am not convinced that this does not present a security problem (essentially granting API access to someone else by virtue of deleting a user).

It might be better here for us to delete the oauth_client - what do you think?

lib/Entity/User.php
```diff
@@ -847,6 +853,15 @@ public function validate()
         catch (NotFoundException $e) {
             throw new InvalidArgumentException(__('Selected home page does not exist'), 'homePageId');
         }
+
+        // System User
+        if ($this->userId == $this->configService->getSetting('SYSTEM_USER') &&  $this->userTypeId != 1) {
+            throw new InvalidArgumentException('This User is set as System User and needs to be super admin', 'userId');
```
dasgarner, line 859: Translation

lib/Factory/LayoutFactory.php
```diff
@@ -1660,6 +1682,15 @@ public function createNestedPlaylistWidgets($widgets, $combined, &$playlists)
         return $playlists;
     }
 
+    /**
+     * Get all Codes assigned to Layouts
+     * return string[]
+     */
+    public function getLayoutCodes()
```
dasgarner, line 1689: As I've indicated above, my feeling is that this will give us issues if the number of Layouts with codes in a CMS expands above a small handful.

I agree with separating this out into its own DB call (no need to get all the other Layout details), but this ought to implement paging too, so that it can be used in a select2.

lib/Factory/UserFactory.php
```diff
@@ -189,14 +194,11 @@ public function getDoohUsers()
...
+     * @throws NotFoundException
      */
     public function getSystemUser()
     {
-        $user = $this->create();
-        $user->userId = 1;
-        $user->userName = 'system';
-        $user->userTypeId = 1;
-        $user->email = $this->configService->getSetting('mail_to');
+        $user = $this->getById($this->configService->getSetting('SYSTEM_USER'));
```
dasgarner, line 201: Do we still need to override the email address here? Or have we traced through where we use the system user for emailing, and replaced with the global `mail_to` at that point?

lib/Service/DisplayNotifyService.php
```diff
@@ -623,4 +623,40 @@ public function notifyByPlaylistId($playlistId)
...
+                       INNER JOIN action on widget.widgetId = action.sourceId
+                WHERE
+                    action.layoutCode = :code AND
+                    layout.publishedStatusId = 1',
+            ['code' => $code]);
+
+        $this->log->debug('Notify by Layout Code, found following Campaigns to notify: ' . json_encode($campaignIds));
+
+        foreach ($campaignIds as $row) {
+            $this->notifyByCampaignId($row['campaignId']);
```
dasgarner, line 659: I can see why you've done this - but I think it might grind to a halt if there are more than a few to process.

There are two options I think - either modify `notifyByCampaignId` to take an array of CampaignIds, or modify it to take some SQL which we substitute in the `campaigns` view.

lib/Xmds/Soap.php
```diff
@@ -427,6 +429,47 @@ protected function doRequiredFiles($serverKey, $hardwareKey, $httpDownloads)
             return new \SoapFault('Sender', 'Unable to get a list of layouts');
         }
 
+        // workout if any of the layouts we have in our list has Actions pointing to another Layout.
+        foreach ($layouts as $layoutId) {
```
dasgarner, line 433: Looking at this, do we need to have a method in layout factory which just returns the ID? I'm a bit concerned we'll end up running getByCode potentially a bunch of times? which then runs the full query.

Would it actually be possible to have a `$layout->getActionLayoutIds()` method which just runs some SQL to get a bunch of IDs? That would be super efficient by comparison.

views/action-form-add.twig
```diff
@@ -58,19 +58,24 @@
...
                     { typeid: "next", type: next },
                     { typeid: "previous", type: previous },
                     { typeid: "navLayout", type: navLayout },
-                    { typeid: "navRegion", type: navRegion },
-                    { typeid: "navPlaylist", type: navPlaylist },
                     { typeid: "navWidget", type: navWidget },
                 ] %}
                 {{ forms.dropdown("actionType", "single", title,"", options, "typeid","type", helpText) }}
 
+                {% set title %}{% trans "Widget ID" %}{% endset %}
```
dasgarner, line 71: We have called this field Widget ID, but actually we are showing a list of Widgets (not IDs) - should we rename?

views/common.twig
```diff
@@ -515,6 +515,8 @@
                     actionType: "{{ "Action Type"|trans }}",
                     target: "{{ "Target"|trans }}",
                     targetId: "{{ "Target Type"|trans }}",
+                    widgetId: "{{ "Widget ID"|trans }}",
```
dasgarner, line 518: Any way to show the Widget here instead?

views/layout-form-add.twig
```diff
@@ -92,6 +92,10 @@
                         {% set helpText %}{% trans "Choose the resolution this Layout should be designed for." %}{% endset %}
                         {{ forms.dropdown("resolutionId", "single", title, "", resolutions, "resolutionId", "resolution", helpText, "resolution-group") }}
 
+                        {% set title %}{% trans "Code" %}{% endset %}
+                        {% set helpText %}{% trans "The Code identifier for this layout - used in Interactive Actions" %}{% endset %}
```
dasgarner, line 96:
```suggestion
                        {% set helpText %}{% trans "The Code identifier for this Layout - used in Interactive Actions" %}{% endset %}
```

lib/Controller/Task.php
```diff
@@ -447,8 +447,9 @@ public function run(Request $request, Response $response, $id)
         // Run
         try {
             // Instantiate
-            if (!class_exists($task->class))
-                throw new NotFoundException('Task with class name ' . $task->class . ' not found');
+            if (!class_exists($task->class)) {
+                throw new NotFoundException(__('Task with class name ') . $task->class . ' not found');
```
dasgarner, line 451: This translation needs a sprintf for the class name.

lib/Factory/DataSetFactory.php
```diff
@@ -662,7 +662,7 @@ private function processEntry(array $entry, array $dataSetColumns) {
                                 $result[$column->heading] = $date->format(DateFormatHelper::getSystemFormat());
                             } catch (\Exception $e) {
                                 $this->getLog()->error('Incorrect date provided ' . $date . ' Expected date format Y-m-d H:i:s ');
-                                throw new InvalidArgumentException('Incorrect date provided ' . $date . ' Expected date format Y-m-d H:i:s ', 'date');
+                                throw new InvalidArgumentException(__('Incorrect date provided ') . $date . __(' Expected date format Y-m-d H:i:s '), 'date');
```
dasgarner, line 665: This wants to be a sprintf to substitute in the date.

lib/Service/DisplayNotifyService.php
```diff
@@ -219,13 +219,15 @@ public function notifyByDisplayGroupId($displayGroupId)
...
+    public function notifyByCampaignId($campaignIds)
     {
-        $this->log->debug('Notify by CampaignId ' . $campaignId);
+        foreach ($campaignIds as $campaignId) {
+            $this->log->debug('Notify by CampaignId ' . $campaignId);
 
-        if (in_array('campaign_' . $campaignId, $this->keysProcessed)) {
-            $this->log->debug('Already processed ' . $campaignId . ' skipping this time.');
-            return;
+            if (in_array('campaign_' . $campaignId, $this->keysProcessed)) {
```
dasgarner, line 227: This means that if any of the campaignIds in $campaignIds has already been processed, we don't process any of the others?

lib/Service/DisplayNotifyService.php
```diff
@@ -219,13 +219,15 @@ public function notifyByDisplayGroupId($displayGroupId)
     }
 
     /** @inheritdoc */
-    public function notifyByCampaignId($campaignId)
+    public function notifyByCampaignId($campaignIds)
```
dasgarner, line 222: Would it be easier to make it so that we can pass an int or an array into this method?

dasgarner, line 222: Sorry, but having looked at this I don't think I like passing round the campaignIds.

I think it would be better if this method went back to how it was, and we implemented the full SQL in the `notifyByLayoutCode` method.

Even though that is more code and repeated code, what actually runs is clean, parameterised and memory efficient.

PR #894, opened 2020-04-22
xibosignage#2161 Jalali and Gregorian new datepickers

maurofmferrao:


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
dasgarner, line 1: Should this be here?

maurofmferrao: That file came with the new Cypress version, shall I remove it?

PR #895, opened 2020-04-22
Report Schedule Tests

ifarzana:


lib/Controller/Report.php
```diff
@@ -441,18 +450,36 @@ public function reportScheduleDelete(Request $request, Response $response, $id)
      */
     public function reportScheduleDeleteAllSavedReport(Request $request, Response $response, $id)
     {
+        $sanitizedParams = $this->getSanitizer($request->getParams());
 
-        $reportSchedule = $this->reportScheduleFactory->getById($id);
+        $disableUserCheck = $sanitizedParams->getInt('disableUserCheck');
```
dasgarner, line 455: Does this mean that anyone calling the API can pass in "disableUserCheck" and delete someone else's scheduled reports?

I am not sure why this is a parameter here.

ifarzana: you are right, this means anyone can delete saved reports which is not good. I have changed this now.

PR #898, opened 2020-04-23
3.0.0 alpha

PeterMis:
Layout Lock feature
Effective Page permissions for User ACL
Remote DataSet fixes and implementation of row limit with row policy
Various other small code adjustments


lib/Controller/DataSetData.php
```diff
@@ -158,7 +158,7 @@ public function grid(Request $request, Response $response, $id)
         $filter = trim($filter, 'AND');
 
         // Work out the limits
-        $filter = $this->gridRenderFilter(['filter' => $request->getParam('filter', $filter)], $request);
+        $filter = $this->gridRenderFilter(['filter' => $filter], $request);
```
dasgarner, line 161: I think this was right how it was

lib/Controller/Layout.php
```diff
@@ -1230,6 +1231,7 @@ function grid(Request $request, Response $response)
 
             // Populate the status message
             $layout->getStatusMessage();
+            $layout->isLocked = $this->container->get('pool')->getItem('locks/layout/' . $layout->layoutId)->get();
```
dasgarner, line 1234: Although we have the container here, that is actually a workaround and shouldn't be used as routine.

We should pass in the Pool as a named parameter please.

dasgarner: I think you also need to check if it `isMiss()` or not, expired locks might hang around for some time after they have actually expired, depending on the cache engine and when it does garbage collection.

lib/Entity/DataSet.php
```diff
@@ -538,6 +550,12 @@ public function getData($filterBy = [], $options = [])
         $limit = '';
         if ($start != 0 || $size != 0) {
             // Substitute in
+
+            // handle case where lower limit is set to > 0 and upper limit to 0 https://github.com/xibosignage/xibo/issues/2187
+            if ($start != 0 && $size == 0) {
+                $size = 18446744073709551615;
```
dasgarner, line 556: I think this might blow up on 32-bit systems?

lib/Factory/DataSetFactory.php
```diff
@@ -453,7 +455,17 @@ public function callRemoteService(DataSet $dataSet, DataSet $dependant = null, $
...
                     $result->entries[] = json_decode($request->getBody());
-                    $result->number = $result->number + 1;
+                    foreach ($result->entries as $entry) {
+                        $data = $this->getDataRootFromResult($dataSet->dataRoot, $entry);
+
+                        if (is_array($data)) {
+                            $result->number = count($data);
+                        }
+
+                        if (is_object($data)) {
```
dasgarner, line 465: Could be an elseif?

lib/Factory/DataSetFactory.php
```diff
@@ -657,7 +669,7 @@ private function processEntry(array $entry, array $dataSetColumns) {
                             break;
                         case 3:
                             // This expects an ISO date
-                            $date =  $sanitizer->getDate($value[1]);
+                            $date =  $sanitizer->getDate(1);
```
dasgarner, line 672: Is this right?

lib/routes.php
```diff
@@ -110,20 +113,23 @@
  *  description="Layouts"
  * )
  */
+$app->group('/layout', function (RouteCollectorProxy $group) {
```
dasgarner, line 116: Do you think we should do the same on the forms in routes-web ?

PR #917, opened 2020-06-02
Revision cache busting

dasgarner:


lib/Controller/Display.php
```diff
@@ -662,20 +663,20 @@ function grid(Request $request, Response $response)
             }
 
             // Delete
-            if ($this->getUser()->checkDeleteable($display)) {
-                $display->buttons[] = array(
+            if (Environment::isDevMode() && $this->getUser()->checkDeleteable($display)) {
```
PeterMis, line 666: this will make it so the delete button is only available in devMode, is that correct?

dasgarner: No, not correct! I will fix it, thanks.

PR #942, opened 2020-07-07
Commands: Improve the UI for creating commands

maurofmferrao:
fixes xibosignage/xibo#2247

 - Command types renamed
 - Fix for non default commands to become free text command


ui/src/core/xibo-cms.js
```diff
@@ -479,8 +479,8 @@ function XiboInitialise(scope) {
         $input.hide();
 
         const commandTypes = {
-            'freetext': 'freeText',
```
dasgarner, line 482: I think we need a translation for this one (probably not the other one as its a product name).

PR #950, opened 2020-07-20
Report Restructuring

ifarzana:


lib/Controller/Report.php
```diff
@@ -805,6 +808,19 @@ public function savedReportGrid(Request $request, Response $response)
             /** @var \Xibo\Entity\SavedReport $savedReport */
 
             $savedReport->includeProperty('buttons');
+            
+            $schemaVersion = $savedReport->schemaVersion;
```
dasgarner, line 812: Unnecessary assignment?

lib/Controller/Report.php
```diff
@@ -805,6 +808,19 @@ public function savedReportGrid(Request $request, Response $response)
...
+
+            // Show only convert button for schema version 1
+			if ($schemaVersion == 1) {
+
+                $savedReport->buttons[] = [
+                    'id' => 'button_convert_report',
+                    'url' => $this->urlFor($request,'savedreport.convert.form', ['id' => $savedReport->savedReportId] ),
+                    'text' => __('Convert')
+                ];
+                continue;
```
dasgarner, line 822: `continue` because you don't want the rest of the buttons to be output?

I think an `else` would be much clearer here for someone coming to this at a later date.

lib/Controller/Report.php
```diff
@@ -1090,6 +1106,68 @@ public function savedReportExport(Request $request, Response $response, $id, $na
...
+     * @throws AccessDeniedException
+     * @throws GeneralException
+     * @throws NotFoundException
+     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     */
+    public function convertSavedReportForm(Request $request, Response $response, $id)
+    {
+        $savedReport = $this->savedReportFactory->getById($id);
+
+        if (!$this->getUser()->checkDeleteable($savedReport)) {
```
dasgarner, line 1124: Do we want delete permissions for this operation? or edit permissions?

lib/Controller/Report.php
```diff
@@ -1090,6 +1106,68 @@ public function savedReportExport(Request $request, Response $response, $id, $na
...
+     * @return \Psr\Http\Message\ResponseInterface|Response
+     * @throws GeneralException
+     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     */
+    public function savedReportConvert(Request $request, Response $response, $id, $name)
+    {
+        $savedReport = $this->savedReportFactory->getById($id);
+
+        if ($savedReport->schemaVersion == 2) {
+            throw new AccessDeniedException(__('You cannot convert this saved report'));
```
dasgarner, line 1153: Access denied doesn't seem right? Perhaps a "General Exception" and the message "This report has already been converted to the latest version."

lib/Factory/SavedReportFactory.php
```diff
@@ -135,6 +135,7 @@ public function query($sortOrder = null, $filterBy = [])
                saved_report.savedReportId,
                saved_report.saveAs,
                saved_report.userId,
+               saved_report.schemaVersion,
```
dasgarner, line 138: I think you might want to hydrate this as an `intProperty` ?

lib/Service/ReportService.php
```diff
@@ -294,6 +300,61 @@ public function getSavedReportResults($savedreportId, $reportName)
 
      }
 
+    /**
+     * @inheritdoc
+     */
+    public function convertSavedReportResults($savedreportId, $reportName)
+    {
+        throw new InvalidArgumentException(__('Can\'t create ZIP. Error Code:'));
```
dasgarner, line 308: Is this right?

lib/Controller/Report.php
```diff
@@ -808,75 +808,87 @@ public function savedReportGrid(Request $request, Response $response)
...
             }
 
-            $savedReport->buttons[] = [
-                'id' => 'button_show_report.now',
-                'class' => 'XiboRedirectButton',
-                'url' => $this->urlFor($request,'savedreport.open', ['id' => $savedReport->savedReportId, 'name' => $savedReport->reportName] ),
-                'text' => __('Open')
-            ];
-            $savedReport->buttons[] = ['divider' => true];
+            if ($compatible == 1) {
```
dasgarner, line 823: Sorry to be pedantic, but this should be `if ($compatible) {` ? because compatible is a boolean?

lib/Controller/Report.php
dasgarner: This has gone? We aren't checking permissions at all on convert now. Is that right?

PR #952, opened 2020-07-27
Tags with values enhancement

PeterMis:


lib/Factory/TagFactory.php
```diff
@@ -104,6 +106,11 @@ public function tagFromString($tagString)
         // Add to the list
         try {
             $tag = $this->getByTag($explode[0]);
+
+            if ($tag->isRequired == 1 && !isset($explode[1])) {
+                throw new InvalidArgumentException(sprintf('Selected Tag %s requires a value, please enter the Tag in %s|Value format', $explode[0], $explode[0]), 'options');
```
dasgarner, line 111: While this will work, I wonder if its user friendly enough?
Can we have it so that a text input field is shown on the form when the tag is entered?

PeterMis: I'm honestly not sure on this one.

The `tag|value` format is also how the filters work for tags with values.

Making it so instead of one field where you enter `tag|value` you get
one field with just tag
second field with just value

is not as straightforward as it may seem, I'm not entirely sure if it would be more user friendly, but I can certainly do that if that's what we want - not sure if it would be well tested for 2.3.6 today though 🤔

PR #957, opened 2020-07-28
Cherry Pick from 2.3.6 into 3.0.0-beta

dasgarner:


lib/Controller/Layout.php
```diff
@@ -2219,10 +2220,17 @@ public function export(Request $request, Response $response, $id)
             throw new InvalidArgumentException(__('Cannot manage tags on a Draft Layout'), 'layoutId');
         }
 
+        // Save As?
+        $saveAs = $this->getSanitizer()->getString('saveAs');
```
PeterMis, line 2224: That probably should be `$saveAs = $sanitizedParams->getString('saveAs')`

dasgarner: ah yes, good catch thanks!

PR #960, opened 2020-07-29
Report restructure - Fixes

ifarzana:


lib/routes.php
```diff
@@ -356,7 +356,6 @@
  * )
  */
 $app->get('/stats', ['\Xibo\Controller\Stats','grid'])->setName('stats.search');
-$app->get('/stats/timeDisconnected', ['\Xibo\Controller\Stats', 'gridTimeDisconnected'])->setName('stats.timeDisconnected.search');
```
dasgarner, line 359: This is an API which we put in for a user request

lib/Controller/Stats.php
```diff
@@ -837,430 +836,6 @@ public function export(Request $request, Response $response)
...
-     *      )
-     *  )
-     * )
-     * @param Request $request
-     * @param Response $response
-     * @return \Psr\Http\Message\ResponseInterface|Response
-     * @throws InvalidArgumentException
-     * @throws \Xibo\Support\Exception\GeneralException
-     */
-    public function gridTimeDisconnected(Request $request, Response $response): Response
```
dasgarner, line 1150: This is an API we put in for a user request (hence the API documentation, etc)

ifarzana:
This Pull Request includes: 

- [x] Summary Report: Saved report conversion from old schema to new schema returned an error `Undefined index for Period Start' - This problem is due to very old reportschedule.json file. The conversion works for latest saved reports.
- [x] Refined Category names 
- [x] Proof of Play report fix 
- [x] Time disconnected report showing 200% for a player having two null displayevents.

lib/routes.php
ifarzana: If I put this back, I think that is fine. We are not using this method in any of our reports

lib/Controller/Stats.php
ifarzana: If I put this back, I think that is fine.

PR #968, opened 2020-08-20
Tour: Welcome page and tour fixes

maurofmferrao:
resolves xibosignage/xibo#2218


views/welcome-page.twig
```diff
@@ -25,65 +25,196 @@
...
+            box-shadow: 0px 0px 10px 3px #00000026;
+        }
+    </style>
+{% endblock %}
+
+{% block pageContent %}
+
+    <div class="container-fluid welcome-container">
+        <div class="col-xs-12">
+            <h2>{{ "Welcome to the [[PRODUCTNAME]] CMS"|trans|replace({'[[PRODUCTNAME]]': theme.getThemeConfig('theme_title') }) }}</h2>
```
dasgarner, line 135: I think we can do this with `trans` substitutions instead.

```
{% set productName = theme.getThemeConfig('theme_title') %}
{% trans %}Welcome to the %productName% CMS{% endtrans %}
```

views/welcome-page.twig
```diff
@@ -25,65 +25,196 @@
...
+            <p>{{ "The [[PRODUCTNAME]] system comprises a central content management system (the CMS you are currently logged into) and Player software to download content from the CMS and play on screens."|trans|replace({'[[PRODUCTNAME]]': theme.getThemeConfig('theme_title') }) }}</p>
```
dasgarner, line 136: Then

```
{% trans %}The %productName% system comprises a central content management system (the CMS you are currently logged into) and Player software to download content from the CMS and play on screens.{% endtrans %}
```

views/welcome-page.twig
```diff
@@ -25,65 +25,196 @@
...
+
+        {# CMS players #}
+        {% include "welcome-page_players.twig" %}
+
+        <div class="col-xs-12 welcome-title">{% trans "We’d like to take you on a tour of the CMS using a series of short tutorials to get you up and running in no time. If you already know your way around, please skip." %}</div>
+        
+        <div class="tour-cards-section">
+            <div class="col-xs-12 col-sm-6 col-md-4 tour-card-container">
+                <div class="tour-card tour-link" data-tour-name="authDisplayTour">
+                    <img class="tour-image" src="{{ theme.rootUri() }}theme/default/img/tours/authDisplayTour.png">
```
dasgarner, line 154: If we do it like this, then these cannot be replaced in a theme.

Instead:

```
{{ theme.uri("img/tours/authDisplayTour.png") }}
```

dasgarner (changes requested):
I have some suggestions around Twig usage to make the translations easier to translate, but otherwise this looks good.

views/welcome-page.twig
maurofmferrao: @dasgarner , using the %variable% inside the trans element doesn't really seem to work for the replacement, but {{variable}} does, is it the latter the right one?

PR #987, opened 2020-09-24
Feature/3.0.0 beta

PeterMis:
Video cover image created on video upload
Display Screenshot TTL
Display Status window data under screenshot

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -590,7 +590,9 @@ class Library extends Base
             $media->thumbnailUrl = '';
             $media->downloadUrl = '';
 
-            if ($media->mediaType == 'image' || ($media->mediaType == 'video' && file_exists($libraryLocation . $media->mediaId . '_videocover.png'))) {
+            $module = $this->moduleFactory->createWithMedia($media);
+
+            if ($module->hasThumbnail()) {
                 $download = $this->urlFor($request,'library.download', ['id' => $media->mediaId], ['preview' => 1]);
                 $media->thumbnail = '<a class="img-replace" data-toggle="lightbox" data-type="image" href="' . $download . '"><img src="' . $download . '&width=100&height=56&cache=1" /></i></a>';
                 $media->thumbnailUrl = $download . '&width=100&height=56&cache=1';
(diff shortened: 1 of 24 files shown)


lib/Controller/Library.php
```diff
@@ -589,7 +590,7 @@ function grid(Request $request, Response $response)
             $media->thumbnailUrl = '';
             $media->downloadUrl = '';
 
-            if ($media->mediaType == 'image') {
+            if ($media->mediaType == 'image' || ($media->mediaType == 'video' && file_exists($libraryLocation . $media->mediaId . '_videocover.png'))) {
```
dasgarner, line 593: Can we have a new method on the media interface for this? something like `hasThumbnail()` which is default false in ModuleWidget, but the overridden in Image and now also in Video.

Would allow us to extend more easily in the future (say if someone wanted to provide a thumbnail for package html)

lib/Controller/Library.php
```diff
@@ -2545,4 +2546,44 @@ private function getRemoteFileExtension($url)
...
+     * @return Response
+     * @throws InvalidArgumentException
+     */
+    public function videoThumbnail($request, $response)
+    {
+        $sanitizedParams = $this->getSanitizer($request->getParams());
+        $libraryLocation = $this->getConfig()->getSetting('LIBRARY_LOCATION');
+
+        $image = $request->getParam('image');
+        $mediaId = $sanitizedParams->getInt('mediaId');
```
dasgarner, line 2566: At this point, do we need to check permissions for the `mediaId`?

Maybe it could also be called "addThumbnail" so that it is more generic?

lib/Controller/Library.php
```diff
@@ -2545,4 +2546,44 @@ private function getRemoteFileExtension($url)
...
+
+        if (preg_match('/^data:image\/(\w+);base64,/', $image, $type)) {
+            $image = substr($image, strpos($image, ',') + 1);
+            $type = strtolower($type[1]);
+
+            if (!in_array($type, [ 'jpg', 'jpeg', 'gif', 'png' ])) {
+                throw new InvalidArgumentException('invalid image type');
```
dasgarner, line 2573: Do these error messages need to be translated (and perhaps more user friendly)?

lib/Helper/XiboUploadHandler.php
```diff
@@ -327,6 +327,11 @@ protected function handle_form_data($file, $index)
             $file->enableStat = $media->enableStat;
             $file->mediaType = $module->getModuleType();
 
+            // we need original fileName returned for video thumbnails logic.
+            if ($media->mediaType == 'video') {
```
dasgarner, line 331: we could always return this - so that the thumbnail function can be used generically.

lib/routes.php
```diff
@@ -252,6 +252,9 @@
 $app->post('/library/{id}/tag', ['\Xibo\Controller\Library','tag'])->setName('library.tag');
 $app->post('/library/{id}/untag', ['\Xibo\Controller\Library','untag'])->setName('library.untag');
 
+// video thumbnail
+$app->post('/library/video/thumbnail', ['\Xibo\Controller\Library','videoThumbnail'])->setName('library.video.thumbnail.save');
```
dasgarner, line 256: Maybe make this more generic.

dasgarner (changes requested):
The FE stuff looks perfect, as does the screenshot/status window stuff.
I've made some minor suggestions for the BE to make it more generic

dasgarner (approved):

PR #988, opened 2020-09-25
Permissions: Bulk set for selected items

maurofmferrao:
resolves xibosignage/xibo#2217

### DIFF
--- a/lib/Controller/User.php
+++ b/lib/Controller/User.php
@@ -1603,27 +1603,27 @@ class User extends Base
         foreach ($permissions as $permission) {
 
             if(!array_key_exists($permission->groupId, $newPermissions)) {
-                $newPermissions[$permission->groupId] = array(
+                $newPermissions[$permission->groupId] = [
                     "groupId" => $permission->groupId,
                     "group" => $permission->group,
                     "isUser" => $permission->isUser,
                     "entity" => $permission->entity,
-                    "permissions" => array(
-                        $permission->objectId => array(
+                    "permissions" => [
+                        $permission->objectId => [
                             "permissionId" => $permission->permissionId,
                             "view" => $permission->view,
                             "edit" => $permission->edit,
                             "delete" => $permission->delete
-                        )
-                    )
-                );
+                        ]
+                    ]
+                ];
             } else {
-                $newPermissions[$permission->groupId]["permissions"][] = array(
+                $newPermissions[$permission->groupId]["permissions"][] = [
                     "permissionId" => $permission->permissionId,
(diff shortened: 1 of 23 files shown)


lib/Controller/User.php
```diff
@@ -1524,6 +1524,116 @@ public function permissionsGrid(Request $request, Response $response,$entity, $i
...
+        // Check if the array of ids is passed
+        if($sanitizedParams->getString('ids') == "") {
+            throw new InvalidArgumentException(__('The array of ids is empty!'));
+        }
+
+        // Get array of ids
+        $ids = explode(',', $sanitizedParams->getString('ids'));
+
+        // Array of all the permissions
+        $permissions = array();
```
dasgarner, line 1581: We should try to use the newer array syntax for new code - so `[]` instead of `array()`

lib/Controller/User.php
```diff
@@ -1524,6 +1524,116 @@ public function permissionsGrid(Request $request, Response $response,$entity, $i
...
+
+            // List of all Groups with a view / edit / delete check box
+            $permissions = array_merge_recursive($permissions, $this->permissionFactory->getAllByObjectId($this->getUser(), $objects[$i]->permissionsClass(), $objectId, $this->gridRenderSort($request), $this->gridRenderFilter(['name' => $sanitizedParams->getString('name')], $request)));
+        }
+
+        // Change permissions structure to be grouped by user group
+        foreach ($permissions as $permission) {
+
+            if(!array_key_exists($permission->groupId, $newPermissions)) {
+                $newPermissions[$permission->groupId] = array(
```
dasgarner, line 1606: New array syntax.

lib/Controller/User.php
```diff
@@ -1524,6 +1524,116 @@ public function permissionsGrid(Request $request, Response $response,$entity, $i
...
+     * @throws GeneralException
+     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     * @throws \Xibo\Support\Exception\NotFoundException
+     */
+    public function permissionsMultiGrid(Request $request, Response $response, $entity)
+    {
+        $sanitizedParams = $this->getSanitizer($request->getParams());
+
+        // Check if the array of ids is passed
+        if($sanitizedParams->getString('ids') == "") {
```
dasgarner, line 1573: Quotes should be `'` instead of `"` where possible.

dasgarner (changes requested):
The actual code/logic looks fine to me. I've commented on some syntax things which would be good to correct before we merge please.

dasgarner (approved):

PR #991, opened 2020-10-09
Widget Interactive JavaScript

maurofmferrao:
resolves xibosignage/xibo#2340

 - Include new Xibo IC library
 - Add the library to the modules
 - Create a call to the layout designer viewer to enable trigger actions

### DIFF
--- a/lib/Widget/Text.php
+++ b/lib/Widget/Text.php
@@ -37,7 +37,9 @@ class Text extends ModuleWidget
      */
     public function installFiles()
     {
-        $this->mediaFactory->createModuleSystemFile(PROJECT_ROOT . '/modules/vendor/jquery.min.js')->save();
+        // Extends parent's method
+        parent::installFiles();
+        
         $this->mediaFactory->createModuleSystemFile(PROJECT_ROOT . '/modules/vendor/moment.js')->save();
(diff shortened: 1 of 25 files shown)


lib/Widget/Text.php
```diff
@@ -273,8 +276,15 @@ public function getResource($displayId = 0)
 
         // Finalise some JavaScript to run.
         $javaScriptContent = '$(document).ready(function() { ';
-        $javaScriptContent .= '       $("#content").xiboTextRender(options, items); $("body").xiboLayoutScaler(options); $("#content").find("img").xiboImageRender(options); ';
+        
+        // Run on document ready
+        $javaScriptContent .= '     $("body").xiboLayoutScaler(options); $("#content").find("img").xiboImageRender(options); ';
 
+        // Run based only if the element is visible or not
+        $javaScriptContent .= '     const runOnVisible = function() { $("#content").xiboTextRender(options, items); }; ';
```
dasgarner, line 284: So I think we can apply this logic to other types of Widget, except Embedded/Web Page/Package HTML?

dasgarner (changes requested):
I think this looks good - one minor bit of feedback and one other for follow on discussion

dasgarner (approved):

PR #998, opened 2020-10-24
File Uploader Refactor

dasgarner:
Refactor the file upload tool so that it is called generically with options - there is only one set of templates and only one set of JS to open and configure the uploader.

All additional templates / code to create a file uploader has been removed.

I also found some files which were not used and tidied them up.

In addition issue #2264 has been addressed.

Draft until prior PR is merged and #2210 has been included.


maurofmferrao (dismissed):
All looks good to me

dasgarner:
Peter, please can you check the result of merging your folder work into this. Thanks.

ui/src/helpers/form-helpers.js
```diff
@@ -873,28 +873,38 @@ let formHelpers = function() {
...
-                    main: {
-                        label: translations.done,
-                        className: 'btn-primary',
-                        callback: function() {
-                            if(typeof self.namespace.timeline.resetZoom === 'function') {
-                                self.namespace.timeline.resetZoom();
+                    url: libraryAddUrl,
+                    title: uploadTrans.uploadMessage,
+                    animateDialog: false,
+                    initialisedBy: "library-upload",
```
dasgarner, line 881: I was not sure about this...

PeterMis (approved):

views/include-file-upload.twig
```diff
@@ -29,6 +29,9 @@
         {{/if}}
         {{#if oldMediaId}}
             <input type="hidden" name="oldMediaId" value="{{ oldMediaId }}" />
+            <input name="folderId" type="hidden" id="folderId" value="{{oldFolderId}}">
```
PeterMis, line 32: I'm not entirely sure, but this may need to be formFolderId (id), we may need to try that really and see

ui/src/helpers/form-helpers.js
PeterMis: I'm not sure either if I'm honest, I've added a comment below, but I think it might be best to just try it on develop.

views/include-file-upload.twig
dasgarner: OK good shout - lets try it and see.

PeterMis (approved):

PR #1008, opened 2020-10-30
Grid Folder feature: Fix double click and table horizontal resizing

maurofmferrao:
fixes xibosignage/xibo#2347

 - Table inline style with at 100%
 - Row select fix
 - Display grid "remote" field warning fix

### DIFF
--- a/views/campaign-page.twig
+++ b/views/campaign-page.twig
@@ -89,7 +89,7 @@
                     </div>
                     <div id="datatable-container" class="col-sm-10">
                         <div class="XiboData">
-                            <table id="campaigns" class="table table-striped" data-content-type="campaign" data-content-id-name="campaignId" data-state-preference-name="campaignGrid">
+                            <table id="campaigns" class="table table-striped" data-content-type="campaign" data-content-id-name="campaignId" data-state-preference-name="campaignGrid" style="width: 100%;">
                                 <thead>
                                     <tr>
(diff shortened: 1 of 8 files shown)


views/campaign-page.twig
```diff
@@ -89,7 +89,7 @@
                     </div>
                     <div id="datatable-container" class="col-sm-10">
                         <div class="XiboData">
-                            <table id="campaigns" class="table table-striped" data-content-type="campaign" data-content-id-name="campaignId" data-state-preference-name="campaignGrid">
+                            <table id="campaigns" class="table table-striped" data-content-type="campaign" data-content-id-name="campaignId" data-state-preference-name="campaignGrid" style="width: 100%;">
```
dasgarner, line 92: Any reason we can't apply this to all `table` within `#datatable-container` via a stylesheet instead? I think that would fix resizing the window on other pages too?

dasgarner (changes requested):
Just one comment :-)

views/campaign-page.twig
maurofmferrao: Datatables recommends adding the inline style property to have a flexible table: https://datatables.net/examples/basic_init/flexible_width.html

We can use something like this in our CSS:
`#datatable-container table.dataTable {
    width: 100% !important;
}`
But I'm not sure this will break anything in the datatable's flow :thinking: 
What do you think @dasgarner?

PR #1010, opened 2020-11-02
(3.1.0 alpha) Stat API and Archive Task*

ifarzana:


dasgarner (approved):
Looks good.

I think we can put this in 3.0.0-rc1 as they are relatively minor things (not new features). I will leave this PR open until we're closer to releasing that.

PR #1011, opened 2020-11-02
Feature/3.1.0 alpha

maurofmferrao:


lib/Controller/Display.php
```diff
@@ -1444,6 +1460,101 @@ public function membershipForm(Request $request, Response $response, $id)
...
+            // convert bandwidth to kb based on form units
+            if ($bandwidthLimitUnits == 'mb') {
+                $bandwidthLimit = $bandwidthLimit * 1024;
+            } else if ($bandwidthLimitUnits == 'gb') {
+                $bandwidthLimit = $bandwidthLimit * 1024 * 1024;
+            }
+
+            // save display group
+            $displayGroup->bandwidthLimit = $bandwidthLimit;
+            $displayGroup->save(['validate' => false, 'audit' => false]);
```
dasgarner, line 1545: So I think in the grand scheme of things this doesn't really save us much in terms of processing, because we're still saving each one individually. It might even be slightly worse because we're adding multiple updates to one big transaction :-)

What we need is a method on the displayGroupFactory called `setBandwidth` which takes an array of displayGroupId's and then does a single update statement for all of them.

Equally I think we'd want a single audit message indicating bandwidth limits were modified on a batch of display groups at the same time.

If you'd like some direction on this, please let me know.

dasgarner (changes requested):

dasgarner (approved):

dasgarner (dismissed):

dasgarner (approved):
Great!

dasgarner (approved):

dasgarner:
Will promote these small features into 3.0.0-rc1

PR #1036, opened 2021-01-19
3.0.0 rc1

PeterMis:


lib/Controller/DisplayProfileConfigFields.php
```diff
@@ -34,407 +34,501 @@ trait DisplayProfileConfigFields
...
+     * @param SanitizerInterface $sanitizedParams
      * @param null|array $config if empty will edit the config of provided display profile
-     * @param Request $request
+     * @param \Xibo\Entity\Display $display
      * @return null|array
      * @throws InvalidArgumentException
      * @throws \Xibo\Support\Exception\NotFoundException
      */
-    public function editConfigFields($displayProfile, $config = null, Request $request)
+    public function editConfigFields($displayProfile, $sanitizedParams, $config = null, $display = null)
```
dasgarner, line 44: This change in method signature makes a lot of sense

lib/Controller/DisplayProfileConfigFields.php
```diff
@@ -34,407 +34,501 @@ trait DisplayProfileConfigFields
...
     {
-
-        /** @var SanitizerInterface $sanitizedParams */
-        $sanitizedParams = $this->getSanitizer($request->getParams());
-
         // Setting on our own config or not?
         $ownConfig = ($config === null);
 
+        $changedSettings = [];
```
dasgarner, line 49: I'm not sure that this is the right approach here. What I thought we might end up doing was auditing in the calling method. What I mean here is that before we call `editConfigFields` we capture the current settings for that object (Display Profile we capture the config, Display we capture the override config). Then `editConfigFields` is called, and we capture the current settings again.

We then audit the differences between those two arrays. `handleChangedSettings` can stay in the trait, but take two arrays, an original and a new.

There might be a very good reason why this won't work, and if so lets talk it through.

dasgarner (changes requested):

dasgarner (approved):

lib/Controller/DisplayProfileConfigFields.php
dasgarner: After review, approved.

PR #1050, opened 2021-02-02
Improved User Onboarding

dasgarner:
https://github.com/xibosignage/xibo/issues/1910


PeterMis (approved):
Looks good to me, we just need to fix the small issues raised during checks ;)

PeterMis (approved):

PR #1059, opened 2021-02-09
Feature/3.0.0 rc1 multiple

maurofmferrao:


ui/src/designer/widget.js
```diff
@@ -56,7 +56,7 @@ let Widget = function(id, data, regionId = null, layoutObject = null) {
     this.mediaIds = data.mediaIds;
 
     // Widget colouring
-    if(playlistRegionColouring === 'Permissions Colouring') {
```
dasgarner, line 59: I think we might need to cater for old settings values here - perhaps we can have it in an `or` ?

dasgarner (changes requested):
Just one change I think :crossed_fingers:

dasgarner (approved):

dasgarner (approved):

PR #1062, opened 2021-02-18
Commands: HTTP helper

maurofmferrao:
resolves xibosignage/xibo#2248


dasgarner (commented):
This looks good to me - but please don't merge until we've branched out a `release30`

I also wonder whether we should separate the commands logic into a different file when we look at webpack for xibo-cms ?

dasgarner (approved):

PR #1064, opened 2021-02-22
Wolf 2.3.9

PeterMis:


lib/Widget/Image.php
```diff
@@ -198,7 +198,14 @@ public function getResource($displayId = 0)
         if (stripos($media->storedAs, '.') > -1) {
             $extension = explode('.', $media->storedAs)[1];
         } else if (stripos($media->fileName, '.')) {
-            $extension = explode('.', $media->fileName)[1];
+            // trim the string if we have any characters after the extension with ?
```
dasgarner, line 201: Is this to solve the Media RSS issue? My comment on the issue may have been misleading, sorry.

I think this `?sdsd` suffix needs to be handled on the incoming image rather than on preview. I.e. when we parse the RSS feed. 

I've no idea if the image will be correctly downloaded and shown on the players with a `?` in the file name..

dasgarner (changes requested):
One comment here, partly my fault I think ;-)

lib/Widget/Image.php
PeterMis: Yes it is about the RSS issue and I don't think it was ever a player issue.

Ticker getResource() for images in rss feed goes to Library download with preview flag for preview, which eventually ends up in the Image getResource() - hence why the change is there.

For non preview, we just send the ticker file with storedAs ie for example `ticker_88dff64eb0e0fbd9b60adc2ad35ac99d` , that's what is sent to player and that's what we put as image source in the html sent to the Player.
ie we don't care about fileName or extension there from the looks of things.

dasgarner (approved):
Great

PR #1066, opened 2021-02-24
DataSetTicker has a global cache only

guile2912:
Even if you can set a sql filter on the DataSetTicker, it is cache globally, making the [DisplayId] filtering useless and source of errors as it will display one random display content ( the one in cache) on the display that requested content between 2 refresh.

This is also very bad because the DataSetTicker is a very powerfull widget. We are using it to push display specific messages on each of our 150 screens, but using only one global layout for all display. Just fill in a DataSet entry, and thats it, message is live on the screens ( with publishing dates, priorities, differents css layout and background images )

This patch fixes the DataSetTicker to make its cache Display specific if there is a [DisplayId] of [DisplayGeoLocation] in the DataSet filter


dasgarner:
Thank you for your PR, much appreciated. Please can you change the base to `release23` so we can complete testing before merging into master? Thanks!

dasgarner:
@PeterMis when you review this I think there is something strange with how we have `sourceId` still inside the DataSetTicker Widget. When we separated out DataSetTicker and Ticker quite a long time ago, I thought we removed all sourceId references, but clearly not. 

So I do think @guile2912 is correct, DataSetTickers will always be non-display specific. I've created a bug: https://github.com/xibosignage/xibo/issues/2426

PeterMis:
Thank you for your PR, what you did is correct and would work.

You're right that the cache isn't correctly set on this Widget, this is due to leftover code that checks the `sourceId` as Dan has mentioned.

We will change it so DataSetTicker cache will always be display specific and remove the unnecessary code from that function. 

I will close this PR and for your reference this is my PR with said change - https://github.com/xibosignage/xibo-cms/pull/1067

Thank you for bringing this to our attention!

PR #1072, opened 2021-03-04
3.0.0 rc2

PeterMis:


web/index.php
```diff
@@ -89,8 +89,7 @@
 //
 // Middleware (onion, outside inwards and then out again - i.e. the last one is first and last);
 //
-// Handle additional Middleware
-\Xibo\Middleware\State::setMiddleWare($app);
+
```
dasgarner, line 92: As discussed offline, I think this change will mean that we can't use the store/etc in custom middleware. From memory that was certainly the case with the Authentication middleware.

We can actually use the same strategy for custom middleware as we do for authentication middleware and provide an `addRoutes()` method when we initialise in state?

```
if (method_exists($object, 'addRoutes')) {
    $object->addRoutes($app);
}
```

PeterMis: I have added that and moved the custom middleware just after Authentication middleware (just before in terms of execution). 
This way we will able to use store etc and any public/csrf excluded routes will be in the request before we hit Authentication middleware.

dasgarner (approved):

PR #1076, opened 2021-03-08
Feature/menu boards

PeterMis:


dasgarner (commented):
I've checked through file edits and all seems good to me. I've not checked new files.

My only comment so far is that I think we need to add some Phpunit/cypress tests before we go further.

Great work so far.

I'll try this week and pull it locally to give feedback on the UI/etc.

dasgarner (approved):

PR #1084, opened 2021-03-15
Feature/2.3.10

maurofmferrao:


lib/Widget/Countdown.php
```diff
@@ -71,7 +71,7 @@ public function installOrUpdate($moduleFactory)
             $module->regionSpecific = 1;
             $module->renderAs = 'html';
             $module->schemaVersion = $this->codeSchemaVersion;
-            $module->defaultDuration = 5;
+            $module->defaultDuration = 60;
```
dasgarner, line 74: What triggered the change to the default duration?

dasgarner (commented):
This looks good - with one comment.

Before we merge into release23, it would be good to cherry pick and PR into release30 so that we can deploy internally to staging and have the OP test the preview.

lib/Widget/Countdown.php
maurofmferrao: I just noticed ( while testing the video loop duration) that the default duration for an added countdown in the designer was 60 seconds, and the preview wasn't matching that value, should I revert this back to 5?

dasgarner (approved):

PR #1094, opened 2021-03-31
Wolf 2.3.10

PeterMis:


views/layout-form-copy.twig
```diff
@@ -31,6 +31,12 @@
             <form id="layoutCopyForm" class="XiboForm form-horizontal" method="post" action="{{ urlFor("layout.copy", {id: layout.layoutId}) }}")>
                 <div class="tab-content">
                     <div class="tab-pane active" id="general">
+                        {% if layout.publishedStatusId == 2 %}
+                            {% set warning %} {% trans "Copying this Layout will create an exact copy of the last time this Layout was Published. Any changes already made to this Layout will be lost.
```
dasgarner, line 35: I don't think this is correct, changes made to the original layout will not be lost, they just won't be copied.

Suggest:

> Copying this Layout will create an exact copy of the last time this Layout was Published. Any changes made to this Layout while it has been a Draft will not be copied. Publish the Layout before making a copy if the Draft changes should be included in the copy.

dasgarner (approved):

PR #1095, opened 2021-03-31
Bugfix/3.0.0

PeterMis:
- Adjusted Cherry picks from 2.3.10
- Fixed DataSet csv import
- Fixed DataSet grid filtering on load
- Fixed Layout grid filtering by display groups
- Add imported tag in install migration
- Include system tags in auto complete tag search

### DIFF
--- a/lib/Factory/TagFactory.php
+++ b/lib/Factory/TagFactory.php
@@ -383,8 +383,8 @@ class TagFactory extends BaseFactory
         }
 
         //isSystem filter, by default hide tags with isSystem flag
-        if ($sanitizedFilter->getCheckbox('isSystem') !== null && $sanitizedFilter->getInt('allTags') !== 1) {
-            $body .= " AND `tag`.isSystem = :isSystem ";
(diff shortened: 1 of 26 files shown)


lib/Factory/TagFactory.php
```diff
@@ -383,10 +383,9 @@ public function query($sortOrder = null, $filterBy = [])
         }
 
         //isSystem filter, by default hide tags with isSystem flag
-        if ($sanitizedFilter->getCheckbox('isSystem') === 1) {
-            $body .= " AND `tag`.isSystem = 1 ";
-        } else {
-            $body .= " AND `tag`.isSystem = 0 ";
+        if ($sanitizedFilter->getCheckbox('isSystem') !== null && $sanitizedFilter->getInt('allTags') !== 1) {
```
dasgarner, line 386: I don't think `getCheckbox` can return null - only true/false?

EDIT: what I mean is, won't this statement always be true because true !== null and false is also !== null :-)

dasgarner (commented):
Just one question :smile:

dasgarner (approved):

PR #1096, opened 2021-04-05
Update MySQL from 5.6 to 5.7

jokay:
Updates MySQL from **5.6** to **5.7**, as support for **5.6** ended on Feb 2021, see [here](https://en.wikipedia.org/wiki/MySQL#Release_history).

Relates https://github.com/xibosignage/xibo/issues/2469, https://github.com/xibosignage/xibo-docker/pull/136.

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -13,7 +13,7 @@ jobs:
           fetch-depth: 1
       - name: Pull
         run: |
-          docker pull mysql:5.6
+          docker pull mysql:5.7
           docker pull xibosignage/xibo-xmr:latest
       - name: Get Diff
         run: |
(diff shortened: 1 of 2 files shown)


dasgarner:
Thank you for your PR - you're right that we should consider moving to MySQL 5.7.

There are a few things which do not work as a result that we have to investigate - an issue has been created to track that: https://github.com/xibosignage/xibo/issues/2469

jokay:
@dasgarner I wonder why you waited until the EOL of MySQL 5.6, as MySQL 5.7 exists since 21 October 2015? 🧐

jokay:
> Thank you for your PR - you're right that we should consider moving to MySQL 5.7.

@dasgarner I wonder why you waited until the EOL of MySQL 5.6, as MySQL 5.7 exists since 21 October 2015? 🧐

dasgarner:
As a rule, we try to support the earliest version of any dependencies that we can, assuming its safe and sensible to do so. We did not want to exclude all users who are fixed at 5.6 if it isn't necessary to do so.

dasgarner:
We have done some work to fix the various issues with 5.7 and you can find a new PR for that here: https://github.com/xibosignage/xibo-cms/pull/1099

As you can see, its not completely solved yet, but we will continue to work on it and hopefully have a 2.3.10 release out soon which ships with 5.7 as standard and fully supports it. We will also look at putting in a test suite step to check for MySQL 8.0 (although that will need some configuration or PHP 7.4).

I'll leave your PR open until we've merged the above linked PR.

Thanks again for bringing the support EOL to our attention.

jokay:
Thank you very much for all the effort you put in the further development of Xibo.

dasgarner:
Closing this out - 2.3.10 has been released with 5.7 enabled. Thanks.

PR #1106, opened 2021-04-14
Stats: export error on loading sanitizer with BSONDocument[]

dasgarner:


lib/Storage/TimeSeriesMongoDbResults.php
```diff
@@ -58,13 +58,14 @@ public function __construct($cursor = null)
...
     {
+        $this->object->setTypeMap(['root' => 'array']);
         return $this->object->toArray();
     }
 
     /** @inheritDoc */
     public function getIdFromRow($row)
     {
-        return $row['id'];
+        return (string)$row['id'];
```
ifarzana, line 68: this (string) is not required. Id comes as string as well after setting the typeMap to array

ifarzana (approved):

PR #1111, opened 2021-04-27
Validate if a statId is a mongo ID

ifarzana:


lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -492,7 +491,7 @@ public function getStats($filterBy = [])
         }
 
         if (!empty($statId)) {
```
dasgarner, line 493: I am not 100% on whether this works anymore now that we're expecting this to be an ObjectId (it would depend if that class evaluates to false).

I think `$statId !== null` would be better

While there perhaps we can also tidy up the space in this array definition `[ '$gt` ?

dasgarner (approved):

dasgarner (approved):

PR #1112, opened 2021-04-27
Validate Mongo StatId xibosignage/xibo#2494 (v3)

ifarzana:


dasgarner:
Apologies, but this needs to be done against `release30` please. 
Closing this due to conflicts (it would be easier to cherry pick your change into release30 I think - or do it again and copy/paste)

PR #1141, opened 2021-05-27
Fix xibosignage/xibo#2519

alexharrington:


alexharrington:
Will need to be put in to 2.3 as well as 3.x please

dasgarner:
I'm just going to change the base of this to release30 so that it gets addressed for 3.0 - we can then cherry pick into release23 and merge release30 into develop in due course.

dasgarner:
Ah, that doesn't work as you started with develop... we will sort this out tomorrow.
Thanks

dasgarner (approved):

PR #1144, opened 2021-06-03
Bugfix/3.0.0

PeterMis:


lib/Controller/Display.php
```diff
@@ -2275,6 +2275,10 @@ public function moveCms(Request $request, Response $response, $id)
                 throw new InvalidArgumentException(__('Provided CMS URL is invalid'), 'newCmsUrl');
             }
 
+            if (!v::stringType()->length(1, 1000)->validate($newCmsAddress)) {
```
dasgarner, line 2278: Doesn't a length from 1 to 1000 mean that we must enter a new CMS URL?

dasgarner (commented):
Just a quick check on something

lib/Controller/Display.php
PeterMis: Well, yes, couple of lines above that we check if it's empty anyway ie it cannot be empty (or invalid or now over 1000 characters).

Is that not correct?

dasgarner: Understood, makes sense if its always required! :+1:

dasgarner (approved):

PR #1145, opened 2021-06-03
Agenda view and final tweaks

maurofmferrao:
resolves xibosignage/xibo#2464


dasgarner (approved):

dasgarner:
Merge with PHPCS errors as the file in question is a copy of the existing calendar widget file before changes.

PR #1150, opened 2021-06-09
Bugfix/3.0.0

PeterMis:


lib/Factory/BaseFactory.php
```diff
@@ -183,7 +183,7 @@ public function viewPermissionSql($entity, &$sql, &$params, $idColumn, $ownerCol
 
         // Check the whether we need to restrict to the DOOH user.
         // we only do this for entities which have an owner, and only if the user check hasn't been disabled.
-        if ($ownerColumn !== null && $performUserCheck) {
+        if ($ownerColumn !== null && $performUserCheck && $user->showContentFrom != 3) {
```
dasgarner, line 186: I think there might be a regression here... but I am not sure.

Before this change we would always restrict userTypeId 4 (DOOH usertype) so that they did not see standard content, regardless of showContentFrom which should not effect them. After this change we would only do that if they did not have showContentFrom set to both - but the DOOH user should always be restricted to DOOH content?

PeterMis: It was under the assumption that DOOH user will never have showContentFrom set to 3 - for this user type it could only be edit via API, which is unlikely to happen.

That said, we could just add elseif with ``` $user->showContentFrom != 3``` instead of else to restrict to standard content, if that's better?

dasgarner: I think that maybe would be better - it is concrete then and can't be worked around under any circumstances.

dasgarner (approved):

PR #1161, opened 2021-06-21
Pull 3.0 into 3.1

dasgarner:


PeterMis (approved):
Looks good to me, but perhaps we could wait for my bugfix 3.0.0 PR to be merged into release30 and then add it here as well?

PeterMis (approved):

PR #1194, opened 2021-07-08
DataSetView : post upgrade fix

PeterMis:
allow colorPicker values  to be applied with overrideTemplate and basic template


maurofmferrao (approved):
All looks good!

PR #1217, opened 2021-07-22
Bugfix/3.0.2

PeterMis:


lib/Widget/NotificationView.php
```diff
@@ -256,7 +256,8 @@ private function getNotifications($isPreview, $displayId = null)
                         break;
 
                     case '[Date]':
-                        $replace = Carbon::createFromTimestamp($notification->releaseDt)->format($dateFormat);
+                        Carbon::setLocale($this->getConfig()->getSetting('DEFAULT_LANGUAGE', 'en'));
```
dasgarner, line 259: This certainly addresses this particular instance of locale, but in v2 we put the whole CMS in the right locale rather than individual items as we seem to do now.

I'm not sure on the reasoning behind that?

dasgarner (approved):
Thanks.

PR #1218, opened 2021-07-26
Wolf 2.3.12

PeterMis:


lib/Controller/DayPart.php
```diff
@@ -540,6 +540,10 @@ public function delete($dayPartId)
     {
         $dayPart = $this->dayPartFactory->getById($dayPartId);
 
+        if ($dayPart->isSystemDayPart()) {
+            throw new \InvalidArgumentException('Cannot delete system dayPart', 'dayPartId');
```
dasgarner, line 544: Is this the right type of exception? and should it be translated?

lib/Controller/User.php
```diff
@@ -861,8 +862,17 @@ public function delete($userId)
     {
         $user = $this->userFactory->getById($userId);
 
-        if (!$this->getUser()->checkDeleteable($user))
+        if (!$this->getUser()->checkDeleteable($user)) {
             throw new AccessDeniedException();
+        }
+
+        if ($this->getUser()->userId === $userId) {
+            throw new \InvalidArgumentException(__('Cannot delete your own User from the CMS.'));
```
dasgarner, line 870: Is this the right type of InvalidArgumentException? Don't we use the support library?

lib/Entity/DayPart.php
```diff
@@ -254,6 +266,10 @@ public function save($options = [])
      */
     public function delete()
     {
+        if ($this->isSystemDayPart()) {
+            throw new \InvalidArgumentException('Cannot delete system dayParts');
```
dasgarner, line 270: Exception/Translation

dasgarner (commented):
Some queries

dasgarner (approved):

PR #1228, opened 2021-08-02
Widget: Provide a user interface for creating/editing templates

maurofmferrao:
relates to xibosignage/xibo#2215

- Bottom bar added
- Toggleable timeline
- Properties Panel revamp


views/layout-form-edit.twig
```diff
@@ -65,6 +65,14 @@
                         {% set helpText %}{% trans "The Name of the Layout - (1 - 50 characters)" %}{% endset %}
                         {{ forms.input("name", title, layout.layout, helpText) }}
 
+                         {% set title %}{% trans "Start Time" %}{% endset %}
```
dasgarner, line 68: I'm not sure what this does?

maurofmferrao: That's clearly not supposed to be there, good catch! :P

PR #1231, opened 2021-08-10
3.1.0 alpha

PeterMis:


db/migrations/20210806122814_add_number_of_items_to_playlist_migration.php
```diff
@@ -0,0 +1,36 @@
...
+            ],
+            [
+                'setting' => 'DEFAULT_DYNAMIC_PLAYLIST_MAXNUMBER_LIMIT',
+                'value' => 100,
+                'userSee' => 1,
+                'userChange' => 1
+            ]
+        ])->save();
+
+        $this->execute('UPDATE `playlist` SET maxNumberOfItems = 30 WHERE isDynamic = 1');
```
dasgarner, line 34: Do we want to try and count them do you think? So that we don't break use cases on upgrade?

Untested SQL incoming...

```
SELECT COUNT(*) AS cnt, widget.playlistId
  FROM `widget` 
 WHERE playlistId IN (SELECT playlistId FROM `playlist` WHERE isDynamic = 1)
GROUP BY widget.playlistId
ORDER BY COUNT(*) DESC
```

Could we then use `min($record[0]['cnt'], 30)` and `min($record[0]['cnt'], 100)` respectively?

PR #1232, opened 2021-08-10
Bugfix/3.0.3

PeterMis:


dasgarner (approved):
You found some nice little extra :bug: there too!

PR #1238, opened 2021-08-13
3.1.0 alpha

PeterMis:
Cycle Based Playback


lib/Entity/Layout.php
```diff
@@ -1474,6 +1474,10 @@ public function toXlf()
                     $mediaNode->setAttribute('playlist', $widget->playlist);
                     $mediaNode->setAttribute('displayOrder', $widget->displayOrder);
                     $mediaNode->setAttribute('parentWidgetId', $widget->tempId);
+                    $mediaNode->setAttribute('isRandom', $widget->getOptionValue('isRandom', 0));
+                    $mediaNode->setAttribute('playCount', $widget->getOptionValue('playCount', 0));
+                    $mediaNode->setAttribute('groupKey', $widget->playlistId);
```
dasgarner, line 1479: Thinking about this a bit more, I wonder if actually what we're after here is the widgetId of the sub-playlist widget? Otherwise don't we risk not grouping the right things if the sub-playlist has more than one playlist in it... :sweat_smile:

dasgarner (approved):

PR #1241, opened 2021-08-25
Widget: Provide a user interface for creating/editing templates

maurofmferrao:
relates to xibosignage/xibo#2215

 - Inline snippets and fullscreen to viewer
 - Tools to bottom bar
 - Timeline fixed
 - Cypress updated
 - Layout designer renamed to layout editor


lib/Widget/Agenda.php
```diff
@@ -85,7 +85,7 @@ public function installFiles()
     }
 
     /** @inheritdoc */
-    public function layoutDesignerJavaScript()
+    public function layoutEditorJavaScript()
```
dasgarner, line 88: I appreciate why you've done this, but it has downstream implications we should consider before we do it. For example it would break any custom modules developers have written which provide this method. It would be pretty annoying for them to have to upgrade their custom code just for a naming change.

PR #1258, opened 2021-11-01
Feature/3.0.4

maurofmferrao:


dasgarner (approved):
This all looks good

dasgarner:
PHPunit failure is most likely to be a DST issue - rerunning just in case, but it not this will need looking at.

PR #1273, opened 2021-11-23
Layout Editor Preview: Broken video playback

maurofmferrao:
fixes xibosignage/xibo#2666

### DIFF
--- a/ui/src/preview/html-preview.js
+++ b/ui/src/preview/html-preview.js
@@ -639,7 +639,7 @@ function media(parent, id, xml, options, preload) {
     self.singlePlay = false;
     self.timeoutId = undefined;
     self.ready = true;
-    self.checkIframeStatus = true;
+    self.checkIframeStatus = false;
 
     if (self.render == undefined)
         self.render = "module";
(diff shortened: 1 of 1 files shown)


ui/src/preview/html-preview.js
```diff
@@ -812,7 +817,8 @@ function media(parent, id, xml, options, preload) {
         preload.addFiles(tmpUrl);
         
         self.iframe = $('<video id="' + self.containerName + '-vid" preload="auto" ' + ((self.options["mute"] == 1) ? 'muted' : '') + ' ' + (loop ? 'loop' : '') + '><source src="' + tmpUrl + '">Unsupported Video</video>');
-        
+        self.checkIframeStatus = false;
```
dasgarner, line 820: Would I be right in saying that we should also set this to false on flash? 

Is it also clearer to set it to false for image/audio even though those don't have an `iframe` set?

maurofmferrao: Yeah, I think we also need to do that :)

dasgarner (approved):

PR #1276, opened 2021-12-09
Bugfix/3.0.5

PeterMis:


lib/XTR/MaintenanceRegularTask.php
```diff
@@ -405,34 +405,33 @@ private function publishLayouts()
...
-                                ]);
-                            } finally {
-                                $this->layoutFactory->concurrentRequestRelease($layout, true);
+                        // publish the layout
+                        $layout = $this->layoutFactory->concurrentRequestLock($layout, true);
+                        try {
+                            $draft = $this->layoutFactory->getByParentId($layout->layoutId);
+                            if ($draft->status === ModuleWidget::$STATUS_INVALID
+                                && isset($draft->statusMessage)
+                                && (count($draft->getStatusMessage()) > 1 || count($draft->getStatusMessage()) === 1 && !in_array(__('Empty Region'), $draft->getStatusMessage()))
```
dasgarner, line 418: I see what you've done here and why, but the "magic" nature of it worries me. For example if we ever change the "Empty Region" error message, this then breaks.

We may need to give this some more thought - an idea which comes to mind (completely unverified) is whether we call `xlfToDisk` on the `$layout`. We could catch an error from there easily enough?

Another potential is a helper method which just checks for empty regions and returns a boolean?

dasgarner (changes requested):

dasgarner (approved):

PR #1284, opened 2021-12-17
Prevent JavaScript SyntaxError if vars are undefined

JasonCook599:
Resolves xibosignage/xibo#2698
Resolves xibosignage/xibo#2699

### DIFF
--- a/views/dashboard-status-page.twig
+++ b/views/dashboard-status-page.twig
@@ -450,7 +450,7 @@
             }
         });
 
-        var displayData {% if displayStatus|raw %}= {{ displayStatus|raw }}{% endif %};
+        var displayMediaStatusData {% if displayMediaStatus|raw %}= {{ displayMediaStatus|raw }}{% endif %};
 
         var displayContentChart = new Chart($("#displayContentChart"), {
             type: 'doughnut',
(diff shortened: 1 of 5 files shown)


dasgarner:
Thanks Jason, I've created two issues to track these changes so we can more easily see what was changes and why (and to pull into the release notes).

https://github.com/xibosignage/xibo/issues/2698
https://github.com/xibosignage/xibo/issues/2699

I'll review the changes shortly! Happy New Year!

views/dashboard-status-page.twig
```diff
@@ -450,7 +450,7 @@
             }
         });
 
-        var displayMediaStatusData = {{ displayMediaStatus|raw }};
+        var displayData {% if displayStatus|raw %}= {{ displayStatus|raw }}{% endif %};
```
dasgarner, line 453: Should this actually be as below?

```suggestion
        var displayMediaStatusData {% if displayMediaStatus|raw %}= {{ displayMediaStatus|raw }}{% endif %};
```

dasgarner (changes requested):

dasgarner:
There is one bit that didn't quite make sense to me - the rest looks great!

We're really happy to have your contributions, please let me know if I can help you in anyway.

views/dashboard-status-page.twig
JasonCook599: You are correct. I think I just copy-pasted incorrectly.

dasgarner (approved):

PR #1308, opened 2022-01-26
3.1.0 beta - Pixabay fixes

PeterMis:


lib/Controller/Library.php
```diff
@@ -2616,16 +2621,39 @@ public function connectorImport(Request $request, Response $response)
...
+                // Success
+                // if we have pictureId (pixabay) get the thumbnail for the video and save it to library
+                foreach ($importQueue as $import) {
+                    /** @var ProviderImport $import */
+                    if ($import->media->getId() === $media->getId() && $media->mediaType === 'video' && !empty($import->searchResult->pictureId)) {
+                        try {
+                            $client = new Client();
+                            $client->request(
+                                'GET',
+                                str_replace('pictureId', $import->searchResult->pictureId, 'https://i.vimeocdn.com/video/pictureId_960x540.png'),
```
dasgarner, line 2635: Could we do this string replace inside the search results instead? that way we could in theory show a thumbnail in the UI if we wanted to?

dasgarner (changes requested):
I've got a few comments on this one

dasgarner (dismissed):

dasgarner (approved):

PR #1316, opened 2022-02-02
Library duration on edit changes

dasgarner:


lib/Controller/Library.php
```diff
@@ -1313,8 +1313,7 @@ public function edit(Request $request, Response $response, $id)
         // Should we update the media in all layouts?
         if ($sanitizedParams->getCheckbox('updateInLayouts') == 1 || $media->hasPropertyChanged('enableStat')) {
             foreach ($this->widgetFactory->getByMediaId($media->mediaId, 0) as $widget) {
-                /* @var Widget $widget */
-                $widget->duration = $media->duration;
+                $widget->calculateDuration($this->moduleFactory->createWithWidget($widget));
```
dasgarner, line 1316: This is the main change here.

lib/Widget/ModuleWidget.php
```diff
@@ -670,13 +670,15 @@ final public function getDuration($options = [])
             'real' => false
         ], $options);
 
-        if ($options['real']) {
+        $isRegionSpecific = ($this->module !== null && $this->module->regionSpecific === 1);
+
+        if ($options['real'] && !$isRegionSpecific) {
```
dasgarner, line 675: Only use real if we're a library media

PeterMis (approved):

PR #1323, opened 2022-02-07
3.1.0 beta

PeterMis:


views/authed-user-menu.twig
```diff
@@ -13,6 +13,10 @@
             <a class="dropdown-item XiboFormButton" href="{{ url_for("user.edit.profile.form") }}" title="{% trans "Edit Profile" %}">{% trans "Edit Profile" %}</a>
         {% endif %}
 
+        {% if currentUser.featureEnabled("application.view") %}
```
dasgarner, line 16: Is this definitely the right check, or should a user always have access to see what applications they have authorised?

dasgarner (changes requested):
I've got two comments - the meat of the PR is fine!

dasgarner (approved):

PR #1333, opened 2022-02-24
3.1.0

PeterMis:


lib/Controller/Template.php
```diff
@@ -135,8 +135,13 @@ function grid(Request $request, Response $response)
                 ]);
             }
 
-            if ($this->isApi($request))
-                break;
+            // Published status, draft with set publishedDate
+            $template->publishedStatusFuture = __('Publishing %s');
```
dasgarner, line 139: Are we just using this to push a message through to the front end? If so, couldn't we just put a trans into the front end instead?

dasgarner (approved):

PR #1336, opened 2022-02-28
Usage report improvements

dasgarner:
xibosignage/xibo#2764


maurofmferrao (approved):
All seems good to me!

PR #1371, opened 2022-04-25
Added Missing php7-sockets Library for WOL

itshorty:
WOL was not working without it.


dasgarner:
Thank you for this, much appreciated!
Would you be willing to add the same to `Dockerfile.ci` and `Dockerfile.dev`?

itshorty:
Shure - added them both

dasgarner (approved):

dasgarner:
Great, thank you!

PR #1375, opened 2022-05-09
Dispatcher: the entity dispatcher wasn't hooked up in all cases.

dasgarner:
Push the dispatcher into the base dependencies service.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
 	"prefer-stable": true,
   "require": {
     "php": ">=7.2.9",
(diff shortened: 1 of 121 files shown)


dasgarner:
This is probably more drastic than necessary and could have been solved by selectively calling `useDispatcher()` on the entities that needed it. However as we increase our usage of events, it seems sensible to make sure that the dispatcher is always available throughout the stack.

PeterMis (approved):

PR #1401, opened 2022-07-12
Hotfix: 3.1.4 fix for select boxes not closing

dasgarner:


dasgarner:
Aborting the test suite as it didn't catch the error that this solves anyway. Acknowledging that we can't wait 30 minutes for the test suite to run.

PR #1405, opened 2022-07-20
Map view for displays

ifarzana:
https://github.com/xibosignage/xibo/issues/2433#event-6958826223


ui/bundle_vendor.js
```diff
@@ -127,6 +127,16 @@ require('leaflet-search');
 window.L = require('leaflet');
 window.leafletPip = require('@mapbox/leaflet-pip');
 
+delete L.Icon.Default.prototype._getIconUrl;
+L.Icon.Default.mergeOptions({
+    iconRetinaUrl: 'https://assets.xibo-adspace.com/map/marker-icon-2x.png',
```
ifarzana, line 132: @dasgarner , can you suggest what should be the url for these map icons?

ifarzana: @maurofmferrao, Can we go through this in the morning please?

composer.json
```diff
@@ -107,7 +107,9 @@
     "doctrine/annotations": "^1",
     "squizlabs/php_codesniffer": "3.*",
     "exussum12/coverage-checker": "^0.11.2",
-    "micheh/phpcs-gitlab": "^1.1"
+    "micheh/phpcs-gitlab": "^1.1",
+    "jmikola/geojson": "^1.0",
```
dasgarner, line 111: Unless I am mistaken, we aren't doing any geo-stuff on the backend (we don't have any polygons, etc, or any calculations to make).

Can you comment why we need these two libraries?

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
 
         // Call to render the template
         $this->getState()->template = 'display-page';
+
+        $mapConfig = [
+        'setArea' => [
+            'lat' => 51.505,
```
dasgarner, line 191: The CMS has a default lat/long which is available in global settings, we could use that here instead.

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
...
+     * @return \Psr\Http\Message\ResponseInterface|Response
+     * @throws GeneralException
+     * @throws NotFoundException
+     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     */
+    function displayMap(Request $request, Response $response)
+    {
+        $parsedQueryParams = $this->getSanitizer($request->getQueryParams());
+
+        $filter = [
```
dasgarner, line 219: Is this filter common between this method and the existing `grid` one? If it is, could we abstract it?

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
...
+            'lastAccessed' => ($parsedQueryParams->getDate('lastAccessed') != null) ? $parsedQueryParams->getDate('lastAccessed')->format('U') : null,
+            'displayGroupIdMembers' => $parsedQueryParams->getInt('displayGroupIdMembers'),
+            'orientation' => $parsedQueryParams->getString('orientation'),
+            'commercialLicence' => $parsedQueryParams->getInt('commercialLicence'),
+            'folderId' => $parsedQueryParams->getInt('folderId'),
+            'logicalOperator' => $parsedQueryParams->getString('logicalOperator'),
+        ];
+
+        // Get a list of displays
+        $displays = $this->displayFactory->query($this->gridRenderSort($parsedQueryParams), $this->gridRenderFilter($filter, $parsedQueryParams));
```
dasgarner, line 246: Do you need to read sort/filter here? We're always going to be requesting the whole lot?

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
...
+        $results = [];
+        $status = [
+            '1' => 'Up to date',
```
dasgarner, line 249: Translations :smile:

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
...
+            '2' => 'Downloading',
+            '3' => 'Out of date'
+        ];
+        foreach ($displays as $display) {
+         $geo = new Point([(double)$display->longitude, (double)$display->latitude]);
+
+            $results[] = new Feature($geo, [
+                'display' => $display->display,
+                'status' => $display->mediaInventoryStatus ? $status[$display->mediaInventoryStatus] : 'Unknown',
```
dasgarner, line 258: Translations

lib/Controller/Display.php
```diff
@@ -178,14 +185,86 @@ function displayPage(Request $request, Response $response)
...
+                'orientation' => ucwords($display->orientation),
+                'displayId' => $display->getId(),
+                'licensed' => $display->licensed,
+            ]);
+        }
+
+        return $response->withJson(new FeatureCollection($results));
```
dasgarner, line 265: Ah, this explains why you have `jmikola/geojson` included, but not the other one.

ui/bundle_vendor.js
dasgarner: Yes, I think we should download and include the icons in the `web/theme/default` folder (maybe in a `map` sub folder).

I think you might need Mauro's help to get the right URL into the vendor bundle though - we have to set something to handle installs in sub-folders.

composer.json
dasgarner: Edit: I see why you need one of them, not the other :-)

ui/src/core/xibo-cms.js
```diff
@@ -1463,6 +1465,17 @@ function dataTableDraw(e, settings) {
         $folderController.removeClass('d-none').addClass('d-inline-flex');
     }
 
+    // Move and show map button inside of the table container
+    if ($mapController.length > 0 && target.closest(".dataTables_wrapper").find('.dataTables_folder .map-controller').length == 0) {
```
dasgarner, line 1469: Mauro will know better, but I think this could be handled as a custom callback to the dataTableDraw function.

views/display-page.twig
```diff
@@ -562,6 +575,145 @@
...
+
+                    // Make an ajax request for the displays
+                    $.ajax($map.data('displaysUrl'), {
+                        method: 'GET',
+                        data: $('.XiboGrid').find(".XiboFilter form").serialize(),
+                        success: function(response) {
+                            // displays
+                            if (response.features.length > 0) {
+                                let licensedIcon = L.icon({
+                                    iconUrl: 'https://assets.xibo-adspace.com/map/marker-icon-green.png',
```
dasgarner, line 659: These need to be local

views/display-page.twig
```diff
@@ -562,6 +575,145 @@
...
+                    filename: 'Network Overview',
+                    hideControlContainer: true
+                }).addTo(map);
+
+                // Cluster marker group for individual plays
+                // create a cluster group
+                let cluster = L.markerClusterGroup({
+                    iconCreateFunction: function (cluster) {
+                        let childCount = cluster.getChildCount();
+                        let playCount = getPlayCount(cluster.getAllChildMarkers());
```
dasgarner, line 619: Get play count?

dasgarner (commented):
Just some initial comments on the BE ahead of your review with Mauro.

composer.json
ifarzana: I have removed the latter.

lib/Controller/Display.php
ifarzana: Abstracted in a public function.

views/display-page.twig
ifarzana: (－‸ლ)

ui/bundle_vendor.js
maurofmferrao: @dasgarner I suggested to have those assets in `ui/src/assets/map` instead, so we can use them in both `twig` and `js` side easily

ui/src/core/xibo-cms.js
maurofmferrao: Israt and I were a bit confused about the dataTableDraw callback, the new buttons will have the same behaviour as the Folder one, so maybe this is the best approach?

Unfortunately with Datatables, we do need to create them outside the table container and move them inside after the draw

dasgarner: As this is the only page which has a map currently, it may be more elegant to change the signature of the `dataTableDraw` function so that it can take an optional callback.

We can then pass in a callback only on the displays page, to do the map jiggle wiggle.

```
function dataTableDraw(e, settings, callBack) {
  if (callBack && callBack instanceof Function) {
    callBack();
  }
}```

Comment (maurofmferrao) [reply ui/src/core/xibo-cms.js:1469]: I see, that makes total sense! The `if` statement would prevent the code to run, but we would still be checking for those in all the other pages indeed

**Review by maurofmferrao — APPROVED** [review]
Looks good to me 👍🏻

File: cypress/e2e/displays.cy.js [hunk 12]
```diff
@@ -29,14 +29,18 @@ describe('Displays', function () {
 
         cy.visit('/display/view');
 
-        cy.get('#displays_wrapper.dataTables_wrapper').check();
+        cy.wait(500);
```
maurofmferrao, line 32: The use of wait isn't really recommended if there's any requests happening, a slower system could take more than the 500ms and make the test fail. The possible solution for this is to have a custom wait flag for a specific request on top of the test, and make the next step wait for the request to be finished.

### To define the waiting variable:
```
cy.server();
cy.route('**/path/*').as('requestName');
```

### When you need to use wait ( instead of a time in ms ):
```
cy.wait('@requestName');
```

This functionality will need to be replaced in all Cypress tests for `cy.intercept()` though, so if you want to implement that instead: https://docs.cypress.io/guides/references/migration-guide#Migrating-cy-route-to-cy-intercept

cypress/e2e/displays.cy.js
```diff
@@ -29,14 +29,18 @@ describe('Displays', function () {
...
+
+        cy.get('#displays_wrapper.dataTables_wrapper').should('have.length', 1);
```
maurofmferrao, line 34: Instead of checking for the element length, there's a much cleaner approach, with:
```
cy.get('#displays_wrapper.dataTables_wrapper').should('be.visible');
```

maurofmferrao (dismissed):

maurofmferrao (approved):

ifarzana:
@dasgarner This is ready to be merged once you approve, thanks

dasgarner (approved):
This all looks great and I think we should merge to make it available in the `develop` container for users to try.

I do have a questions for you to consider next...

What happens to displays which do not have a lat/long recorded? Should we add a filter for them and optionally set their lat/long to the CMS default lat/long if they don't have one?

PR #1408, opened 2022-07-29
Map view display

ifarzana:


lib/Factory/DisplayFactory.php
```diff
@@ -224,8 +224,8 @@ public function query($sortOrder = null, $filterBy = [])
                   display.broadCastAddress,
                   display.secureOn,
                   display.cidr,
-                  ' . $functionPrefix . 'X(display.GeoLocation) AS latitude,
-                  ' . $functionPrefix . 'Y(display.GeoLocation) AS longitude,
+                  IFNULL( ' . $functionPrefix . 'X(display.GeoLocation), ' . $this->config->getSetting('DEFAULT_LAT'). ') AS latitude,
```
dasgarner, line 227: I don't think we can do this here, because that would mean that the displays in the grid also show as the default lat/long. It is very useful to see which displays actually have a lat/long, so we don't want to affect the grid.

Instead you could output the default lat/long to the page, and then handle this client side when you pass the pin to the map?

lib/Factory/DisplayFactory.php
```diff
@@ -296,8 +296,8 @@ public function query($sortOrder = null, $filterBy = [])
         // Filter by map bound?
         if ($parsedBody->getString('bounds') !== null) {
             $coordinates = explode(',', $parsedBody->getString('bounds'));
-            $body .= ' AND  ' . $functionPrefix . 'X(display.GeoLocation)  BETWEEN ' . $coordinates['1'] . ' AND ' . $coordinates['3'] .
-                ' AND  ' . $functionPrefix . 'Y(display.GeoLocation) BETWEEN  ' . $coordinates['0'] . ' AND ' . $coordinates['2'] . ' ';
+            $body .= ' AND IFNULL( ' . $functionPrefix . 'X(display.GeoLocation), ' . $this->config->getSetting('DEFAULT_LAT'). ')  BETWEEN ' . $coordinates['1'] . ' AND ' . $coordinates['3'] .
```
dasgarner, line 299: I think this does work here though, because we're asking for displays within a boundary and it makes sense to include those ones without a lat/long as the cms default.

dasgarner (changes requested):
Just a comment about always defaulting the output of the display factory with a lat/long

dasgarner (approved):

PR #1410, opened 2022-08-03
Map Resize and Legend

ifarzana:


ui/src/core/xibo-cms.js
```diff
@@ -3745,10 +3745,19 @@ function initJsTreeAjax(container, id, isForm, ttl)
...
+        if (typeof refreshDisplayMap === "function") {
+            refreshDisplayMap();
+        }
+    }
+
     reload = (typeof reload == 'undefined') ? true : reload;
     // Shrink table to ease animation
     if($('#grid-folder-filter').is(":hidden")) {
         $('#datatable-container').addClass('col-sm-10').removeClass('col-sm-12');
+        refreshDisplayMap();
```
maurofmferrao, line 3760: Shouldn't this one be `resizeDisplayMap()` instead?

ui/src/core/xibo-cms.js
```diff
@@ -3760,6 +3769,7 @@ function adjustDatatableSize (reload) {
 
             // if the folder tree is hidden, then make it so datatable can take whole available width
             $('#datatable-container').addClass('col-sm-12').removeClass('col-sm-10');
+            refreshDisplayMap();
```
maurofmferrao, line 3772: Same as above comment

ui/src/core/xibo-cms.js
ifarzana: yeah, you are right, my mistake, thanks

maurofmferrao (dismissed):
Looks good to me 👍🏻

maurofmferrao (commented):
Looks good to me!

maurofmferrao (approved):

PR #1417, opened 2022-08-31
Convert widgets

ifarzana:


modules/audio.xml
```diff
@@ -0,0 +1,44 @@
...
+    <name>Audio</name>
+    <author>Core</author>
+    <description>Upload Audio files to assign to Layouts</description>
+    <class></class>
+    <type>audio</type>
+    <dataType></dataType>
+    <schemaVersion>1</schemaVersion>
+    <assignable>1</assignable>
+    <regionSpecific>0</regionSpecific>
+    <renderAs>html</renderAs>
```
dasgarner, line 32: The older `render_as` column was null, but I think it would be clearer here to set this to `native`

modules/audio.xml
```diff
@@ -0,0 +1,44 @@
...
+  ~ Xibo is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU Affero General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<module>
+    <id>core-audio</id>
```
dasgarner, line 22: Once you're happy with a module, you can delete the old code. So that would be any Widget class file, form, designer javascript, json file, etc.

modules/hls.xml
```diff
@@ -0,0 +1,103 @@
...
+            <title>Mute?</title>
+            <helpText>Should the video be muted?</helpText>
+            <default>0</default>
+        </property>
+    </properties>
+    <preview></preview>
+    <stencil>
+        <twig>
+            <![CDATA[
+                <video id='video' poster='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVQYV2NgYAAAAAMAAWgmWQ0AAAAASUVORK5CYII=' {% if mute == 0 %} muted {% endif %} ></video>
```
dasgarner, line 50: We can use this opportunity to make better formatted HTML, in this case:
 - `"` instead of `'`
 - no spaces in the final output

modules/hls.xml
```diff
@@ -0,0 +1,103 @@
...
+                                            // cannot recover
+                                            hls.destroy();
+                                            break;
+                                    }
+                                }
+                            });
+                        }
+                    });
+                </script>
+            ]]>
```
dasgarner, line 100: Can we have no gap between the closing CDATA and the closing</twig> ?

modules/hls.xml
```diff
@@ -0,0 +1,103 @@
...
+        </property>
+        <property id="mute" type="checkbox">
+            <title>Mute?</title>
+            <helpText>Should the video be muted?</helpText>
+            <default>0</default>
+        </property>
+    </properties>
+    <preview></preview>
+    <stencil>
+        <twig>
```
dasgarner, line 48: Can we have no gap between the opening CDATA and the parent <twig>?

dasgarner (changes requested):
Nice, this looks good.

I have a couple of comments to address please before we merge.

modules/flash.xml
```diff
@@ -0,0 +1,46 @@
...
+    <name>Flash</name>
+    <author>Core</author>
+    <description>Upload SWF files to assign to Layouts</description>
+    <class></class>
+    <type>flash</type>
+    <dataType></dataType>
+    <schemaVersion>1</schemaVersion>
+    <assignable>1</assignable>
+    <regionSpecific>0</regionSpecific>
+    <renderAs>html</renderAs>
```
dasgarner, line 32: I am pretty sure flash is native rendering?

modules/genericfile.xml
```diff
@@ -0,0 +1,46 @@
...
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<module>
+    <id>core-genericfile</id>
+    <name>Generic File</name>
+    <author>Core</author>
+    <description>A generic file to be stored in the library</description>
+    <class></class>
+    <type>pdf</type>
```
dasgarner, line 27: I don't think this should be PDF?

modules/powerpoint.xml
```diff
@@ -0,0 +1,46 @@
...
+    <name>PowerPoint</name>
+    <author>Core</author>
+    <description>Upload a PowerPoint file to assign to Layouts</description>
+    <class></class>
+    <type>powerpoint</type>
+    <dataType></dataType>
+    <schemaVersion>1</schemaVersion>
+    <assignable>1</assignable>
+    <regionSpecific>0</regionSpecific>
+    <renderAs></renderAs>
```
dasgarner, line 32: perhaps this should be native

dasgarner (changes requested):

dasgarner (approved):
:+1:

PR #1419, opened 2022-09-01
Dashboards Widget and Connector

PeterMis:


lib/Controller/Module.php
```diff
@@ -784,13 +785,20 @@ public function editWidgetForm(Request $request, Response $response, $id)
             $templates = $module->templatesAvailable(true);
         }
 
+        // do we have special options requested?
+        // Dashboards Widget needs to pull available services from connector
+        $event = new WidgetEditOptionRequestEvent($module->widget);
+        $this->getDispatcher()->dispatch($event, $event::$NAME);
+        $options = $event->getOptions();
```
dasgarner, line 792: Can we make this return a keyed array of options? So that if we wanted to use it for more than one field we could?

dasgarner (approved):

PR #1453, opened 2022-09-26
Convert existing Widgets

ifarzana:


modules/webpage.xml
```diff
@@ -0,0 +1,153 @@
...
+                    "originalWidth":1258,
+                    "originalHeight":938,
+                    "iframeWidth":1258,
+                    "iframeHeight":938,
+                    "offsetTop":0,
+                    "offsetLeft":0,
+                    "scale":1
+                    }
+                $(document).ready(function() {
+                    $("#content").xiboLayoutScaler(options);
```
ifarzana, line 147: Dan, do we have this XiboLayoutScaler anymore?

modules/shellcommand.xml
```diff
@@ -0,0 +1,124 @@
...
+            <default>storedCommand</default>
+            <options>
+                <option name="storedCommand">Stored Command</option>
+                <option name="createCommand">Create Command</option>
+            </options>
+        </property>
+        <property id="commandCode" type="dropdown" mode="single">
+            <title>Stored Command</title>
+            <helpText>Pick a stored command</helpText>
+            <default>storedCommand</default>
```
dasgarner, line 58: I don't think we need to set a default value here?

dasgarner: also, are we missing some visibility rules?

modules/shellcommand.xml
```diff
@@ -0,0 +1,124 @@
...
+            <options>
+                <option name="storedCommand">Stored Command</option>
+                <option name="createCommand">Create Command</option>
+            </options>
+        </property>
+        <property id="useGlobalCommand" type="checkbox">
```
dasgarner, line 64: Missing some visibility rules?

modules/shellcommand.xml
```diff
@@ -0,0 +1,124 @@
...
+        <property id="launchThroughCmd" type="checkbox">
+            <title>Launch the command via Windows Command Line</title>
+            <helpText>On Windows, should the player launch this command through the windows command line (cmd.exe)? This is useful for batch files. If you try to terminate this command only the command line will be terminated.</helpText>
+            <default>1</default>
+        </property>
+    </properties>
+    <preview>
+        <twig><![CDATA[
+         {% if commandType == 'storedCommand' %}
+            <p>Stored Command: {{commandCode}}</p>
```
dasgarner, line 108: Missing translations?

modules/shellcommand.xml
```diff
@@ -0,0 +1,124 @@
...
+              <p>Android Command: {{androidCommand}}</p>
+              <p>Windows Command: {{windowsCommand}}</p>
+              <p>Linux Command: {{linuxCommand}}</p>
+              <p>webOS Command: {{webosCommand}}</p>
+              <p>Tizen Command: {{tizenCommand}}</p>
+         {% endif %}
+]]></twig>
+
+    </preview>
+    <stencil>
```
dasgarner, line 121: No need to provide this for native rendering.

modules/webpage.xml
dasgarner: Yeah we do, but its handled automatically - however, I think this is another case we've not handled in the new framework.

The old v3 widget shows an icon if the webpage is set to "open natively" and otherwise uses getResource. The new framework allows for outputting some HTML from inside <preview> or if the module render is HTML, then getResource.

For the time being I think we should just remove the preview from this widget completely (so that it shows an icon) and add it to our list of snagging to discuss.

dasgarner (changes requested):

dasgarner:
Thanks Israt, we will pull those in and amend when we have a better idea how this goes together.

PR #1455, opened 2022-09-29
Calendar module, xiboIC update and fixes

maurofmferrao:
- Calendar module added
 - xiboIC added with globar vars/methods
 - Change input type input to text
 - Save widget that reloaded property form as layout fix
 - Header input
 - Color input: set default on onfocus


modules/templates/event-static.xml
```diff
@@ -0,0 +1,1762 @@
...
+
+// Render
+$(target).find('img').xiboImageRender(properties);
+$(target).xiboLayoutScaler(properties);
+
+var runOnVisible = function() {
+    $(target).xiboTextRender(properties, items);
+};
+
+if (xiboIC.checkVisible()) {
```
dasgarner, line 109: Does this need to be in `onVisible`?

modules/templates/article-static.xml
```diff
@@ -138,18 +137,8 @@ if (properties.copyright) {
...
 
 // Render
-$(target).find('img').xiboImageRender(globalOptions);
-$(target).xiboLayoutScaler(globalOptions);
-
-var runOnVisible = function() {
-    $(target).xiboTextRender(properties, items);
-};
-
-if (xiboIC.checkVisible()) {
```
dasgarner, line 148: Do we need an `onVisible` node to replace this?

ui/src/style/common.scss
```diff
@@ -206,7 +206,7 @@ $fa-font-path: "~font-awesome/fonts";
     @extend .fa-calendar-o;
 }
 
-.module-icon-calendaradvanced {
+.module-icon-ics-calendar {
```
dasgarner, line 209: I suppose we need to think about how we do this!

dasgarner (approved):
I raised a few points to address in the next PR - you can merge this one I think!

modules/templates/article-static.xml
maurofmferrao: I forgot to update these files, I will update my branch ( and the PR ) with it

PR #1457, opened 2022-10-07
Fonts Refactor

PeterMis:
fixes xibosignage/xibo#2876


dasgarner (commented):
This looks good to me - the only thing we need to add before it can go are some Swagger docs for the new font APIs.

dasgarner (approved):

PR #1474, opened 2022-10-25
Reporting Improvement

ifarzana:


lib/Factory/DisplayFactory.php
```diff
@@ -304,6 +304,15 @@ public function query($sortOrder = null, $filterBy = [])
             $params['displayGroupId'] = $parsedBody->getInt('displayGroupId');
         }
 
+        // Restrict to members of display groups
+        if ($parsedBody->getIntArray('displayGroupIds') !== null) {
```
dasgarner, line 308: displayGroupIds can come directly from the front end, which means we need to parameterise it in the query. We can do this by having a for loop which creates multiple variables, so we end up with

```IN (:displayGroupId1, :displayGroupId2, etc)```

dasgarner (changes requested):
This looks good!
Just one comment regarding the display factory.

dasgarner (approved):

PR #1475, opened 2022-10-28
Reporting

ifarzana:


lib/Report/SummaryReport.php
```diff
@@ -922,6 +922,49 @@ private function getSummaryReportMongoDb(
             'query' => $cursorPeriodQuery
         ]);
 
+        if (count($displayIds) > 0) {
```
dasgarner, line 925: We have to duplicate this whole block to get the displayIds in?

Can't we just append to the array?

```
$matchExpr[] = [
    '$in' => [ '$displayId', $displayIds ]
];
```

lib/Report/TimeConnected.php
```diff
@@ -211,10 +211,32 @@ public function getSavedReportResults($json, $savedReport)
     /** @inheritdoc */
     public function getResults(SanitizerInterface $sanitizedParams)
     {
-        // Filter by displayId?
-        $displayIds = $this->getDisplayIdFilter($sanitizedParams);
-
-
+        $displayId = $sanitizedParams->getInt('displayId');
```
dasgarner, line 214: I think we need to take a look at this together please - we're using `displayIds` as a filter again, which is not what we want to do.

dasgarner (changes requested):
I've left two comments for us to look at please

dasgarner (approved):

PR #1476, opened 2022-11-01
Display MetaData

ifarzana:


ifarzana:
@dasgarner The PR has an issue that I could not be able to resolve, which is I keep getting 0 in the grid for custom id . Can you please check why it is like that?

db/migrations/20221101130018_add_display_meta_data_migration.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
dasgarner, line 1: Needs some copyright info here - please set up the copyright profile for your IDE so its automatically added/updated

db/migrations/20221101130018_add_display_meta_data_migration.php
```diff
@@ -0,0 +1,28 @@
...
+
+
+use Phinx\Migration\AbstractMigration;
+
+class AddDisplayMetaDataMigration extends AbstractMigration
```
dasgarner, line 6: You'll need a comment here to disable the phpcs check - we should also say what the migration is for.

db/migrations/20221101130018_add_display_meta_data_migration.php
```diff
@@ -0,0 +1,28 @@
...
+    public function change()
+    {
+        $this->table('display')
+            ->addColumn('screenSize', 'integer', ['after' => 'display', 'default' => null, 'null' => true])
+            ->addColumn('displayTypeId', 'integer', ['after' => 'display', 'default' => null, 'null' => true])
+            ->addColumn('impressionsPerPlay', 'integer', ['after' => 'lanIpAddress'])
+            ->addColumn('costPerPlay', 'integer', ['after' => 'lanIpAddress'])
+            ->addColumn('isOutdoor', 'integer', ['after' => 'lanIpAddress', 'limit' => \Phinx\Db\Adapter\MysqlAdapter::INT_TINY, 'default' => 0])
+            ->addColumn('customId', 'string', ['after' => 'lanIpAddress', 'limit' => 254])
+        ->save();
```
dasgarner, line 18: We can also add a foreign key to the display_type

db/migrations/20221101130018_add_display_meta_data_migration.php
```diff
@@ -0,0 +1,28 @@
...
+    /** @inheritDoc */
+    public function change()
+    {
+        $this->table('display')
+            ->addColumn('screenSize', 'integer', ['after' => 'display', 'default' => null, 'null' => true])
+            ->addColumn('displayTypeId', 'integer', ['after' => 'display', 'default' => null, 'null' => true])
+            ->addColumn('impressionsPerPlay', 'integer', ['after' => 'lanIpAddress'])
+            ->addColumn('costPerPlay', 'integer', ['after' => 'lanIpAddress'])
+            ->addColumn('isOutdoor', 'integer', ['after' => 'lanIpAddress', 'limit' => \Phinx\Db\Adapter\MysqlAdapter::INT_TINY, 'default' => 0])
+            ->addColumn('customId', 'string', ['after' => 'lanIpAddress', 'limit' => 254])
```
dasgarner, line 17: we want this to allow nulls and be default to null I think

lib/Controller/Display.php
```diff
@@ -573,6 +581,12 @@ function grid(Request $request, Response $response)
             $displayProfiles[$displayProfile->displayProfileId] = $displayProfile->name;
         }
 
+        // Get all Display Types
+        $displayTypes = [];
```
dasgarner, line 585: Instead of having this separate, we can have a `LEFT OUTER JOIN` in the displayFactory which sets the display type from there.

lib/Controller/Display.php
```diff
@@ -612,6 +626,7 @@ function grid(Request $request, Response $response)
 
             // Add in the display profile information
             $display->displayProfile = (!array_key_exists($display->displayProfileId, $displayProfiles)) ? $displayProfileName . __(' (Default)') : $displayProfiles[$display->displayProfileId];
+            $display->displayTypeId = $displayTypes[$display->displayTypeId];
```
dasgarner, line 629: Not sure what this is doing here? setting the ID to a string?

Don't we want a `$display->displayType` property instead?

lib/Factory/DisplayFactory.php
```diff
@@ -205,6 +205,12 @@ public function query($sortOrder = null, $filterBy = [])
               SELECT display.displayId,
                   display.display,
                   display.defaultLayoutId,
+                  display.displayTypeId,
+                  display.screenSize,
+                  display.isOutdoor,
+                  display.customId,
```
dasgarner, line 211: You will want to put these in `stringProperties` when you hydrate otherwise it will get converted to an int (we do that for all fields that end in `Id`

dasgarner (changes requested):
This looks like great progress - in particular your forms and entity changes look good.

I've made some comments to address (one of which fixes your customId issue). 

Don't forget we're going to need some updated tests too - i'd suggest a test which sets a display type, and a reference field as a good start.

dasgarner (approved):

PR #1496, opened 2022-11-11
Bug Fixes and Report Improvements [3.3.1]

ifarzana:


reports/proofofplay-report-form.twig
```diff
@@ -218,7 +218,7 @@
                                         <th>{% trans "Type" %}</th>
                                         <th>{% trans "Display ID" %}</th>
                                         <th>{% trans "Display" %}</th>
-                                        <th>{% trans "Campaign" %}</th>
+                                        <th>{% trans "Ad Campaign" %}</th>
```
dasgarner, line 221: This is not necessarily the ad campaign - it could also be a campaign of type list.

dasgarner (changes requested):
I think there is one change to be made here - otherwise it looks good!

dasgarner (approved):

PR #1498, opened 2022-11-14
Allow use of internal CA signed certificates for msmtp

dagonix:
By default, msmtp checks the validity of the certificate chain of the mail server. The new option "CMS_SMTP_CHECK_TLS_CERTIFICATE" allows use of TLS certificates signed by internal Certificate Authority that do not pass the default check. If the user sets "CMS_SMTP_CHECK_TLS_CERTIFICATE" to "NO" then the line "tls_nocertcheck" will be added to /etc/msmtprc (see  dagonix/ xibo-docker/config.env.template  commit).


dasgarner:
For this to work, I think we also need to add the environment variable to `Dockerfile` [here](https://github.com/xibosignage/xibo-cms/blob/develop/Dockerfile#L144)?

dagonix:
Yes !  You're right, I added the line there. Thank you.

docker/entrypoint.sh
```diff
@@ -289,6 +289,11 @@ then
       /bin/sed -i "s/tls_starttls .*$/tls_starttls off/" /etc/msmtprc
     fi
 
+    if [ "$CMS_SMTP_CHECK_TLS_CERTIFICATE" == "NO" ]
+    then
+      /bin/sed -i "s/tls .*$/tls_nocertcheck/" /etc/msmtprc
```
alexharrington, line 294: I think this will match the line:

```
tls on
```

and replace it with `tls_nocertcheck`?

But I think the correct syntax would be to have both? ie

```
tls on
tls_nocertcheck
```

Perhaps it might be better to have 

```
# CMS_SMTP_USE_TLS
tls on
#CMS_SMTP_CHECK_TLS_CERTIFICATE
#tls_nocertcheck
```

and then update the sed command to remove the comment if necessary?

alexharrington (changes requested):

dasgarner:
I've moved this to 3.3.2 to allow more time to ensure we get it right. Issue: https://github.com/xibosignage/xibo/issues/2939

docker/entrypoint.sh
dagonix: Yes, absolutly. Both options are necessary :
```
tls on
tls_nocertcheck
```
You are right Alex.

dagonix:
Thank you Dan, Alex is right, both options are necessary.

dasgarner:
Closing due to inactivity

PR #1511, opened 2022-12-08
Audience Connector - POP and Report

ifarzana:


lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -64,4 +111,195 @@ public function processSettingsForm(SanitizerInterface $params, array $settings)
         // TODO: Implement processSettingsForm() method.
         return $settings;
     }
+
+    /**
+     * @throws NotFoundException
+     * @throws GeneralException
+     */
+    public function onRegularMaintenance(MaintenanceRegularEvent $event)
+    {
```
dasgarner, line 120: We should only do this if the connector is enabled and if we have an API key

lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -64,4 +111,195 @@ public function processSettingsForm(SanitizerInterface $params, array $settings)
...
+                    'headers' => [
+                        'X-API-KEY' => $this->getSetting('apiKey')
+                    ],
+                ]);
+
+            $body = $response->getBody()->getContents();
+            $json = json_decode($body, true);
+            $watermark = $json['watermark'];
+        } catch (RequestException $requestException) {
+            $event->addMessage(__('Error getting watermark: '. $requestException->getMessage()));
```
dasgarner, line 134: If we can't get the watermark, should we actually stop?

dasgarner (changes requested):
Am I right in saying that the report is hidden currently? (which would correct)

lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -64,4 +111,195 @@ public function processSettingsForm(SanitizerInterface $params, array $settings)
...
+    /**
+     * @throws NotFoundException
+     * @throws GeneralException
+     */
+    public function onRegularMaintenance(MaintenanceRegularEvent $event)
+    {
+        $watermark = null;
+        // Get Watermark
+        try {
+            $response = $this->getClient()->get('https://test-exchange.xibo-adspace.com/api/audience/watermark', [
```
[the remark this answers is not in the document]
ifarzana: I am not entirely sure how to override settings in `settings-custom.php`

lib/Connector/XiboAudienceReportingConnector.php
ifarzana: Same for this, how can we validate whether the connector is enabled or not

lib/Connector/XiboAudienceReportingConnector.php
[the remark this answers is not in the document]
dasgarner: You can set "provider settings" for connectors using the following syntax:

```
$connectorSettings = [
    'xibo-ssp-connector' => [
        'serviceUrl' => 'https://test-exchange.xibo-adspace.com/api'
    ]
];
```

The array key is the connectors ID

When you're in the connector and call getSetting it will return you to provider setting, or the user setting if one has been set with the same key.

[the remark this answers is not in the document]
ifarzana: Thanks, that is done.

lib/Connector/XiboAudienceReportingConnector.php
dasgarner: The connector has an enabled flag, and you can test if the apiKey is empty or not.

dasgarner (approved):

PR #1515, opened 2022-12-12
Decrease the character limit in Task lastRunMessage

ifarzana:


ifarzana:
Alex mentioned he does not have access to the customer CMS  now, but his local db has VARCHAR(254), so the truncate condition should handle message more than having 254 characters.

dasgarner (approved):

PR #1520, opened 2022-12-13
Connector Reports

ifarzana:


lib/Middleware/ListenersMiddleware.php
```diff
@@ -88,6 +90,13 @@ public static function setListeners(App $app)
             ->useLogger($c->get('logger'))
             ->registerWithDispatcher($dispatcher);
 
+        // Listen for events that affect reports
```
dasgarner, line 93: I think the listener should be in the Audience Reporting Connector?

views/xibo-audience-connector-form-settings.twig
```diff
@@ -27,7 +27,13 @@
 {% block callBack %}audienceFormOpen{% endblock %}
 {% block connectorFormFields %}
 
-    <h4>Coming soon</h4>
-    <p>We're working on bringing you this exciting connector in the next update.</p>
+    {% if not interface.isProviderSetting("apiKey") %}
+    <h4>Settings</h4>
+    <p>Your API key allows for secure communication between the CMS and the Xibo audience service. It is used
```
dasgarner, line 32: I think we can translate this message, and update it to reflect what it does (analyse your proof of play data for Ad Campaigns and retrieve reports).

lib/Event/ConnectorReportEvent.php
```diff
@@ -0,0 +1,42 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+namespace Xibo\Event;
+
+class ConnectorReportEvent extends Event
```
dasgarner, line 24: It would be good to have a brief comment here explaining what this event is used for (how its called and why)

lib/Middleware/ListenersMiddleware.php
```diff
@@ -38,6 +38,8 @@
 use Xibo\Event\SystemUserChangedEvent;
 use Xibo\Event\UserDeleteEvent;
 use Xibo\Listener\CampaignListener;
+use Xibo\Event\ConnectorReportEvent;
```
dasgarner, line 41: These can be removed now I think?

dasgarner (changes requested):

lib/Middleware/ListenersMiddleware.php
ifarzana: ConnectorReportEvent is needed as it sets the connector reports object

dasgarner: But it isn't used in this middleware?

lib/Service/ReportService.php
```diff
@@ -124,6 +145,12 @@ public function listReports()
 
         $this->log->debug('Reports found in total: '.count($reports));
 
+        // Get reports that are allowed by connectors
+        $event = new ConnectorReportEvent();
```
ifarzana, line 149: @dasgarner I am getting an error in  test ReportScheduleDataTest because of the dispatched event, how do we resolve this in the test https://github.com/ifarzana/xibo-cms/blob/campaignPop/tests/integration/ReportScheduleDataTest.php#L146?

ifarzana (commented):
https://github.com/ifarzana/xibo-cms/blob/campaignPop/tests/integration/ReportScheduleDataTest.php#L146

views/xibo-audience-connector-form-settings.twig
```diff
@@ -27,7 +27,13 @@
 {% block callBack %}audienceFormOpen{% endblock %}
 {% block connectorFormFields %}
 
-    <h4>Coming soon</h4>
-    <p>We're working on bringing you this exciting connector in the next update.</p>
+    {% if not interface.isProviderSetting("apiKey") %}
+        <h4>{% trans "Settings" %}</h4>
+        <p>{% trans "Your API key allows for secure communication between the CMS and the Xibo audience service. It is used
```
dasgarner, line 32: We shouldn't have carriage returns in translated strings i'm afraid

ifarzana:
@dasgarner , the PR is ready to be merged

dasgarner (approved):

PR #1524, opened 2022-12-22
Receive Stat takes only ad campaign stats

ifarzana:


lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -192,6 +193,9 @@ public function onRegularMaintenance(MaintenanceRegularEvent $event)
                     $entry['campaignEnd'] = $campaignCache[$parentCampaignId]['end'];
                 } else {
                     $parentCampaign = $this->campaignFactory->getById($parentCampaignId);
+                    if ($parentCampaign->type !== 'ad') {
```
dasgarner, line 196: I think we should cache the result of this lookup, otherwise a query returning 10000 stats all for layout list campaigns will make the lookup 10000 times?

lib/Storage/MongoDbTimeSeriesStore.php
```diff
@@ -569,6 +570,11 @@ public function getStats($filterBy = [])
             $match['$match']['parentCampaignId'] = $parentCampaignId;
         }
 
+        // Has Parent Campaign Filter
+        if ($hasParentCampaign) {
```
dasgarner, line 574: I don't think this filter is name correctly, because if you provided false to this you might expect to get back records without a parentCampaign only. - would a better name be `mustHaveParentCampaign` ?

lib/Storage/MySqlTimeSeriesStore.php
```diff
@@ -379,6 +380,11 @@ public function getStats($filterBy = [])
             $params['parentCampaignId'] = $parentCampaignId;
         }
 
+        // Has Parent Campaign Filter
+        if ($hasParentCampaign) {
```
dasgarner, line 384: Same comment is in other store regarding the naming of the parameter.

dasgarner (changes requested):
Makes sense to me, although I have a few comments.

dasgarner:
Let me know when this is ready to be reviewed again - we should merge this ahead of #1525

dasgarner (approved):

PR #1530, opened 2023-01-02
Add translation strings for the dropdown list items

dagonix:


dasgarner (changes requested):
I'm not sure if that works because you're using a `{%` inside an existing `{%` and they shouldn't be nested.

What you can do instead is use a `|` syntax, so `"Off"|trans`

I hope that makes sense.

dasgarner:
I'm closing this out as stale; happy for you to open an issue to track what you were trying to solve.

PR #1536, opened 2023-01-06
Bugfix/3.3.2 misc fixes

PeterMis:
- Interactive : Fix download for deeper nested Layouts.
- Sub Playlist : save (and get) spots and spotsLength as integers.
- DataSet : Fix issue when adding a row with empty value for Date column.
- User Preferences : Fix saving preference with value 0 when it does not exist in the database.

### DIFF
--- a/lib/Widget/SubPlaylist.php
+++ b/lib/Widget/SubPlaylist.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2022 Xibo Signage Ltd
+ * Copyright (C) 2023 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - http://www.xibo.org.uk
  *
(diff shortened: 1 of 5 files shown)


lib/Widget/SubPlaylist.php
```diff
@@ -109,8 +109,8 @@ public function getAssignedPlaylists(): array
             $item->rowNo = intval($playlist['rowNo']);
             $item->playlistId = $playlist['playlistId'];
             $item->spotFill = $playlist['spotFill'] ?? null;
-            $item->spotLength = $playlist['spotLength'] ?? null;
-            $item->spots = $playlist['spots'] ?: null;
+            $item->spotLength = intval($playlist['spotLength']) ?? null;
```
dasgarner, line 112: Doesn't this mean that it will never be null?

lib/Widget/SubPlaylist.php
```diff
@@ -109,8 +109,8 @@ public function getAssignedPlaylists(): array
...
+            $item->spots = intVal($playlist['spots']) ?: null;
```
dasgarner, line 113: Capital V? :smile:

dasgarner (changes requested):
Just one query on the subplaylist intval statements.

dasgarner (approved):

PR #1537, opened 2023-01-06
Fix reports

ifarzana:


lib/Report/MobileProofOfPlay.php
```diff
@@ -358,24 +358,31 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+            try {
+                if (!empty($entry['displayId'])) {
+                    if (!array_key_exists($row['displayId'], $displayCache)) {
+                        $display = $this->displayFactory->getById($row['displayId']);
+                        $displayCache[$row['displayId']] = $display->display;
+                    }
                 }
+                $entry['display'] = $displayCache[$row['displayId']] ?? '';
+            } catch (\Exception $e) {
+                $entry['display'] = 'Not found';
```
dasgarner, line 370: I think we can translate this?

lib/Report/MobileProofOfPlay.php
```diff
@@ -358,24 +358,31 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+            try {
+                if (!empty($entry['layoutId'])) {
+                    if (!array_key_exists($row['layoutId'], $layoutCache)) {
+                        $layout = $this->layoutFactory->getById($row['layoutId']);
+                        $layoutCache[$row['layoutId']] = $layout->layout;
+                    }
                 }
+                $entry['layout'] = $layoutCache[$row['layoutId']] ?? '';
+            } catch (\Exception $e) {
+                $entry['layout'] = 'Not found';
```
dasgarner, line 384: Translations here too.

dasgarner (changes requested):
Just a translation issue to solve.

dasgarner (approved):

PR #1545, opened 2023-01-11
Env: pull 3.3.2 back into v4 work in progress.

dasgarner:
There were several conflicts in JS which were rather hard to resolve, one of which is highlighed with a TODO to make sure that the right choice was made.


maurofmferrao (approved):
Looks good to me, the merge comments can be removed but we can do that in a next release for v4

PR #1551, opened 2023-01-16
Widgets : Send Display tags and values in the xml on Register.

PeterMis:
- On RegisterDisplay function, add `<tags>` xml element, populated with `<tag>` elements containing `<tagName>` element and if no not null also `<tagValue>` element per each Tag assigned to the Display.

Fixes xibosignage/xibo#2773


dasgarner (approved):
Looks good!

PR #1554, opened 2023-01-17
Widget: Dataset Ticker front-end

maurofmferrao:
First version of Dataset Ticker, without the order and filter controls


lib/Listener/DataSetDataProviderListener.php
```diff
@@ -169,6 +169,7 @@ private function getData(DataSet $dataSet, DataProviderInterface $dataProvider):
             // Start should be the lower limit
             // Size should be the distance between upper and lower
             $filter['start'] = $lowerLimit;
+            $filter['end'] = $upperLimit;
```
dasgarner, line 172: I am not clear on what this line here is for.

dasgarner (commented):
Everything makes sense except the one line shown.

lib/Listener/DataSetDataProviderListener.php
maurofmferrao: I think this was just to try to fix the limits not working and I left it there, I'm going to remove it. Good catch! :)

dasgarner (approved):

PR #1555, opened 2023-01-18
Audience Connector: minor refactor to try and solve stalling task

dasgarner:
xibosignageltd/xibo-private#97


ifarzana (approved):

dasgarner:
We think this is done and ready to merge, but we will hold off while we work out the maximum number of records we can process in the allocated time.

PR #1556, opened 2023-01-19
Modules/Templates: support for using assets included with the widget/module xml. Support data provider cache invalidation by date.

dasgarner:
xibosignageltd/xibo-private#99

This PR will need two changes to in progress dev databases:

```sql
alter table display_media
         add modifiedAt timestamp default null null on update CURRENT_TIMESTAMP

alter table requiredfile
         modify realId varchar(254) null

```


PeterMis (approved):
I think this looks good

PR #1557, opened 2023-01-19
Widgets : Notification Data Provider

PeterMis:
- Notification Provider
- NotificationData Event and listener
- DataProvider added isPreview and getWidgetId

https://github.com/xibosignageltd/xibo-private/issues/98

### DIFF
--- a/lib/Widget/Provider/DataProvider.php
+++ b/lib/Widget/Provider/DataProvider.php
@@ -117,16 +117,6 @@ class DataProvider implements DataProviderInterface
     }
 
     /**
-     * @param $isPreview
-     * @return DataProviderInterface
-     */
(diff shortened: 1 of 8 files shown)


lib/Widget/Provider/DataProvider.php
```diff
@@ -153,6 +166,22 @@ public function getDisplayLongitude(): float
         return $this->longitude;
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function isPreview(): bool
+    {
+        return $this->isPreview;
```
dasgarner, line 174: Could this be as simple as `return !empty($this->displayId)` so that we didn't have to have `setPreview`, etc ?

dasgarner (approved):

PR #1558, opened 2023-01-19
Widgets : AlphaVantage Connector

PeterMis:
- Data for Stocks Widget, cache per stock symbol inside Connector
- Data for Currencies Widget, cache per from/to currencies inside Connector

https://github.com/xibosignageltd/xibo-private/issues/104
https://github.com/xibosignageltd/xibo-private/issues/105


lib/Widget/Provider/DataProviderInterface.php
```diff
@@ -76,6 +76,12 @@ public function getDisplayLongitude(): float;
      */
     public function getGuzzleClient(array $requestOptions = []): \GuzzleHttp\Client;
 
+    /**
+     * Get Stash pool interface
+     *  User it if any data needs to be cached inside the Data Provider
+     */
+    public function getPool() : \Stash\Interfaces\PoolInterface;
```
dasgarner, line 83: This should be inside whatever is using the data provider - not the actual data provider.

lib/Connector/AlphaVantageConnector.php
```diff
@@ -0,0 +1,532 @@
...
+use Xibo\Widget\Provider\DataProviderInterface;
+
+/**
+ * A connector to get data from the AlphaVantage API for use by the Currencies and Stocks Widgets
+ */
+class AlphaVantageConnector implements ConnectorInterface
+{
+    use ConnectorTrait;
+
+    public function registerWithDispatcher(EventDispatcherInterface $dispatcher): ConnectorInterface
```
dasgarner, line 44: You can use `public function setFactories(ContainerInterface $container): ConnectorInterface` here to get the Cache Pool

dasgarner (changes requested):
A slight misunderstand on where the cache pool should be injected!

dasgarner (approved):

PR #1559, opened 2023-01-20
Audience connector: Running indefinitely for multiple blocks of stat

ifarzana:
- Connector should only send a block with a size of 5000 or batch size defined in connector provider settings
- Connector sets the watermark, sends a request to audience reporting service to update the watermark
- Watermark is the last statId of the block, that has been returned by the getStats()

https://github.com/xibosignageltd/xibo-private/issues/107

### DIFF
--- a/views/xibo-audience-connector-form-settings.twig
+++ b/views/xibo-audience-connector-form-settings.twig
@@ -34,6 +34,10 @@
         {% set title %}{% trans "API Key" %}{% endset %}
         {% set helpText %}{% trans "Enter your API Key from Xibo." %}{% endset %}
         {{ forms.input("apiKey", title, interface.getSetting("apiKey"), helpText) }}
+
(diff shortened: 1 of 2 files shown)


views/xibo-audience-connector-form-settings.twig
```diff
@@ -34,6 +34,10 @@
         {% set title %}{% trans "API Key" %}{% endset %}
         {% set helpText %}{% trans "Enter your API Key from Xibo." %}{% endset %}
         {{ forms.input("apiKey", title, interface.getSetting("apiKey"), helpText) }}
+
+        {% set title %}{% trans "Batch Size" %}{% endset %}
```
dasgarner, line 38: This is inside the check for apiKey being a provider setting? I think it needs its own condition.

dasgarner (approved):

PR #1563, opened 2023-01-25
Connector Report/Receive Stat: Send the start/end date in the local time zone

ifarzana:
Audience Connector/Send Stat:

- Start/End dates are saved in UTC format in MongoDB
- Convert those dates to the display time zone if a timezone is set on the display, otherwise use the CMS time zone

Proof of play CSV export:
- allow users to select a date format (UTC/CMS local time) when exporting stats

 https://github.com/xibosignageltd/xibo-private/issues/114

### DIFF
--- a/lib/Connector/XiboAudienceReportingConnector.php
+++ b/lib/Connector/XiboAudienceReportingConnector.php
@@ -287,7 +287,7 @@ class XiboAudienceReportingConnector implements ConnectorInterface
                             $display = $this->displayFactory->getById($displayId);
                             $displayCache[$displayId]['costPerPlay'] = $display->costPerPlay;
                             $displayCache[$displayId]['impressionsPerPlay'] = $display->impressionsPerPlay;
-                            $displayCache[$displayId]['timeZone'] = $display->timeZone;
+                            $displayCache[$displayId]['timeZone'] = empty($display->timeZone) ? $defaultTimezone : $display->timeZone;
                         } catch (NotFoundException $notFoundException) {
                             $this->getLogger()->error('onRegularMaintenance: display not found with ID: '
                                 . $displayId);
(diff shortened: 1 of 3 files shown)


lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -280,6 +287,7 @@ public function onRegularMaintenance(MaintenanceRegularEvent $event)
                             $display = $this->displayFactory->getById($displayId);
                             $displayCache[$displayId]['costPerPlay'] = $display->costPerPlay;
                             $displayCache[$displayId]['impressionsPerPlay'] = $display->impressionsPerPlay;
+                            $displayCache[$displayId]['timeZone'] = $display->timeZone;
```
dasgarner, line 290:
```suggestion
                            $displayCache[$displayId]['timeZone'] = empty($display->timeZone) ? $defaultTimezone : $display->timeZone;
```

lib/Connector/XiboAudienceReportingConnector.php
```diff
@@ -293,13 +301,19 @@ public function onRegularMaintenance(MaintenanceRegularEvent $event)
...
+                                : Carbon::instance($row['start']->toDateTime())->timezone($defaultTimezone);
+                            $end = ($displayCache[$displayId]['timeZone'] != null)
+                                ? Carbon::instance($row['end']->toDateTime())->timezone($displayCache[$displayId]['timeZone'])
+                                : Carbon::instance($row['end']->toDateTime())->timezone($defaultTimezone);
                         } else {
-                            $start = Carbon::createFromTimestamp($row['start'])
-                                ->format(DateFormatHelper::getSystemFormat());
-                            $end = Carbon::createFromTimestamp($row['end'])
-                                ->format(DateFormatHelper::getSystemFormat());
+                            $start = Carbon::createFromTimestamp($row['start']);
```
dasgarner, line 315: These dates come from MySQL, which will get created in CMS time zone ✅, however, we still want to convert them into their display time zone, correct?

dasgarner (changes requested):
I think there are some minor tweaks to address here - also can you include the other items from xibosignage/xibo#3003 ?

lib/Controller/Stats.php
```diff
@@ -710,7 +710,7 @@ public function export(Request $request, Response $response)
         $toDt = $sanitizedParams->getDate('toDt');
         $displayId = $sanitizedParams->getInt('displayId');
         $tempFileName = $this->getConfig()->getSetting('LIBRARY_LOCATION') . 'temp/stats_' . Random::generateString();
-        $dateFormat = $sanitizedParams->getString('dateFormat'); // UTC/CMS local time
+        $dateFormat = $sanitizedParams->getCheckbox('dateFormat');
```
dasgarner: Again, I think having this parameter called `dateFormat` is confusing - it isn't a date format. I think I suggested an alternative parameter name in my last review?

dasgarner (changes requested):

lib/Controller/Stats.php
ifarzana: I have done that in the latest commit.

dasgarner (approved):

PR #1569, opened 2023-01-30
Correcting data types of columns in Ad Campaigns/Displays migrations

ifarzana:
https://github.com/xibosignage/xibo/issues/2953


dasgarner:
Have we tested what happens if this migration runs and those columns are already set that way?

ifarzana:
If they are same, the migration still runs successfully.

dasgarner (approved):

PR #1572, opened 2023-02-01
WIP: Convert Currencies and Stocks modules

maurofmferrao:
relates to
xibosignageltd/xibo-private#117
xibosignageltd/xibo-private#118

 - Module Twig translated
 - **Still need to add flags for Currencies**


dasgarner (approved):
Makes sense

PR #1576, opened 2023-02-03
Ad Campaigns : tidy up interrupt schedules after 90 days.

PeterMis:
- Regular Maintenance : new function that looks for 90 days + old interrupt events with set parentCampaignId and deletes them making sure to not notify Displays (notify => false).
- Schedule Entity delete : add array of options with default notify true, do not even attempt to check inScheduleLookAhead if notify is set to false

Fixes xibosignage/xibo#2977

### DIFF
--- a/lib/XTR/MaintenanceRegularTask.php
+++ b/lib/XTR/MaintenanceRegularTask.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (c) 2022 Xibo Signage Ltd
+ * Copyright (C) 2023 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - http://www.xibo.org.uk
  *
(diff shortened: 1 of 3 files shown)


lib/XTR/MaintenanceRegularTask.php
```diff
@@ -523,4 +532,27 @@ private function assessDynamicDisplayGroups()
...
-}
\ No newline at end of file
+
+    private function tidyAdCampaignSchedules()
+    {
+        $this->runMessage .= '## ' . __('Tidy Ad Campaign Schedules') . PHP_EOL;
+        Profiler::start('RegularMaintenance::tidyAdCampaignSchedules', $this->log);
+        $count = 0;
+
+        foreach ($this->scheduleFactory->query(null, [
```
dasgarner, line 542: Would it be better to make it possible to query the `parentCampaignId` in schedule factory

dasgarner (approved):

PR #1577, opened 2023-02-03
Display Map: add display status as pie chart 

ifarzana:
- We used html and css to create pie charts for each clustered displays, Chart.js or D3 are not required.


dasgarner (approved):
Looks great to me

PR #1581, opened 2023-02-06
Bug Fix 3.3.3: Fix for MySQL Audience Connector send stats

ifarzana:
To ensure that the results of the "getStats" query are stored in memory, set the MYSQL_ATTR_USE_BUFFERED_QUERY attribute to true when sending stats.

Fixes: https://github.com/xibosignage/xibo/issues/3009


dasgarner (approved):
Looks good.

PR #1594, opened 2023-02-09
Widgets : Open Weather Map Connector

PeterMis:
- New Connector, get data from OWM and add to the provider on dataRequest
- Migration that adds the new connector
- Settings twig file for the new connector

### DIFF
--- a/views/openweathermap-form-settings.twig
+++ b/views/openweathermap-form-settings.twig
@@ -1,0 +1,67 @@
+{#
+/**
+ * Copyright (C) 2023 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - http://www.xibo.org.uk
+ *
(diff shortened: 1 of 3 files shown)


views/openweathermap-form-settings.twig
```diff
@@ -0,0 +1,67 @@
...
+ */
+#}
+
+{% extends "connector-form-edit.twig" %}
+{% import "forms.twig" as forms %}
+
+{% block connectorFormFields %}
+    <h3>Open Weather Map</h3>
+
+    {% set title %}
```
dasgarner, line 30: Should this message only be shown if the provider hasn't provided settings?

dasgarner (commented):
I think just one point to raise here.

dasgarner (approved):

PR #1599, opened 2023-02-15
Convert Module: Webpage

ifarzana:
Relates to https://github.com/xibosignageltd/xibo-private/issues/82


maurofmferrao:
@ifarzana , I had to make some changes to `xibo-webpage-render.js` to make the widget to work, and then I grabbed the code from your PR and update the `onRender` method to translate the values from the form to the renderer. I think we can close this PR, and if you don't mind reviewing mine [here](https://github.com/xibosignage/xibo-cms/pull/1602) so we can test it :)

PR #1607, opened 2023-02-20
Group for non datatype modules

ifarzana:
Fixes https://github.com/xibosignageltd/xibo-private/issues/135


maurofmferrao (approved):
Looks good to me!

PR #1612, opened 2023-02-22
Modules: Input option to set values on other inputs

maurofmferrao:
resolves xibosignageltd/xibo-private#101

 - Visibility now added via JS like dependsOn


ifarzana (approved):
This looks good

maurofmferrao:
@ifarzana I've added some more commits, one to fix setting the value for a colour input. And the 2 others to implement colour templates for Metro template in social

ifarzana (approved):

PR #1613, opened 2023-02-22
DataSet Widgets : Fix direct library references.

PeterMis:
Fixes xibosignage/xibo#3024


dasgarner (approved):
Simple solution, makes sense. We won't need this for v4 either (which is fine as this is in the provider now anyway)

PR #1614, opened 2023-02-22
Display Group : Fix Display Group delete when linked to ad Campaign

PeterMis:
Fixes xibosignage/xibo#3022


dasgarner (approved):
Makes sense!

PR #1616, opened 2023-02-24
Convert Module: Dashboards

PeterMis:
xibosignageltd/xibo-private#136

### DIFF
--- a/lib/Connector/XiboDashboardConnector.php
+++ b/lib/Connector/XiboDashboardConnector.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (c) 2022 Xibo Signage Ltd
+ * Copyright (C) 2023 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - http://www.xibo.org.uk
  *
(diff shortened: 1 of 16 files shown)


lib/Connector/XiboDashboardConnector.php
```diff
@@ -483,4 +487,81 @@ public function onWidgetEditOption(WidgetEditOptionRequestEvent $event)
...
+            $tokenEvent->setTtl(3600 * 24 * 2);
+            $dispatcher->dispatch($tokenEvent, XmdsConnectorTokenEvent::$NAME);
+            $token = $tokenEvent->getToken();
+
+            if (empty($token)) {
+                throw new ConfigurationException(__('No token returned'));
+            }
+
+            if ($event->getDataProvider()->isPreview()) {
+                $url = $event->getDataProvider()->getRouteParser()->urlFor('layout.preview.connector', [], ['token' => $token]);
```
dasgarner, line 506: To save having a route parser in here we could output a placeholder `[[connector=$token]]` and parse that out in `WidgetHtmlRenderer::decorateForPreview` ?

I suggest we have a helper for it in the connector trait to output the placeholder in case we change that later? `getLayoutPreviewUrl($token)` ?

modules/dashboard.xml
```diff
@@ -0,0 +1,124 @@
...
+    <type>dashboard</type>
+    <dataType></dataType>
+    <schemaVersion>1</schemaVersion>
+    <assignable>1</assignable>
+    <regionSpecific>1</regionSpecific>
+    <renderAs>html</renderAs>
+    <defaultDuration>60</defaultDuration>
+    <settings></settings>
+    <properties>
+        <property id="type" type="dashboardSelector">
```
dasgarner, line 37: We've made this type specific to dashboards, but it is actually more generically a `connectorProperties` control? I.e. any connector can handle this request and return additional properties?

lib/Controller/Widget.php
```diff
@@ -987,7 +989,7 @@ public function getData(Request $request, Response $response, $regionId, $id)
         $module = $this->moduleFactory->getByType($widget->type);
 
         // This is always a preview
-        if (!$module->isDataProviderExpected() && !$module->isWidgetProviderAvailable()) {
+        if (!$module->isDataProviderExpected() && !$module->isWidgetProviderAvailable() && $module->type !== 'dashboard') {
```
dasgarner, line 992: Can't we use a Widget Provider instead of adding this condition?

dasgarner (commented):
Some ideas before we merge.

dasgarner (approved):
Looks fab

PR #1622, opened 2023-02-28
Android Intent Command Control: Added extra shows plus instead of minus

ifarzana:
Fixes: https://github.com/xibosignage/xibo/issues/3023

### DIFF
--- a/views/authed.twig
+++ b/views/authed.twig
@@ -306,7 +306,7 @@
                 <option value="intArray" {{#eq type "intArray"}}selected="selected"{{/eq}}>intArray</option
             </select>
             <input class="form-control form-control-inline validated-element extra-value" placeholder="{% endverbatim %}{% trans "Value" %}{% verbatim %}" value="{{ value }}">
-            <button type="button" class="btn btn-sm btn-warning pull-right intent-remove-extra"><i class="fa fa-minus fa-plus"></i></button>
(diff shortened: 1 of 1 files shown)


views/authed.twig
```diff
@@ -306,7 +306,7 @@
                 <option value="intArray" {{#eq type "intArray"}}selected="selected"{{/eq}}>intArray</option
             </select>
             <input class="form-control form-control-inline validated-element extra-value" placeholder="{% endverbatim %}{% trans "Value" %}{% verbatim %}" value="{{ value }}">
-            <button type="button" class="btn btn-sm btn-warning pull-right intent-remove-extra"><i class="fa fa-minus fa-plus"></i></button>
+            <button type="button" class="btn btn-sm btn-warning pull-right intent-remove-extra"><i class="fa fa-minus"></i></button>
```
maurofmferrao, line 309: This one works, but I think you've missed the same fix in line 400 for the `command-input-http-key-value-template` template

maurofmferrao (changes requested):
We just need the same fix for `command-input-http-key-value-template` :)

views/authed.twig
ifarzana: Thanks for pointing that

maurofmferrao (approved):
Looks good to me!

PR #1626, opened 2023-03-01
Ticker widget: add option to set the user agent

ifarzana:
- In Layout Editor, Ticker widget will now have the option to set userAgent
Relates to: https://github.com/xibosignage/xibo/issues/2999


dasgarner (approved):
This looks good - please can you target the issue to the latest v3 patch and then make the same changes on v4?

PR #1631, opened 2023-03-02
Ticker widget: add option to set the user agent (4.0.0)

ifarzana:
In Layout Editor, RSS Ticker widget will now have the option to set userAgent
Relates to: https://github.com/xibosignage/xibo/issues/2999

### DIFF
--- a/lib/Widget/RssProvider.php
+++ b/lib/Widget/RssProvider.php
@@ -60,14 +60,20 @@ class RssProvider implements WidgetProviderInterface
             ->format('U');
 
         try {
+            $httpOptions = [
+                'headers' => [
+                    'Accept' => 'application/rss+xml, application/rdf+xml;q=0.8, application/atom+xml;q=0.6,'
+                        . 'application/xml;q=0.4, text/xml;q=0.4, text/html;q=0.2, text/*;q=0.1'
+                ],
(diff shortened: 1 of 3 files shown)


lib/Widget/RssProvider.php
```diff
@@ -60,14 +60,20 @@ public function fetchData(DataProviderInterface $dataProvider): WidgetProviderIn
...
+            $httpOptions = [
+                'headers' => [
+                    'Accept' => 'application/rss+xml, application/rdf+xml;q=0.8, application/atom+xml;q=0.6,'
+                        . 'application/xml;q=0.4, text/xml;q=0.4, text/html;q=0.2, text/*;q=0.1'
+                ],
+                'timeout' => 20, // wait no more than 20 seconds
+            ];
+
+            if (!empty($dataProvider->getProperty('userAgent'))) {
+                $httpOptions['User-Agent'] = $dataProvider->getProperty('userAgent');
```
PeterMis, line 72: I might be wrong, please do double check with Guzzle documentation, but I *think* that the User-Agent option should be in headers
ie 
```
$httpOptions['headers']['User-Agent']
```

ifarzana: You are right, it should be part of headers, Thanks for noticing that.

PeterMis (approved):

PR #1645, opened 2023-03-08
Modules: snippets implementation for the BE

dasgarner:
* a second attempt at a more structured approach. xibosignageltd/xibo-private#154


dasgarner:
Please ignore the copyright changes - i've updated the KB with a copyright profile which should prevent this in future

PeterMis (approved):
I think this approach is indeed better 👍

PR #1647, opened 2023-03-08
Modules: Tidy old files and check form input retrocompatibility

maurofmferrao:
resolves xibosignageltd/xibo-private#153


dasgarner (approved):
Looks scary but correct :smile:

PR #1652, opened 2023-03-09
Assets: prevent certain assets from being sent to the player.

dasgarner:
xibosignageltd/xibo-private#152


maurofmferrao (approved):
Looks good to me!

PR #1654, opened 2023-03-10
Reports: saved reports should not be stored in the media

ifarzana:
Fixes: https://github.com/xibosignage/xibo/issues/2893

### DIFF
--- a/lib/XTR/ReportScheduleTask.php
+++ b/lib/XTR/ReportScheduleTask.php
@@ -1,6 +1,6 @@
 <?php
-/**
- * Copyright (C) 2019 Xibo Signage Ltd
+/*
+ * Copyright (C) 2023 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - http://www.xibo.org.uk
  *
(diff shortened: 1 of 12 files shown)


lib/XTR/ReportScheduleTask.php
```diff
@@ -161,8 +165,10 @@ private function runReportSchedule()
                     fwrite($out, json_encode($result));
                     fclose($out);
 
+                    $runDateTimestamp = Carbon::now()->format('U');
+
                     // Create a ZIP file and add our temporary file
-                    $zipName = $this->config->getSetting('LIBRARY_LOCATION') . 'temp/reportschedule.json.zip';
+                    $zipName = $this->config->getSetting('LIBRARY_LOCATION') . 'savedreport/reportschedule_'.$runDateTimestamp.'.json.zip' ;// random numb
```
PeterMis, line 171: We likely don't need the random numb comment here 🤔

lib/XTR/ReportScheduleTask.php
```diff
@@ -214,7 +221,7 @@ private function runReportSchedule()
      * @throws \Twig\Error\SyntaxError
      * @throws \Xibo\Support\Exception\GeneralException
      */
-    private function createPdfAndNotification($reportSchedule, $savedReport, $media)
+    private function createPdfAndNotification($reportSchedule, $savedReport, $runDateTimestamp)
```
PeterMis, line 224: I wonder if we perhaps should use `$savedReport` id here instead of the timestamp? Perhaps with some prefix if needed

PeterMis (commented):
Overall looks good to me, just some small questions below

lib/XTR/ReportScheduleTask.php
ifarzana: I agree.

PeterMis (approved):

PR #1657, opened 2023-03-14
Modules: Show a thumbnail inside the module card on the toolbar

ifarzana:
- Fixes https://github.com/xibosignageltd/xibo-private/issues/173


maurofmferrao (approved):
Looks good to me!

PR #1667, opened 2023-03-16
Modules: Group properties

ifarzana:
- Fixes https://github.com/xibosignageltd/xibo-private/issues/175

### DIFF
--- a/lib/Entity/Module.php
+++ b/lib/Entity/Module.php
@@ -2,7 +2,7 @@
 /*
  * Copyright (C) 2023 Xibo Signage Ltd
  *
- * Xibo - Digital Signage - http://www.xibo.org.uk
+ * Xibo - Digital Signage - https://xibosignage.com
  *
  * This file is part of Xibo.
  *
--- a/lib/Entity/Module.php
+++ b/lib/Entity/Module.php
@@ -227,7 +227,7 @@ class Module implements \JsonSerializable
      *     type="array",
      *     @SWG\Items(type="string")
      * )
-     * @var \Xibo\Widget\Definition\GroupProperty[]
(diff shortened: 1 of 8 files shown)


lib/Widget/Definition/GroupProperty.php
dasgarner, line 1: Would it make more sense for this to be called `PropertyGroup` to match the XML?

lib/Entity/Module.php
```diff
@@ -221,6 +221,16 @@ class Module implements \JsonSerializable
...
 
+    /**
+     * @SWG\Property(
+     *     description="An array of additional module specific group properties",
+     *     type="array",
+     *     @SWG\Items(type="string")
+     * )
+     * @var \Xibo\Widget\Definition\GroupProperty[]
+     */
+    public $propertyGroups = [];
```
dasgarner, line 232: @maurofmferrao are we insisting that the property groups are defined at the module level? Or will we have those also on the Module Template and combine that on the front end when rendering the form?

dasgarner (changes requested):

lib/Entity/Module.php
maurofmferrao: Hmm yeah, I think it would be nice to have them all in the templates as well 🤔

dasgarner: I think it comes down to the question of whether as a template developer, would you want to add your own tabs, or be limited to the ones provided by the module developer?

maurofmferrao: That a good question. I see the possibility of a template having too many properties and have to group them. Something like Metro Static template can use different groups than other "Twitter" widgets, So maybe we should also had that property groups for the template as well

dasgarner (approved):

PR #1669, opened 2023-03-16
Modules/templates default size front-end

maurofmferrao:
relates to xibosignageltd/xibo-private#174

### DIFF
--- a/modules/templates/article-static.xml
+++ b/modules/templates/article-static.xml
@@ -1,7 +1,7 @@
 <!--
   ~ Copyright (C) 2023 Xibo Signage Ltd
   ~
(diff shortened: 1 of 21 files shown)


modules/templates/article-static.xml
```diff
@@ -1,7 +1,7 @@
 <!--
   ~ Copyright (C) 2023 Xibo Signage Ltd
   ~
-  ~ Xibo - Digital Signage - https://xibosignage.com
+  ~ Xibo - Digital Signage - http://www.xibo.org.uk
```
dasgarner, line 4: Please can you update your copyright profile with the one from Rosetta?

dasgarner (approved):

PR #1673, opened 2023-03-20
3.3.4 - hide empty row menu and fix group admin userGroup permissions

PeterMis:
- User Group : Group admin permissions fix, specific handling for User Group entity, as we can't rely on ownerId (it's not a thing for user groups ) - https://github.com/xibosignage/xibo/issues/3019
- If we do not have any buttons to show in the row menu dropdown, then do not show the dropdown at all - easiest example would be the system tags on Tags page - https://github.com/xibosignage/xibo/issues/3006

### DIFF
--- a/lib/Controller/UserGroup.php
+++ b/lib/Controller/UserGroup.php
@@ -1,8 +1,8 @@
 <?php
 /*
- * Copyright (c) 2022 Xibo Signage Ltd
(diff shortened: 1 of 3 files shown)


dasgarner (approved):

ifarzana:
Tested :heavy_check_mark:

PR #1677, opened 2023-03-22
Modules: add database and XML to support upgrades

ifarzana:
https://github.com/xibosignageltd/xibo-private/issues/176


lib/Factory/ModuleTemplateFactory.php
```diff
@@ -168,6 +168,7 @@ private function createFromXml(\DOMElement $xml): ModuleTemplate
         $template->dataType = $this->getFirstValueOrDefaultFromXmlNode($xml, 'dataType');
         $template->title = $this->getFirstValueOrDefaultFromXmlNode($xml, 'title');
         $template->thumbnail = $this->getFirstValueOrDefaultFromXmlNode($xml, 'thumbnail');
+        $template->isVisible = $this->getFirstValueOrDefaultFromXmlNode($xml, 'isVisible');
```
dasgarner, line 171: This will return a string, so we need to test it with `===` to convert to a boolean.

lib/Factory/WidgetFactory.php
```diff
@@ -266,6 +266,7 @@ public function query($sortOrder = null, $filterBy = [])
               widget.type,
               widget.duration,
               widget.displayOrder,
+              widget.schemaVersion,
```
dasgarner, line 269: I think we're going to need `schemaVersion` in the `intProperties` when we hydrate?

lib/Widget/Provider/WidgetCompatibilityInterface.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+namespace Xibo\Widget\Provider;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Xibo\Entity\Widget;
+
+/**
+ * The Widget Compatibility Interface
```
dasgarner, line 30: Can we expand this comment with some detail around the purpose of the interface? A summary of the Rosetta topic would be good.

lib/Widget/Provider/WidgetCompatibilityInterface.php
```diff
@@ -0,0 +1,68 @@
...
+interface WidgetCompatibilityInterface
+{
+    public function getLog(): LoggerInterface;
+    public function setLog(LoggerInterface $logger): WidgetCompatibilityInterface;
+
+    /**
+     * Get the event dispatcher
+     * @return \Symfony\Component\EventDispatcher\EventDispatcherInterface
+     */
+    public function getDispatcher(): EventDispatcherInterface;
```
dasgarner, line 41: I don't think we need a dispatcher in here.

lib/Widget/Provider/WidgetCompatibilityInterface.php
```diff
@@ -0,0 +1,68 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Xibo\Entity\Widget;
+
+/**
+ * The Widget Compatibility Interface
+ */
+interface WidgetCompatibilityInterface
+{
+    public function getLog(): LoggerInterface;
```
dasgarner, line 34: Can you make a "WidgetCompatibilityTrait" which implements getLog/setLog?

dasgarner (changes requested):
You've got the idea here - i've left a few comments

lib/Factory/ModuleTemplateFactory.php
```diff
@@ -168,7 +168,7 @@ private function createFromXml(\DOMElement $xml): ModuleTemplate
         $template->dataType = $this->getFirstValueOrDefaultFromXmlNode($xml, 'dataType');
         $template->title = $this->getFirstValueOrDefaultFromXmlNode($xml, 'title');
         $template->thumbnail = $this->getFirstValueOrDefaultFromXmlNode($xml, 'thumbnail');
-        $template->isVisible = $this->getFirstValueOrDefaultFromXmlNode($xml, 'isVisible');
+        $template->isVisible = $this->getFirstValueOrDefaultFromXmlNode($xml, 'isVisible') === 'true';
```
dasgarner, line 171: This means that if it is empty (not provided in the XML) it will be visible false - I think logically we'd want this the other way around, i.e. if it isn't there then it should be visible=true?

lib/Widget/Provider/WidgetCompatibilityInterface.php
```diff
@@ -23,29 +23,23 @@
...
 /**
- * The Widget Compatibility Interface
+ * Widget Compatibility Interface should be implemented by custom widget upgrade classes.
+ * Takes necessary actions to make the existing widgets from v3 compatible with v4.
+ *
+ * The schema from and the schema to (currently set to 1 and 2, respectively).
+ * It also provides a method to save a template to the library in a sub-folder named templates/. This method
+ * is called whenever a widget is loaded with a different schema version.
+ *
+ * @package MyNamespace
```
dasgarner, line 36: Its good apart from the package name here :smile:

dasgarner (changes requested):
Nearly!

dasgarner (approved):

PR #1680, opened 2023-03-22
Layout Editor controls: Improve setDefault

ifarzana:
https://github.com/xibosignageltd/xibo-private/issues/188


dasgarner (changes requested):
You will also need to modify the Module XML where we're already using setDefault so that it uses the new options instead.

ifarzana:
I have removed the old way of setDefault.

`You will also need to modify the Module XML where we're already using setDefault so that it uses the new options instead.`

I don't understand what do you mean here.

dasgarner:
> > You will also need to modify the Module XML where we're already using setDefault so that it uses the new options instead.
> I don't understand what do you mean here.

The social media static templates use <setDefault> in the XML and so we need to swap that over to the new XML properties instead.

dasgarner (approved):

PR #1682, opened 2023-03-22
Env: swap to github container registry 

dasgarner:
relates to xibosignageltd/xibo-private#180


PeterMis (approved):
Look good, I *think* :)

PR #1687, opened 2023-03-23
Modules: refactor WidgetEditOptionRequestEvent to make it generic 

dasgarner:
I've made a change here so that the front end passes in a propertyId with the request and as the initialKey. The back end then passes that directly into the event so that it can be handled entirely by the connector (or whatever handles the event).


PeterMis (approved):
Makes sense 👍

PR #1688, opened 2023-03-23
Modules: certain modules should be excluded from the Layout Editor

ifarzana:
- Relates to https://github.com/xibosignageltd/xibo-private/issues/165


dasgarner (approved):
This looks great - please can we pause until we have implemented the elements based text module (otherwise we won't be able to add text to a layout while that work is in progress).

dasgarner:
This has been included in https://github.com/xibosignage/xibo-cms/pull/1711

PR #1692, opened 2023-03-24
Regions: adding a second media item to a zone should convert to playlist

ifarzana:
- Relates to https://github.com/xibosignageltd/xibo-private/issues/192
- Impemented on addWidget() and libraryAssign()

### DIFF
--- a/lib/Controller/Widget.php
+++ b/lib/Controller/Widget.php
@@ -30,7 +30,6 @@ use Xibo\Event\MediaDeleteEvent;
 use Xibo\Event\WidgetAddEvent;
 use Xibo\Event\WidgetDataRequestEvent;
 use Xibo\Event\WidgetEditOptionRequestEvent;
-use Xibo\Factory\LayoutFactory;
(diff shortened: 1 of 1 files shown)


lib/Controller/Widget.php
```diff
@@ -241,6 +248,27 @@ public function addWidget(Request $request, Response $response, $type, $id)
         $displayOrder = $params->getInt('displayOrder');
         $playlist->assignWidget($widget, $displayOrder);
 
+        // Get the widget region and Layout
+        $widgetRegion = $this->regionFactory->getByPlaylistId($widget->playlistId)[0];
```
dasgarner, line 252: Same as the previous comments - this can be a lot simpler.

```php
$playlist->load();
if ($playlist->isRegionPlaylist() && count($playlist->widgets) >= 1) {
    // Convert this region to a `playlist` (if it is a zone)
    $widgetRegion = $this->regionFactory->getById($playlist->regionId);
    (...)
}

```

dasgarner (changes requested):

PR #1694, opened 2023-03-27
Displays: Commercial license column

ifarzana:
- Displays: Clarify that the commercial license column is only up-to-date as of the last command run. Fixes https://github.com/xibosignage/xibo/issues/3031

### DIFF
--- a/lib/Controller/Display.php
+++ b/lib/Controller/Display.php
@@ -716,8 +716,10 @@ class Display extends Base
                     $display->commercialLicenceDescription = __('Display is not licensed');
             }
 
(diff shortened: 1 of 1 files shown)


lib/Controller/Display.php
```diff
@@ -716,6 +716,8 @@ function grid(Request $request, Response $response)
                     $display->commercialLicenceDescription = __('Display is not licensed');
             }
 
+            $display->commercialLicenceDescription .= ' (' . __('The status will be updated with each Commercial Licence check') . ')';
```
dasgarner, line 719: Please can we future proof this by only showing it if the player clientCode < 400 (i.e. version 4 players shouldn't show it).

dasgarner (changes requested):

PR #1697, opened 2023-03-28
Playlist Editor (External): Add playlist info to the modal header

rubenpingol-xibo:


ui/src/templates/playlist-timeline-header-info.hbs
```diff
@@ -0,0 +1,13 @@
+<h4 class="modal-title editor-modal-title">
+    Edit Playlist - {{ playlist.name }} -
```
maurofmferrao, line 2: We need to keep the first part of this info to be translatable. You can create a new variable in `common.twig` for the playlist editor. Maybe `playlistEditorTrans`, and then pass that variable to `timelineHeaderInfoTemplate`

maurofmferrao (approved):
Looks good to me! :)

PR #1699, opened 2023-03-29
Membership Forms : paging support

PeterMis:
- User, UserGroup, Display and DisplayGroup share the same function for open/submit membership forms
- All forms support paging and the same format, relevant dataTables created as needed
- Submit membership form function handles all possible member changes we may have on a form
- Fixed DisplayGroup `assignDisplayGroup()` function : it should dispatch load event, to ensure we have everything we need if membership change causes notify


lib/Factory/DisplayFactory.php
```diff
@@ -605,8 +605,9 @@ public function query($sortOrder = null, $filterBy = [])
 
         $limit = '';
         // Paging
-        if ($filterBy !== null && $parsedBody->getInt('start') !== null && $parsedBody->getInt('length') !== null) {
-            $limit = ' LIMIT ' . $parsedBody->getInt('start', ['default' => 0]) . ', ' . $parsedBody->getInt('length', ['default' => 10]);
+        if ($parsedBody->hasParam('start') && $parsedBody->hasParam('length')) {
+            $limit = ' LIMIT ' . $parsedBody->getInt('start')
```
dasgarner, line 609: This parameter could be present but not set - i.e. `start=&length=10` which I think would fail here because getInt would return `null` ?

ui/src/core/xibo-forms.js
```diff
@@ -500,30 +500,216 @@ function permissionsMultiFormSubmit(id) {
     });
 }
 
+function createDisplayGroupMembersTable(dialog) {
+    var control = $(dialog).find(".controlDiv");
+    let context = {
+        'tableName' : '#displaysGroupsMembersTable',
```
dasgarner, line 506: It would be good to get ESLint running over these - I think `'tableName'` should be without the single quotes as its a property name? `tableName` ?

Equally we should have `'` instead of `"` in `"displayGroupId"` for example.

ui/src/core/xibo-forms.js
```diff
@@ -532,35 +718,177 @@ function membersFormSubmit(id) {
         return;
     }
 
+    // Create a new queue.
+    window.queue = $.jqmq({
```
dasgarner, line 722: Can you explain why we have a queue here?

lib/Factory/DisplayFactory.php
PeterMis: We should always have defaults via gridRenderFilter, but I can double check, this is basically the same change you made to displayGroupFactory in the past - but we can certainly have default value there too, just in case.

ui/src/core/xibo-forms.js
dasgarner: Is it because the displays and display groups are shown in separate tabs on that form, and the save is taking care of both (submitting one after the other).

If that is still the case then I am happy to leave it!

ui/src/core/xibo-forms.js
PeterMis: Enabling this on an older file...is interesting, but I will make needed changes in the code that's in this PR only, so changed file is still readable ;)

PR #1704, opened 2023-03-30
Modules: Review descriptions and default modules

ifarzana:
- Relates to https://github.com/xibosignageltd/xibo-private/issues/198


db/migrations/20220330111440_modules_table_ver_four_migration.php
```diff
@@ -77,6 +77,29 @@ public function change()
 
             // Drop the old table
             $this->dropTable('module_old');
+
+            // Add more v4 modules
+            $this->execute('
+            INSERT INTO `module` (`moduleId`, `enabled`, `previewEnabled`, `defaultDuration`, `settings`) VALUES
```
dasgarner, line 83: Some of these modules are not installed by default in v3, please can you check them again?

dasgarner: Would it be clearer if we simply did not install them?

dasgarner (approved):

PR #1705, opened 2023-03-31
Calendar Module: Add sample data

rubenpingol-xibo:


modules/calendar.xml
```diff
@@ -128,5 +128,35 @@ if(typeof ongoingEvent != 'undefined' && ongoingEvent && properties.currentEvent
 // Start effects for this widget
 $(target).xiboLayoutAnimate(properties);
     ]]></onVisible>
+    <sampleData><![CDATA[
+[{
+    "summary": "Sample 1",
+    "description": "Something here",
```
dasgarner, line 134: Can we have a go at making some better summaries/descriptions for these events? Even if its just "Sample Event 1" and the description explaining what we're doing with the dates.

dasgarner (commented):
Looks good to me - I had one general comment, but will let Mauro review the code more closely. :+1:

maurofmferrao (approved):

maurofmferrao (approved):
Looks good!

modules/calendar.xml
dasgarner: It looks great, please go ahead!! :+1:

PR #1707, opened 2023-03-31
Move compatibility classes 

ifarzana:
- Fixes https://github.com/xibosignageltd/xibo-private/issues/210


dasgarner:
What you've done here is correct, but you will also need to update the reference in the XML file.

dasgarner (approved):

PR #1717, opened 2023-04-04
Misc fixes during initial private beta testing

dasgarner:
* Fix for the default layout importing without a playlist in the centre spot
 * Fix for an error while uploading a video

related to https://github.com/xibosignageltd/xibo-private/issues/197


ifarzana (approved):
Looks fine!

PR #1718, opened 2023-04-04
Env: MySQL 8 support

dasgarner:
* Fix logging/auditing. 

fixes xibosignage/xibo#2974


composer.json
```diff
@@ -28,14 +20,14 @@
   ],
   "repositories": [{
     "type": "vcs",
-    "url": "https://github.com/xibosignage/support"
+    "url": "https://github.com/dasgarner/support"
```
PeterMis, line 23: After we merge the other PR, I guess this could still point to `xibosignage` support repo? 🤔 
Unless we want to keep like that for now?

PeterMis (approved):
Looks great to me, just a small question below, but it's not overly important

composer.json
dasgarner: You're right - i'll update!

PR #1720, opened 2023-04-05
Layout Editor: placeholder shows behind drop zone when adding a background image #207

rubenpingol-xibo:


maurofmferrao (approved):
Looks good to me!

PR #1721, opened 2023-04-05
Module: Mastodon

ifarzana:
- Relates to  https://github.com/xibosignage/xibo/issues/3042

We could not retrieve location for Mastodon topics 

Widget properties that could not be added as they don't support the following as filter
- Language
- Type (i.e., recent, popular topics)

These are the links we could retrieve data from
- https://mastodon.social/api/v1/timelines/public
- https://mastodon.social/api/v1/timelines/tag (you must provide tag to see results)


modules/mastodon.xml
```diff
@@ -0,0 +1,167 @@
...
+            <rule>required</rule>
+        </property>
+        <property id="tag" type="text">
+            <title>Tag</title>
+            <helpText>Tag. You can test your search tag in the mastodon URL first.</helpText>
+            <default></default>
+        </property>
+        <property id="numItems" type="number">
+            <title>Count</title>
+            <helpText>The number of Tweets to return (default = 15).</helpText>
```
dasgarner, line 50: Should we call this "posts" ?

modules/mastodon.xml
dasgarner, line 1: Can we make sure we remove the word "Tweet" and replace it with whatever is common parlance for Mastodon?

lib/Widget/MastodonProvider.php
```diff
@@ -0,0 +1,147 @@
...
+            }
+            $response = $dataProvider
+                ->getGuzzleClient($httpOptions)
+                ->get($uri);
+
+            $result = json_decode($response->getBody()->getContents(), true);
+
+            $this->getLog()->debug('Mastodon: count: ' . count($result));
+
+            // TODO when should we expire the media, we dont have any settings
```
dasgarner, line 70: We should add this as an option to settings

dasgarner (commented):
This looks like a good start! A few things to fix up and then we can get it into test I think

modules/mastodon.xml
maurofmferrao: I think the official word now is `post`, even if some people refuse to drop the `toot` x)

dasgarner: toot toot 😄 

I think post would be better 👍

ifarzana:
We need the appropriate assets for mastodon though.

dasgarner (commented):
This looks good, but you're going to have to provide those assets now that they have been renamed.

ifarzana:
@dasgarner this `$dataProvider->getSetting('defaultServerUrl', 'https://mastodon.social')` does not give me data from the settings, do you know what I am missing?

dasgarner (approved):
This looks good - looking forward to getting it tested.

dasgarner (approved):

dasgarner (approved):

PR #1725, opened 2023-04-11
Layout Editor: Optimize user experience with background image drag and drop

rubenpingol-xibo:


maurofmferrao (approved):
Looks good to me!

PR #1729, opened 2023-04-11
Modules: Validate/Required input fields on the client before saving

rubenpingol-xibo:


maurofmferrao (approved):
Looks good to me! 

But maybe we also want to apply this to other inputs like dropdown?

maurofmferrao (approved):
Looks good to me!

PR #1731, opened 2023-04-12
Widgets: upgrade on CMS upgrade, widget compatibility task 

ifarzana:
- Relates to https://github.com/xibosignageltd/xibo-private/issues/209


ifarzana:
I am not sure which module should be is using this ArticleWidgetCompatibility class.

ifarzana:
Once all widgets have been upgraded, it should mark any layouts which will have been effected as requires rebuild

- It should not checkout/publish anything - players which already have their content should be left alone and only get the updates when they check in by themselves (we do not want to trigger 1000s of players getting all their content fresh)

db/migrations/20230411090410_add_widget_compatibility_task_migration.php
```diff
@@ -0,0 +1,39 @@
...
+ * Add widget compatibility task
+ * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+ */
+class AddWidgetCompatibilityTaskMigration extends AbstractMigration
+{
+    public function change()
+    {
+        // Check to see if the task exists
+        if (!$this->fetchRow('SELECT * FROM `task` WHERE name = \'Widget Compatibity\'')) {
+            $this->execute('INSERT INTO `task` SET `name`=\'Widget Compatibity\', `class`=\'\\\\Xibo\\\\XTR\\\\WidgetCompatibilityTask\', `status`=2, `isActive`=1, `configFile`=\'/tasks/widget-compatibility.task\', `options`=\'{}\', `runNow`=1, `schedule`=\'0 0 1 1 *\';');
```
dasgarner, line 36: We usually insert tasks using the Phinx syntax - the most recent example being in ad campaign migration. It covers all necessary columns from old dbs.

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,131 @@
...
+    {
+        $this->moduleFactory = $container->get('moduleFactory');
+        $this->widgetFactory = $container->get('widgetFactory');
+        $this->layoutFactory = $container->get('layoutFactory');
+        $this->regionFactory = $container->get('regionFactory');
+        $this->taskFactory = $container->get('taskFactory');
+        return $this;
+    }
+
+    /** @inheritdoc
```
dasgarner, line 68: Can you add some basic commentary to this describing what is happening and why?

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,131 @@
...
+
+            $module = $this->moduleFactory->getByType($widget->type);
+            if ($module->isWidgetCompatibilityAvailable()) {
+                $countWidgets++;
+
+                // Grab a widget compatibility interface, if there is one
+                $widgetCompatibilityInterface = $module->getWidgetCompatibilityOrNull();
+                if ($widgetCompatibilityInterface !== null) {
+                    $this->log->debug('WidgetCompatibilityTask: widgetId ' . $widget->widgetId);
+                    $widgetCompatibilityInterface->setLog($this->log->getLoggerInterface())
```
dasgarner, line 91: Don't you already `setLog` inside `getWidgetCompatibilityOrNull`?

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,131 @@
...
+                        ->upgradeWidget($widget, $widget->schemaVersion, 2);
```
dasgarner, line 92: I think you may want to protect this with a try/catch so that a single widget failing does not stop the whole task?

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,131 @@
...
+
+                    $widget->schemaVersion = 2;
+                    $widget->save(['alwaysUpdate'=>true]);
+                }
+
+                try {
+
```
dasgarner, line 99: Do you have phpcs running in your IDE?

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,131 @@
...
+                    $this->log->error('Failed to set layout rebuild for widgetId: ' . $widget->widgetId .
+                        ', message: ' . $e->getMessage());
+                    $this->appendRunMessage('Layout rebuild error for widgetId: : '. $widget->widgetId);
+                }
+            }
+        }
+
+
+        // Get Widget Compatibility Task
+        $this->compatibilityTask = $this->taskFactory->getByClass('\Xibo\XTR\\WidgetCompatibilityTask');
```
dasgarner, line 117: Does this need to be a member variable?

dasgarner (commented):
Good progress. I had one comment and then I think its time to move onto the Layout Import part of the task.

ifarzana:
@dasgarner I have handled the legacyType condition slightly in a different way. I have arranged the widgetProperties[] to match the condition. Let me know what you think.

lib/Factory/LayoutFactory.php
```diff
@@ -983,7 +983,7 @@ public function loadByJson($layoutJson, $playlistJson, $nestedPlaylistJson, Fold
                 $widget->useDuration = $mediaNode['useDuration'];
                 $widget->tempId = (int)implode(',', $mediaNode['mediaIds']);
                 $widget->tempWidgetId = $mediaNode['widgetId'];
-                $widget->schemaVersion = (int)$mediaNode['schemaVersion'];
+                $widget->schemaVersion = (int)$mediaNode['schemaVersion']; // Todo we get an `undefined array key` error here
```
dasgarner, line 986: I think we can assume its 1 if it doesn't exist.

lib/Factory/LayoutFactory.php
```diff
@@ -1362,6 +1378,29 @@ public function createFromZip(
...
+            $widget->load();
+
+            $module = $this->moduleFactory->getByType($widget->type);
+            if ($module->isWidgetCompatibilityAvailable()) {
+                // Grab a widget compatibility interface, if there is one
+                $widgetCompatibilityInterface = $module->getWidgetCompatibilityOrNull();
+                if ($widgetCompatibilityInterface !== null) {
+                    try {
+                        $widgetCompatibilityInterface->upgradeWidget($widget, $widget->schemaVersion, 2);
+                        if ($widget->schemaVersion == 1) {
```
dasgarner, line 1393: We should only update the schema version if `upgradeWidget` actually did something (i.e. returns true). That way, if we do later make changes to the widget and decide to update its schema, we can run the task/import again and have it upgraded.

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,128 @@
...
+            if ($module->isWidgetCompatibilityAvailable()) {
+                $countWidgets++;
+
+                // Grab a widget compatibility interface, if there is one
+                $widgetCompatibilityInterface = $module->getWidgetCompatibilityOrNull();
+                if ($widgetCompatibilityInterface !== null) {
+                    $this->log->debug('WidgetCompatibilityTask: widgetId ' . $widget->widgetId);
+                    try {
+                        $widgetCompatibilityInterface->upgradeWidget($widget, $widget->schemaVersion, 2);
+                        $widget->schemaVersion = 2;
```
dasgarner, line 92: Same comment here, that we should only do this if `upgradeWidget` has made changes - I suggest we change the interface to return true/false from upgradeWidget?

modules/countdown-text.xml
```diff
@@ -24,6 +24,7 @@
     <author>Core</author>
     <description>A module for displaying a Countdown timer with Simple Text</description>
     <class></class>
+    <compatibilityClass>\Xibo\Widget\Compatibility\CountDownWidgetCompatibility</compatibilityClass>
     <type>countdown-text</type>
     <group id="countdown" icon="countdown">Countdown</group>
     <legacyType>countdown</legacyType>
```
dasgarner, line 30: I think this will need a condition too - if the others do

modules/clock-analogue.xml
```diff
@@ -24,6 +24,7 @@
     <author>Core</author>
     <description>Analogue Clock</description>
     <class></class>
+    <compatibilityClass>\Xibo\Widget\Compatibility\ClockWidgetCompatibility</compatibilityClass>
     <type>clock-analogue</type>
     <group id="clock" icon="clock">Clock</group>
     <legacyType>clock</legacyType>
```
dasgarner, line 30: These clocks might need a condition @maurofmferrao  ?

lib/Factory/ModuleFactory.php
```diff
@@ -264,10 +264,11 @@ public function getEnabled(): array
      * Get module by Type
      * this should return the first module enabled by the type specified.
      * @param string $type
+     * @param array $properties
```
dasgarner, line 267: Can we call this `$conditions` instead and write a short comment about it what it does?

lib/Factory/LayoutFactory.php
```diff
@@ -1037,6 +1028,31 @@ public function loadByJson($layoutJson, $playlistJson, $nestedPlaylistJson, Fold
                     }
                 }
 
+                $widgetProperties = [];
```
dasgarner, line 1031: Interesting way to handle this! Good thinking. I suggest we change the variable to `widgetConditionMatch` or something like that, so that it's clearer what we're doing.

This condition format is also controlled by us, so I think we can make our lives easier here and standardise on a `name==value` syntax (no quotes, etc).

lib/XTR/WidgetCompatibilityTask.php
```diff
@@ -0,0 +1,128 @@
...
+        $widgets = $this->widgetFactory->query(null, [
+            'schemaVersion' => 1
+        ]);
+
+        $countWidgets = 0;
+        foreach ($widgets as $widget) {
+            // Load the widget
+            $widget->load();
+
+            $module = $this->moduleFactory->getByType($widget->type);
```
dasgarner, line 82: You'll need to do the condition matching here too!

dasgarner (commented):
Looks smart!

ifarzana:
@dasgarner, the compatibility classes returns true if it does something has been done in WeatherWidgetCompatibility, do you think that is the right approach? If so, I will do the rest of the classes alike.

dasgarner:
> @dasgarner, the compatibility classes returns true if it does something has been done in WeatherWidgetCompatibility, do you think that is the right approach? If so, I will do the rest of the classes alike.

Yeah I think so - in that class you can probably reduce the repetition by setting a `$newTemplateId=null` variable, changing it in the `case` conditions and then testing if its been set to something after the switch (and then calling setOption/upgraded=true one time instead)

lib/Factory/LayoutFactory.php
```diff
@@ -1017,7 +1017,7 @@ public function loadByJson($layoutJson, $playlistJson, $nestedPlaylistJson, Fold
                 $widget->duration = $mediaNode['duration'];
                 $widget->useDuration = $mediaNode['useDuration'];
                 $widget->tempId = (int)implode(',', $mediaNode['mediaIds']);
-                $widget->tempWidgetId = $mediaNode['widgetId'];
+                $widget->tempWidgetId = $mediaNode['widgetId'];
```
dasgarner, line 1020: Not sure why this is here?

lib/Factory/LayoutFactory.php
```diff
@@ -1397,6 +1411,29 @@ public function createFromZip(
...
+            $module = $this->moduleFactory->getByType($widget->type);
+            if ($module->isWidgetCompatibilityAvailable()) {
+                // Grab a widget compatibility interface, if there is one
+                $widgetCompatibilityInterface = $module->getWidgetCompatibilityOrNull();
+                if ($widgetCompatibilityInterface !== null) {
+                    try {
+                        $upgraded = $widgetCompatibilityInterface->upgradeWidget($widget, $widget->schemaVersion, 2);
+                        if ($upgraded) {
+                            $widget->schemaVersion = 2;
+                            $widget->save(['alwaysUpdate'=>true]);
```
dasgarner, line 1428: It might be wise to explicitly say `notifyDisplays => false` here so that its clear we are not expecting an upgraded widget to trigger a display update?

lib/Factory/LayoutFactory.php
```diff
@@ -1425,7 +1425,7 @@ public function createFromZip(
                         $upgraded = $widgetCompatibilityInterface->upgradeWidget($widget, $widget->schemaVersion, 2);
                         if ($upgraded) {
                             $widget->schemaVersion = 2;
-                            $widget->save(['alwaysUpdate'=>true]);
+                            $widget->save(['alwaysUpdate'=>true, notifyDisplays => false]);
```
dasgarner, line 1428: You're missing some quotes around that string 😄

ifarzana: Sorry :(

dasgarner (approved):

PR #1737, opened 2023-04-14
Misc fixes during testing

dasgarner:
* Widgets: fix issue with saving templateId on edit.
* RSS: fix issue with date comparison deprecation.


maurofmferrao (approved):
Looks good to me!

PR #1749, opened 2023-04-19
Misc fixes during player testing

dasgarner:
This PR addresses some misc issues found during testing, mostly with how widget getData is delivered to the player (and related caching issues with that process).


PeterMis (approved):
Looks good I think

PR #1750, opened 2023-04-19
Scheduling : Full screen schedule and Schedule Grid

PeterMis:
- Full Screen Schedule for Media and Playlist (Schedule Now on their respective grids and new events from Schedule/Calendar pages)
- Schedule Grid
- Fixed OAUTH token

### DIFF
--- a/lib/Controller/Layout.php
+++ b/lib/Controller/Layout.php
@@ -3263,6 +3263,58 @@ class Layout extends Base
     /**
      * Create a Layout with full screen Region with Media/Playlist specific Widget
      * This is called as a first step when scheduling Media/Playlist eventType
+     * @SWG\Post(
+     *  path="/layout/fullscreen",
+     *  operationId="layoutAddFullScreen",
+     *  tags={"layout"},
+     *  summary="Add a Full Screen Layout",
+     *  description="Add a new full screen Layout with specified Media/Playlist",
+     *  @SWG\Parameter(
+     *      name="id",
+     *      in="formData",
+     *      description="The Media or Playlist ID that should be added to this Layout",
+     *      type="integer",
+     *      required=true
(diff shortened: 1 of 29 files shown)


lib/Controller/Layout.php
```diff
@@ -3234,4 +3258,157 @@ public function downloadThumbnail(Request $request, Response $response, $id)
         $this->setNoOutput();
         return $this->render($request, $response);
     }
+
+    public function createFullScreenLayout(Request $request, Response $response)
```
dasgarner, line 3262: Should we have a Swagger Doc for this?

lib/Controller/Layout.php
```diff
@@ -3234,4 +3258,157 @@ public function downloadThumbnail(Request $request, Response $response, $id)
...
+    {
+        $params = $this->getSanitizer($request->getParams());
+        $type = $params->getString('type');
+        $media = null;
+        $playlist = null;
+        $playlistItems = [];
+
+        if (empty($params->getInt('id'))) {
+            throw new InvalidArgumentException('Please select ' . ucfirst($type));
```
dasgarner, line 3271: Translation needed here.

lib/routes.php
```diff
@@ -88,6 +88,7 @@
  *  description="Schedule"
  * )
  */
+$app->get('/schedule/grid/data', ['\Xibo\Controller\Schedule','grid'])->setName('schedule.grid.data');
```
dasgarner, line 91: Is there scope for this to be just `/schedule` and be called `schedule.search` so that it is the same as the other GET's?

dasgarner (commented):
Looks good! I only had a few points

dasgarner (approved):

PR #1753, opened 2023-04-20
Layout Preview: Elements with data #225

rubenpingol-xibo:


modules/src/xibo-elements-render.js
```diff
@@ -19,41 +19,446 @@
...
     // Calculate the dimensions of this item
-    let width = options.width;
-    let height = options.height;
+    let width = 0;
+    let height = 0;
     if (options.scaleOverride !== 0) {
-      width = options.width / options.scaleOverride;
-      height = options.height / options.scaleOverride;
+      // width = options.width / options.scaleOverride;
+      // height = options.height / options.scaleOverride;
```
maurofmferrao, line 63: If scale Override isn't used anymore, should we remove this section altogether?

maurofmferrao (commented):
I just left a minor comment, the rest looks good to me! 👍🏻

maurofmferrao (approved):
Looks good to me!

PR #1755, opened 2023-04-20
Schedule : Fix paging on Schedule grid, default to and from date filters

PeterMis:
- Fix Schedule grid paging
- Add default from/to date filters (to be improved later on with date range input)
- Hide event and campaignId by default (still accessible under `+` button)
- Rename `Layout Duration` to `Duration in Loop` in schedule forms for Media Library event Type
- Attempt at a clever default sort on Schedule grid - first always events, then recurring events, then everything else
- A new grid button to reset the filter (only on Schedule grid at the moment)
- Fix some sorting/ordering issues with Schedule grid

### DIFF
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -2016,7 +2016,11 @@ class Schedule extends Base
         // Call to render the template
         $this->getState()->template = 'schedule-grid-page';
         $this->getState()->setData([
-            'eventTypes' => \Xibo\Entity\Schedule::getEventTypes()
+            'eventTypes' => \Xibo\Entity\Schedule::getEventTypes(),
+            'defaults' => [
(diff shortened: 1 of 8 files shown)


dasgarner (approved):

theofficialgman:
@PeterMis Could you go into more detail on what happened to "Layout Duration" / "Duration in Loop" in schedule forms for media library event types? I cannot find them referenced anywhere in the current codebase but the documentation mentions this feature explicitly for setting the duration of an image when scheduling directly https://xibosignage.com/manual/en/getting_started_showing_an_image

Currently the "select media" form just contains a box to choose the media, no other controls (duration, resolution, colour) which are referenced in the documentation exist. This basically makes this feature unusable for me because the image gets assigned a (seemingly random) duration (not even controlled by the duration set in the media page yet it is different between images which is odd).

dasgarner:
I've replied to the issue you created. It is much easier to keep things in one place please.

PR #1772, opened 2023-05-04
Layout Editor: scale background with CSS

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#230


maurofmferrao (approved):
Looks good to me!

PR #1776, opened 2023-05-04
Layout Import: use basename on files in mapping.json

dasgarner:


dasgarner:
We don't update release18 as 1.8 is EOL.

This PR can be used as a patch if necessary.

PR #1777, opened 2023-05-04
Various bugfixes

dasgarner:


PeterMis (approved):

dasgarner:
I don't think we're going to get Cypress testing working for such an old release (not without significant overhead). We will have a discussion about this internally, but my inclination is to merge this PR without Cypress testing taking place.

PR #1779, opened 2023-05-05
Reporting: Time Connected Summary

ifarzana:
- Fixes https://github.com/xibosignage/xibo/issues/3043

### DIFF
--- a/lib/Report/TimeDisconnectedSummary.php
+++ b/lib/Report/TimeDisconnectedSummary.php
@@ -297,7 +297,6 @@ class TimeDisconnectedSummary implements ReportInterface
 
         $sql = $select . $body;
         $maxDuration = 0;
-        $rows = [];
 
         foreach ($this->store->select($sql, $params) as $row) {
             $maxDuration = $maxDuration + $this->sanitizer->getSanitizer($row)->getDouble('duration');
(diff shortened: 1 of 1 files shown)


ifarzana:
@dasgarner, Can you check if I can refactor the `$displaySql`?

lib/Report/TimeDisconnectedSummary.php
```diff
@@ -386,6 +408,105 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+        if ($onlyLoggedIn) {
+            $displayBody .= ' AND `display`.loggedIn = 1 ';
+        }
+
+        $displayBody .= '
+            GROUP BY display.display, display.displayId
+        ';
+
+        // Get a list of displays by filters
+        $displaySql = $displaySelect . $displayBody;
```
dasgarner, line 490: Since we removed paging, we don't need to separate the displaySelect and displayBody - the whole query can just be displaySql.

dasgarner (commented):
I think the main query is the displays query, and that the disconnected displays query can be simplified to just a list of events with displayId (no need to add tags, etc to that).

lib/Report/TimeDisconnectedSummary.php
```diff
@@ -320,79 +394,47 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
-            $entry['displayId'] = $sanitizedRow->getInt(('displayId'));
-            $entry['display'] = $sanitizedRow->getString(('display'));
-            $entry['timeDisconnected'] =  round($sanitizedRow->getDouble('duration') / $divisor, 2);
-            $entry['timeConnected'] =  round($sanitizedRow->getDouble('filter') / $divisor - $entry['timeDisconnected'], 2);
-            $entry['postUnits'] = $postUnits;
-
-            $rows[] = $entry;
+        // Displays by filter
+        $displays = [];
+        foreach ($this->store->select($displaySql, $params) as $displayRow) {
```
dasgarner, line 422: Doing two loops means that the sorting is broken? 

You should only add to `$rows` in this loop, and always pull in the time disconnected/connected from the `$disconnectedDisplays` array.

dasgarner (changes requested):

dasgarner (approved):

PR #1780, opened 2023-05-05
XMDS PHPUnit tests and other fixes

PeterMis:
## XMDS tests
- Couple new XMDS tests (GetData, GetDependency etc), this is still WIP, there are more tests to be added etc
- New xmdsTestCase - streamlined testCase we only need Guzzle client
- Adjusted phpunit.xml for phpunit 10+
- LocalWebTestCase : Fix return values for functions (phpunit 10+), as it is still included in Bootstrap it will throw fatal error without that.

### Other fixes
- Players : Fix downloading "dependencies" (bundle, fonts, assets etc) on older players (pre xmds 7)
- Layout Build : Do not attempt to include module properties without id in the xlf
- Layout Publish : Fix publishing a Layout with future date (Regular Maintenance)
- Library : set folderId from oldMedia on replace
- Tag Factory : Fix php warning about null in trim for tagValue
- User Group : Fix dynamic property libraryQuotaFormatted

### DIFF
--- a/lib/Factory/RequiredFileFactory.php
+++ b/lib/Factory/RequiredFileFactory.php
@@ -129,6 +129,10 @@ class RequiredFileFactory extends BaseFactory
      */
     public function getByDisplayAndDependency($displayId, $fileType, $id, bool $isUseRealId = true)
     {
+        if (!$isUseRealId && $id < 0) {
+            $fileType = self::getLegacyFileType($id);
+        }
+
         $result = $this->getStore()->select('
             SELECT * 
               FROM `requiredfile` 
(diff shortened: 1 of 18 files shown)


lib/Factory/RequiredFileFactory.php
```diff
@@ -150,6 +154,21 @@ public function getByDisplayAndDependency($displayId, $fileType, $id, bool $isUs
...
 
+    /**
+    * Return the fileType depending on the legacyId range
+    * @param $id
+    * @return string
+    */
+    private static function getLegacyFileType($id): string
+    {
+        return match (true) {
+            $id < 0 && $id > -100000000 => 'bundle',
```
dasgarner, line 165: It would be clearer to use the static variables here I think?

tests/Bootstrap.php
```diff
@@ -27,6 +28,7 @@
 
 require_once PROJECT_ROOT . '/vendor/autoload.php';
 require_once PROJECT_ROOT . '/tests/LocalWebTestCase.php';
+require_once PROJECT_ROOT . '/tests/xmdsTestCase.php';
```
dasgarner, line 31: Should have a capital letter?

dasgarner (commented):
Those test cases look great. I have a few very minor comments.

dasgarner (approved):

PR #1785, opened 2023-05-09
Modules: use sample data text instead of actual data text

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#228


maurofmferrao (approved):
Looks good to me!

PR #1788, opened 2023-05-09
Widgets: Data Providers should be explicitly handled

dasgarner:
So we can present an error message if there are configuration issues. 
related to xibosignageltd/xibo-private#236


PeterMis (approved):
Looks reasonable!

PR #1793, opened 2023-05-12
Stocks Module: Stock symbol search

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#196

### DIFF
--- a/lib/Connector/AlphaVantageConnector.php
+++ b/lib/Connector/AlphaVantageConnector.php
@@ -27,9 +27,11 @@ use GuzzleHttp\Exception\GuzzleException;
 use Stash\Invalidation;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Xibo\Event\WidgetDataRequestEvent;
+use Xibo\Event\WidgetEditOptionRequestEvent;
 use Xibo\Support\Exception\ConfigurationException;
 use Xibo\Support\Exception\GeneralException;
 use Xibo\Support\Exception\InvalidArgumentException;
+use Xibo\Support\Exception\NotFoundException;
 use Xibo\Support\Sanitizer\SanitizerInterface;
 use Xibo\Widget\Provider\DataProviderInterface;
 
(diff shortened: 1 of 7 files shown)


lib/Connector/AlphaVantageConnector.php
```diff
@@ -118,6 +121,83 @@ public function onDataRequest(WidgetDataRequestEvent $event)
...
+
+        // Pull the widget we're working with.
+        $widget = $event->getWidget();
+        if ($widget === null) {
+            throw new NotFoundException();
+        }
+
+        // We handle the twitter widget and the property with id="type"
+        if ($widget->type === 'stocks' && $event->getPropertyId() === 'searchSymbol') {
+
```
dasgarner, line 136: It would be good to make sure `phpcs` checking is working in your IDE, because I think this empty line will flag up here.

lib/Connector/AlphaVantageConnector.php
```diff
@@ -118,6 +121,83 @@ public function onDataRequest(WidgetDataRequestEvent $event)
...
+            if (array_key_exists('bestMatches', $data)) {
+                return $data['bestMatches'];
+            }
+
+            return [];
+
+        } catch (GuzzleException $guzzleException) {
+            throw new GeneralException(
+                'Guzzle exception getting Stocks data . E = ' .
```
dasgarner, line 193: It is our convention to have the `.` on the next line (which is different to the JS convention)

dasgarner (changes requested):
Some suggestions 😄

lib/Connector/AlphaVantageConnector.php
```diff
@@ -164,19 +169,20 @@ public function onWidgetEditOption(WidgetEditOptionRequestEvent $event)
...
      * @throws GeneralException
      */
-    private function getSearchResults(string $keywords): array
+    private function getSearchResults(string $keywords): array|bool
     {
         try {
             $this->getLogger()->debug('AlphaVantage Connector : getSearchResults is served from the API.');
 
-            $request = $this->getClient()->request('GET', 'https://www.alphavantage.co/query', [
+            // $request = $this->getClient()->request('GET', 'https://www.alphavantage.co/query', [
```
dasgarner, line 180: Can we remove this comment and the API key comment?

dasgarner (changes requested):
Just some comments to remove please

dasgarner (approved):

PR #1794, opened 2023-05-12
Widgets: compatibility interface for Sub Playlist

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/237

### DIFF
--- a/lib/Widget/Compatibility/SubPlaylistWidgetCompatibility.php
+++ b/lib/Widget/Compatibility/SubPlaylistWidgetCompatibility.php
@@ -1,0 +1,98 @@
+<?php
+/*
+ * Copyright (C) 2023 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


ifarzana:
@dasgarner, I was not able to test this completely. I could get a 3.2 layout zip file which has `subPlaylistIds` and `subPlaylistOptions` but it appears that the compatibility class isn't recognising those options. Right now, I require your assistance in identifying if I made any errors.

Thanks

lib/Widget/Compatibility/SubPlaylistWidgetCompatibility.php
```diff
@@ -0,0 +1,99 @@
...
+
+                $playlistItems[] = $item;
+            }
+
+            if (count($playlistItems) > 0) {
+                $widget->setOptionValue('subPlaylists', 'attrib', json_encode($playlistItems));
+                $upgraded = true;
+            }
+        }
+
```
dasgarner, line 91: Regardless I think we should `removeOption` for the old legacy ones.

ifarzana:
> I think the LayoutFactory currently does this manually. Can you have a look at `convertOldPlaylistOptions()` in LayoutFactory to see what it does?

Peter pointed out that to me as well. `convertOldPlaylistOptions()`  already doing what I want to achieve in upgradeWidget.
However, If I run the `WidgetCompatibilityTask` the upgradeWidget does convert the options as expected.

> In the case of widgets on the main layout, I think we can probably just remove the call, but I think in the case of nested sub playlists we might also need to pass those through the widget compatibility interface.

Can you clarify a bit more please?

dasgarner:
We should remove the `convertOldPlaylistOptions()` code in LayoutFactory and always have that handled using WidgetCompatibilityInterface. This means having the widget compatibility interface called for nested subplaylists when we import (currently these widgets are no being upgraded).

> However, If I run the WidgetCompatibilityTask for a 3.2 widget the upgradeWidget does convert the options as expected.

We have to find out why - you'll need to add some logging to trace what is happening and why they are not converting - because the code looks good. Happy to step through this with you if that is useful.

ifarzana:
Ready for review and tested with nested playlists which imported perfectly. Thanks

PR #1805, opened 2023-05-19
Element: Line

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#247

### DIFF
--- a/modules/templates/global-elements.xml
+++ b/modules/templates/global-elements.xml
@@ -485,4 +485,155 @@ $('.date').each(function(_idx, dateEl){
             <asset id="image" type="path" cmsOnly="true" mimeType="image/png" path="/modules/assets/template-thumbnails/global/image.png" />
         </assets>
     </template>
+    <template>
+        <id>line</id>
+        <title>Line</title>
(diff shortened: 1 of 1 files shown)


modules/templates/global-elements.xml
```diff
@@ -485,4 +485,155 @@ $('.date').each(function(_idx, dateEl){
             <asset id="image" type="path" cmsOnly="true" mimeType="image/png" path="/modules/assets/template-thumbnails/global/image.png" />
         </assets>
     </template>
+    <template>
+        <id>line</id>
+        <title>Line</title>
+        <type>element</type>
+        <dataType>global</dataType>
+        <thumbnail>square-thumb</thumbnail>
```
maurofmferrao, line 493: We need to change the thumbnail for this element, I will send you one ( that might be temporary ) just we can easily distinguish between Line and Square in the toolbar

maurofmferrao (changes requested):
Looks good, I just have 2 minor changes :)

maurofmferrao (approved):
Looks good to me!

PR #1810, opened 2023-05-22
Minor fixes

dasgarner:
- add missing local video widget
 - fix dataset layout import in light of new v4 widget options
 - don't notify dataset on layout import, it can't be active as its not imported yet


maurofmferrao (approved):
Looks good!

PR #1811, opened 2023-05-22
Content sync

PeterMis:
#### Sync Groups
- new Controller and routes
- new Entity,
- new Factory
- new permissions entity / folder handling
- in short : Create a sync group, add Display members, set lead Display and use it in a new Sync Schedule event

#### Sync Schedule
- new event type
- new add/edit forms with content selector (for now only Layouts) for each member of the Sync group
with shortcut button to set the same layout on all members (Mirror sync)
- Adjustments to grid/calendar views to better represent new sync events

#### XMDS
- New sync event type should be added to schedule/rf calls with layoutId set on the calling Display
- Register will now return `syncGroup` attribute (for Displays in a sync group) populated with either `lead` or lanIpAddress of the lead Display


db/migrations/20230509113820_content_sync_migration.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+class ContentSyncMigration extends AbstractMigration
+{
+    public function change()
+    {
+        $this->table('syncgroup', ['id' => 'syncGroupId'])
+            ->addColumn('name', 'string', ['limit' => 50])
+            ->addColumn('createdDt', 'datetime', ['null' => true, 'default' => null])
+            ->addColumn('modifiedDt', 'datetime', ['null' => true, 'default' => null])
+            ->addColumn('ownerId', 'integer')
```
dasgarner, line 37: `modifiedBy` would be good too - last userId who modified the group in any way.

lib/Service/DisplayNotifyService.php
```diff
@@ -316,6 +316,23 @@ public function notifyByCampaignId($campaignId)
...
+                schedule.recurrence_range AS recurrenceRange,
+                schedule.recurrenceRepeatsOn,
+                schedule.lastRecurrenceWatermark,
+                schedule.dayPartId
+             FROM `schedule`
+                INNER JOIN `schedule_sync`
+                    ON `schedule_sync`.eventId = `schedule`.eventId
+                INNER JOIN `lkcampaignlayout`
+                    ON `lkcampaignlayout`.layoutId = `schedule_sync`.layoutId
+             WHERE `lkcampaignlayout`.campaignId = :assignedCampaignId
```
dasgarner, line 335: You'll need to have the date/time clauses from the first part of this SQL union to restrict to active sync events.

lib/Xmds/Soap.php
```diff
@@ -1222,7 +1224,15 @@ protected function doSchedule($serverKey, $hardwareKey, $options = [])
 
                 foreach ($scheduleEvents as $scheduleEvent) {
                     $eventTypeId = $row['eventTypeId'];
-                    $layoutId = $row['layoutId'];
+                    $layoutId = ($row['eventTypeId'] == Schedule::$SYNC_EVENT)
```
dasgarner, line 1227: This might be clearer with a normal `if` condition

lib/Listener/SyncGroupListener.php
```diff
@@ -0,0 +1,180 @@
...
+     */
+    public function onScheduleSave(ScheduleSaveEvent $event): void
+    {
+        $scheduleEvent = $event->getEvent();
+        $params = $event->getParams();
+        if ($scheduleEvent->eventTypeId === Schedule::$SYNC_EVENT) {
+            $syncGroup = $this->syncGroupFactory->getById($scheduleEvent->syncGroupId);
+
+            foreach ($syncGroup->getSyncGroupMembers() as $display) {
+                $this->storageService->insert('INSERT INTO `schedule_sync` (`eventId`, `displayId`, `layoutId`)
```
dasgarner, line 153: I think we usually do this type of insert in the entity - so in this case the schedule entity. I'd perhaps expect this to be in `schedule->save()`

lib/Controller/Schedule.php
```diff
@@ -1208,6 +1219,13 @@ public function add(Request $request, Response $response)
             ]);
         }
 
+        if ($this->isSyncEvent($schedule->eventTypeId)) {
```
dasgarner, line 1222: I think this is to avoid DI of another factory - however given we've recently added Playlist/Media factories I think it makes sense to add the SyncGroupFactory here too, thereby avoiding the need for an event to load/save those items.

lib/Controller/Schedule.php
```diff
@@ -344,7 +346,14 @@ public function eventData(Request $request, Response $response)
                 && $this->isEventEditable($row);
 
             // Event Title
-            if ($row->campaignId == 0) {
+            if ($this->isSyncEvent($row->eventTypeId)) {
+                $this->getDispatcher()->dispatch(new ScheduleLoadEvent($row), ScheduleLoadEvent::$NAME);
```
dasgarner, line 350: This potentially generates a lot of events and I wonder if we need to do this in this way. For example, can we instead add to the originating query (perhaps optionally via a parameter), and just name the sync group.

dasgarner (commented):
Overall this looks excellent, I just have a few comments.

dasgarner (approved):

PR #1813, opened 2023-05-23
Properties panel: Add check if condition target exists

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#256


maurofmferrao (approved):
Looks good to me!

PR #1814, opened 2023-05-23
Validation of properties/widgets

dasgarner:
Layout status and widget properties were not being validated and we're missing some validation logic from v3. 

 - Add rule parsing to the XML for simple conditional checks
 - Add validation interface for more complex "whole widget" checks


maurofmferrao (approved):
Looks good to me!

PR #1816, opened 2023-05-24
Element Line: Fix positioning for all tips

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#247


maurofmferrao (approved):
Looks good!

PR #1818, opened 2023-05-24
Translations from the config XML files 

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/183

### DIFF
--- a/bin/locale.php
+++ b/bin/locale.php
@@ -40,10 +40,6 @@ use Xibo\Twig\TwigMessages;
 define('PROJECT_ROOT', realpath(__DIR__ . '/..'));
 require_once PROJECT_ROOT . '/vendor/autoload.php';
 
-if (!file_exists(PROJECT_ROOT . '/web/settings.php')) {
-    die('Not configured');
-}
-
 $view = Twig::create([
     PROJECT_ROOT . '/views',
     PROJECT_ROOT . '/modules',
@@ -87,7 +83,7 @@ class MockConfigService extends ConfigService
     }
 }
 
-// Transactor funtion
+// Translator function
 function __($original)
 {
     return $original;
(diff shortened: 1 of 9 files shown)


ifarzana:
@dasgarner I am getting an error `PHP Fatal error:  Uncaught LogicException: Unable to register extension "Slim\Views\TwigExtension" as it is already registered. in /var/www/cms/vendor/twig/twig/src/ExtensionSet.php:132
Stack trace:` 

Do you see anything that I am not doing correctly?

dasgarner:
I don't think creating the whole application stack is the right way to pursue this. We discussed creating the ModuleFactory and ModuleTemplateFactory manually instead?

To do that you'd need to use `new ModuleFactory()` and supply mocked objects for its dependencies.

I think this would be workable because the bit we want (the XML parsing) doesn't rely on any of those dependencies (I think)

bin/locale.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Copyright (C) 2019 Xibo Signage Ltd
```
dasgarner, line 4: Odd that this copyright didn't update? is the profile setup in your IDE?

bin/locale.php
```diff
@@ -38,6 +40,10 @@
 define('PROJECT_ROOT', realpath(__DIR__ . '/..'));
 require_once PROJECT_ROOT . '/vendor/autoload.php';
 
+if (!file_exists(PROJECT_ROOT . '/web/settings.php')) {
```
dasgarner, line 43: I don't think we need this?

bin/locale.php
```diff
@@ -55,14 +61,155 @@
...
+// Mock Config Service
+class MockConfigService extends ConfigService
+{
+    public function getSetting($setting, $default = null, $full = false)
+    {
+        return '';
+    }
+}
+
+// Transactor funtion
```
dasgarner, line 90:
```suggestion
// Translator function
```

bin/locale.php
```diff
@@ -55,14 +61,155 @@
...
+);
+$moduleTemplateFactory->useBaseDependenciesService($baseDepenencyService);
+// Get all module templates
+$moduleTemplates = $moduleTemplateFactory->getAll();
+
+// --------------
+// Create translation file
+// Each line contains title or description or properties of the module/templates
+$file = PROJECT_ROOT. '/locale/moduletranslate.php';
+if (!file_exists($file)) {
```
dasgarner, line 130: I don't think you need to do this - file_put_contents will create a file if it doesn't exist and overwrite one if it does.

bin/locale.php
```diff
@@ -55,14 +61,155 @@
...
+                foreach ($setting->options as $option) {
+                    if (!empty($option->title)) {
+                        $content .= "echo __('$option->title');" . PHP_EOL;
+                    }
+                }
+            }
+        }
+
+        // Properties translation
+        foreach ($module->properties as $property) {
```
dasgarner, line 164: Sorry, I did add another requirement regarding parsing a rule message :smile:

dasgarner (changes requested):
Looks like you've cracked it - just a few suggestions.

dasgarner (approved):
Looks good!

PR #1819, opened 2023-05-25
Stencils: Allow editing element in the group

maurofmferrao:
relates to xibosignageltd/xibo-private#266


dasgarner (approved):
I think this looks good.

PR #1822, opened 2023-05-26
Layout: canvas region should have a global widget automatically 

dasgarner:
This should be created for it in region add.
The canvas region should only be deleted if that widget is empty.


maurofmferrao (approved):
Looks good to me!

PR #1831, opened 2023-05-31
Player bundle: Add babel plugin to support object rest spread

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#278


dasgarner (approved):
Looks good to me.

PR #1840, opened 2023-06-01
3.3.6 issues

PeterMis:


lib/Controller/Module.php
```diff
@@ -799,7 +799,8 @@ public function editWidgetForm(Request $request, Response $response, $id)
             'validExtensions' => str_replace(',', '|', $module->getModule()->validExtensions),
             'templatesAvailable' => $templates,
             'options' => $options,
-            'isTopLevel' => $this->playlistFactory->getById($module->widget->playlistId)->isRegionPlaylist()
+            'isTopLevel' => $this->playlistFactory->getById($module->widget->playlistId)->isRegionPlaylist(),
+            'maxImagePixelSize' => $media->mediaType == 'image' ? $this->getConfig()->getSetting('DEFAULT_RESIZE_LIMIT') : null,
```
dasgarner, line 803: As this is a general setting and we're not returning something module specific (like valid extensions), can we use the value you already have assigned to the Twig view? `libraryUpload.maxImagePixelSize` ?

dasgarner: Or even better, can we just pop it on the file upload template in the main twig include and always get it from there?

PeterMis: Good point!

We have settings in the twig view, so there is actually no need to even add this to the libraryUpload.
Just take it from the settings in the main file upload twig

dasgarner (approved):

PR #1844, opened 2023-06-02
Widget: validation to support type=OR and move message to test node

dasgarner:
fixes xibosignageltd/xibo-private#277


maurofmferrao (approved):
Looks good to me!

PR #1846, opened 2023-06-05
Misc minor fixes

dasgarner:
Widget: agenda/calendar validation of custom interval xibosignage/xibo#3060
Connector: stock breaks on invalid symbol
XMDS: GetDependency without http wasn't working for fonts
Layout Import: using an existing dataset did not work


PeterMis (approved):
Looks good!

PR #1847, opened 2023-06-06
Properties panel: Add date format helper popup

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#274


maurofmferrao (approved):
Looks good!

PR #1851, opened 2023-06-06
Fix Translations

ifarzana:
- Fixes https://github.com/xibosignageltd/xibo-private/issues/183

### DIFF
--- a/lib/Widget/Definition/Rule.php
+++ b/lib/Widget/Definition/Rule.php
@@ -31,9 +31,6 @@ class Rule implements \JsonSerializable
 
     public $onStatus = true;
 
-    /** @var string|null */
-    public $message;
-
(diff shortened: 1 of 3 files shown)


lib/Widget/Definition/Rule.php
```diff
@@ -31,6 +31,9 @@ class Rule implements \JsonSerializable
 
     public $onStatus = true;
 
+    /** @var string|null */
+    public $message;
```
dasgarner, line 35: This has just been moved to the test node (on purpose). Sorry if I was unclear, but I meant "use the property on test in locale.php"

dasgarner (changes requested):

dasgarner (approved):

PR #1860, opened 2023-06-13
Elements Shapes: Add circle, triangle, pentagon and hexagon shape

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#287


maurofmferrao (approved):
Looks good!

PR #1862, opened 2023-06-13
Fix Widget Schema Version

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/286


dasgarner (approved):

dasgarner:
Looking at the `schemeVersion` we have in our module XML definitions, some of them appear to be incorrect. If we set upgraded widgets to have a schemaVersion equal to the module, then any widget which has a compatibilityClass must have at least schema version 2.

I don't understand why the embedded, mastodon, image or vidoe modules have a schemaVersion of 2 (nothing seems to have changed between CMS v3 and CMS v4 for the those modules).

dasgarner (approved):

PR #1863, opened 2023-06-14
Elements Calendar: Add calendar data as widget elements

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#295


maurofmferrao (approved):
Looks good!

PR #1871, opened 2023-06-15
Layout Editor: Automatic save

maurofmferrao:
relates to xibosignageltd/xibo-private#166

 - Layout background image triggers save
 - Toast messages removed
 - Loading overlay changed to a corner icon
 - Fix style for date inputs in read only mode
 - Layout Editor help added to toolbar
 - Toolbar not refreshing when reloading viewer

### DIFF
--- a/lib/Controller/Layout.php
+++ b/lib/Controller/Layout.php
@@ -272,6 +272,8 @@ class Layout extends Base
             'users' => $this->userFactory->query(),
             'modules' => $moduleFactory->getAssignableModules(),
             'timeZones' => $timeZones,
+            'help' => $this->getHelp()->link('Layout', 'Edit'),
+
         ];
 
         // Call the render the template
--- a/lib/Controller/Layout.php
+++ b/lib/Controller/Layout.php
@@ -273,7 +273,6 @@ class Layout extends Base
             'modules' => $moduleFactory->getAssignableModules(),
             'timeZones' => $timeZones,
             'help' => $this->getHelp()->link('Layout', 'Edit'),
-
         ];
 
         // Call the render the template
--- a/modules/templates/social-media-elements.xml
+++ b/modules/templates/social-media-elements.xml
@@ -115,7 +115,7 @@
                     <height>40</height>
                     <rotation>0</rotation>
                     <defaultProperties>
-                        <property id="fontFace">open sans regular</property>
+                        <property id="fontFamily">aileron regular</property>
                         <property id="bold">1</property>
                         <property id="fontSize">24</property>
                         <property id="fontColor">#3c3ad3</property>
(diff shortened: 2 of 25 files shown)


lib/Controller/Layout.php
```diff
@@ -272,6 +272,8 @@ public function displayDesigner(Request $request, Response $response, $id)
             'users' => $this->userFactory->query(),
             'modules' => $moduleFactory->getAssignableModules(),
             'timeZones' => $timeZones,
+            'help' => $this->getHelp()->link('Layout', 'Edit'),
+
```
dasgarner, line 276: phpcs

modules/templates/social-media-elements.xml
```diff
@@ -115,7 +115,7 @@
                     <height>40</height>
                     <rotation>0</rotation>
                     <defaultProperties>
-                        <property id="fontFace">Open Sans</property>
+                        <property id="fontFace">open sans regular</property>
```
dasgarner, line 118: This isn't a font we ship

ui/src/templates/forms/region.hbs
```diff
@@ -7,21 +7,21 @@
         <form id="regionFormEdit" class="XiboForm">
             <div class="tab-content">
                 <div class="tab-pane active" id="generalTab">
-                    {{> inputs/hidden id="type" value=region.type }}
-                    {{> inputs/text value=region.name id="name" title="Name" helpText="An optional name for this widget" }}
-                    {{> inputs/checkbox value=(getOption region.regionOptions "loop") id="loop" title="Loop?" helpText="If there is only one item in this region should it loop?" }}
+                    {{> inputs/hidden id="type" value=region.type customClass="position-input" }}
+                    {{> inputs/text value=region.name id="name" title="Name" helpText="An optional name for this widget" customClass="position-input" }}
```
dasgarner, line 11: We have a bunch of stuff here that isn't translated, so we need to have a think about how this gets handled (not for this PR, but we can open an issue for it please)

dasgarner (changes requested):
A few comments on this one (minor things)

lib/Controller/Layout.php
maurofmferrao: the empty new line between the new code and the end of the array right?

modules/templates/social-media-elements.xml
maurofmferrao: Changed to a font in the default theme

ui/src/templates/forms/region.hbs
maurofmferrao: I've created a new issue for this

dasgarner (approved):

PR #1876, opened 2023-06-20
Schedule : full screen media and playlist schedules and schedule page adjustments

PeterMis:
- New control for full screen media/playlist instead of select2
On button click opens a new small modal with required fields, on submit creates/fetches the relevant layout and add its layout specific campaignId to the main add/edit schedule form for further editing/submit.
- Schedule Now : readonly select2 dropdowns where needed, on submit with Media/Playlist it will first call the endpoint to create/fetch full screen layout and then submit the form
- Schedule grid/calendar : 
-- separate display/displaygroup filters
-- notice/alert if no display/displaygroup selected (this is required to see events in the calendar)
-- closing swiper/drawer on monthly view should no longer switch to agenda view (double click on the calendar cell will)
-- going back to grid tab from agenda view will switch the view/range filter to today to make it easier to understand
-- the `Month` button while on agenda view renamed to `Back`

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -758,6 +758,7 @@ class Library extends Base
 
             // Schedule Now
             if ($this->getUser()->featureEnabled('schedule.now')
+                && in_array($media->mediaType, ['image', 'video'])
                 && ($this->getUser()->checkEditable($media)
                     || $this->getConfig()->getSetting('SCHEDULE_WITH_VIEW_PERMISSION') == 1)
             ) {
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -153,6 +153,7 @@ class Schedule extends Base
         }
 
         $displayGroups = [];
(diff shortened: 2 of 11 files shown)


dasgarner (approved):

theofficialgman:
@PeterMis this is the cause of the regression that I reported in https://github.com/xibosignage/xibo-cms/pull/1755#issuecomment-3315268445 . You removed the ability to edit the duration as part of this PR

theofficialgman:
CC @mgbaybay @maurofmferrao since you two were also part of multiple significant refactors that resulted in the feature (being able to set the duration for an image in a schedule among other things) being removed (likely unintentionally). The last place where that feature can be found is in `views/schedule-form-now.twig` (which basically came from `views/schedule-form-now.twig`) but that file no longer exists and I couldn't root cause what commit deleted it with git.

dasgarner:
I've replied to the issue you created. It is much easier to keep things in one place please.

PR #1886, opened 2023-06-28
Elements: Currencies

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#302


ui/src/core/xibo-cms.js
```diff
@@ -430,12 +430,10 @@ function XiboInitialise(scope, options) {
...
             return settings;
           },
-          filter: function(list) {
-            return $.map(list.data, function(tagObj) {
-              return {
-                tag: tagObj.type,
-              };
-            });
+          transform: function(list) {
+            return list?.data?.length > 0 ? $.map(list.data, function(tagObj) {
```
dasgarner, line 434: `xibo-cms.js` isn't yet run through babel, so when we're in this file we need to make sure we don't use any JS functions which exceed our minimum browser requirements for using the CMS.

dasgarner (commented):
The PHP side looks great - I've added a comment on one part for you to look at.

modules/widget-html-render.twig
rubenpingol-xibo, line 1: @maurofmferrao @dasgarner We are introducing a new callback function `onElementParseData` for element templates. The reason for this is we only have have `onParseData` for widgets. This new callback function can be used for element template.

modules/templates/currency-elements.xml
```diff
@@ -0,0 +1,104 @@
...
+        <id>currency_changePercentage</id>
+        <extends override="text" with="data.ChangePercentage">text</extends>
+        <title>Change Percentage</title>
+        <type>element</type>
+        <dataType>currency</dataType>
+        <canRotate>true</canRotate>
+        <thumbnail>currency-change-percentage</thumbnail>
+        <startWidth>650</startWidth>
+        <startHeight>250</startHeight>
+        <onElementParseData><![CDATA[
```
rubenpingol-xibo, line 60: This is where `onElementParseData` is currently used since we have **%** string for **Change Percentage** property of the currency.

maurofmferrao (approved):
Looks good to me!

ui/src/core/xibo-cms.js
rubenpingol-xibo: This is noted and will updated the PR. Thanks!

maurofmferrao (approved):
Looks good!

PR #1889, opened 2023-07-04
Elements: Stocks and Currencies

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#302 xibosignageltd/xibo-private#296


maurofmferrao (approved):
Looks good to me!

PR #1890, opened 2023-07-05
Replace schedule now and misc fixes

PeterMis:
- Schedule now replaced with main schedule add, relates to https://github.com/xibosignageltd/xibo-private/issues/273
- Fixed folder issue for new Layouts, relates to https://github.com/xibosignageltd/xibo-private/issues/316
- Fixed folder issue on edit Sync groups, relates to https://github.com/xibosignage/xibo/issues/3071
- Fixed issue with Layout code on full screen media/playlist Layouts being an empty string, causing issues with Layout Code dropdown
- Fixed an issue with private folders not being correctly disabled in the folder tree


lib/Controller/Campaign.php
```diff
@@ -282,12 +282,12 @@ public function grid(Request $request, Response $response)
             $campaign->includeProperty('buttons');
             $campaign->buttons = [];
 
-            // Schedule Now
+            // Schedule
             if ($this->getUser()->featureEnabled('schedule.now') && $campaign->type === 'list') {
```
dasgarner, line 286: Should we remove the `schedule.now` feature and then test `schedule.add` instead?

dasgarner (commented):
This looks good, I only have 1 query.

lib/Controller/Campaign.php
PeterMis: That's done now in the commit below

dasgarner (approved):

PR #1907, opened 2023-07-11
Elements: Stocks and Currency

rubenberttpingol:
resolves xibosignageltd/xibo-private#302 xibosignageltd/xibo-private#296


maurofmferrao (approved):
Looks good!

PR #1909, opened 2023-07-11
Modules: add onDataError and API doc generation fixes

dasgarner:
Modules: onDataError, relates to xibosignageltd/xibo-private#264
Swagger: fix openAPI annotations


rubenpingol-xibo (approved):
Looks good to me!

PR #1910, opened 2023-07-11
Widgets: ticker marquee improvements for android

dasgarner:
Added feature.js to see if we can understand why the new marquee plugin didn't work on the DSCS9 - need to test whether it works on the newer 9X/95.

I've also updated the marquee version of the "new" marquee.

I have not need able to fix xibosignageltd/xibo-private#352


dasgarner:
@maurofmferrao requested your review to see if you think this is the right direction?

maurofmferrao (approved):
Looks good to me!

dasgarner:
There is something not right here - when playing in Android (chromium 87) the final part of the marquee shrinks down to the left of the screen.

maurofmferrao (approved):
Looks good!

PR #1911, opened 2023-07-12
Elements: Weather

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#323


modules/forecastio.xml
```diff
@@ -111,6 +111,7 @@
     </stencil>
     <onRender><![CDATA[
 // If we don't get any results, then we should show an error message
+console.log({items});
```
maurofmferrao, line 114: Maybe we want to remove this log? :)

maurofmferrao (changes requested):
Just a really minor change, all the rest looks good to me!

maurofmferrao (approved):
Looks good to me!

PR #1916, opened 2023-07-14
PHPCS: run action with every PR. xibosignageltd/xibo-private#340

dasgarner:
Try adding the phpcs action as a separate workflow.


ifarzana (approved):
Looks good!

PR #1922, opened 2023-07-19
Elements: Weather with added elements and forecast days offset clean up

rubenpingol-xibo:
resolves xibosignageltd/xibo-private#323


maurofmferrao (approved):
Looks good!

PR #1924, opened 2023-07-19
Help system

dasgarner:
fixes xibosignageltd/xibo-private#337

This PR:
 * upgraded phinx in order to include a YAML parser
 * fixes phinx migrations for the new version
 * removes the old help system (help buttons on each form)
 * adds a new help popup IF `help_links.yaml` can be pulled from the manual
 * removes emojione
 * converts all tables to utf8mb4


ifarzana (approved):
I am approving this. Fingers crossed that it works when we merge.

PR #1925, opened 2023-07-19
Fix marquee nesting and add seamless scroll support

maurofmferrao:
relates to xibosignageltd/xibo-private#352


modules/vendor/jquery.marquee.min.js
dasgarner, line 1: Mr latest PR changed this to the most recent version of the plugin - did you put it back to how it was?

PR #1942, opened 2023-07-25
Schedule meta and various fixes

PeterMis:
- Schedule name : relates to xibosignageltd/xibo-private#333
- Schedule meta : relates to xibosignage/xibo#3051
- Schedule Grid : Fixed Event start/end time representation with user created Dayparts
- Schedule add/edit forms : Fixed Display/groups grouping for Display select
- Schedule Grid : vertical aligned all tds to middle
- Members form (mostly relates to sync group) : Fixed the saving cog icon stuck on save button on error
- Removed schedule-grid-page.twig as we no longer use that (it's all on schedule-page)


lib/Controller/Schedule.php
```diff
@@ -1696,6 +1697,8 @@ public function edit(Request $request, Response $response, $id)
         $schedule->actionTriggerCode = $sanitizedParams->getString('actionTriggerCode');
         $schedule->actionLayoutCode = $sanitizedParams->getString('actionLayoutCode');
         $schedule->maxPlaysPerHour = $sanitizedParams->getInt('maxPlaysPerHour', ['default' => 0]);
+        $schedule->name = $sanitizedParams->getString('name');
+        $schedule->modifiedBy = $this->getUser()->userName;
```
dasgarner, line 1701: I think this wants to be userId not userName (which can change)

lib/Entity/Schedule.php
```diff
@@ -304,6 +304,30 @@ class Schedule implements \JsonSerializable
      */
     public $syncGroupId;
 
+    /**
+     * @SWG\Property(description="The userName of the user that last modified this Schedule")
```
dasgarner, line 308: userId

db/migrations/20230725141000_schedule_meta_data_migration.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Add meta data fields to schedule table
+ * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+ */
+class ScheduleMetaDataMigration extends AbstractMigration
+{
+    public function change(): void
+    {
+        $this->table('schedule')
+            ->addColumn('modifiedBy', 'string', ['null' => true, 'default' => null])
```
dasgarner, line 34: `integer` I think? without a reference to the user table though

lib/Factory/ScheduleFactory.php
```diff
@@ -385,7 +385,11 @@ public function query($sortOrder = null, $filterBy = [])
             `schedule`.syncGroupId,
             `daypart`.isAlways,
             `daypart`.isCustom,
-            `syncgroup`.name AS syncGroupName
+            `syncgroup`.name AS syncGroupName,
+            `schedule`.modifiedBy,
```
dasgarner, line 389: You'll probably want to LEFT OUTER JOIN user in again to get modifiedBy

dasgarner (changes requested):
I think we want modifiedBy to be an ID not a name.

dasgarner (approved):

PR #1945, opened 2023-07-27
Player: Add 404 handling when requesting data

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#264


modules/src/player.js
```diff
@@ -160,6 +160,21 @@ $(function() {
...
         }).fail(function(jqXHR, textStatus, errorThrown) {
+          if (
+              typeof window[`onDataError_${widget.widgetId}`] === 'function'
+          ) {
+            const onDataError = window[
+              `onDataError_${widget.widgetId}`
+            ](jqXHR.status, jqXHR.responseJSON);
+
+            if (onDataError == false) {
+              xiboIC.reportFault({
```
dasgarner, line 171: I think you want to do this bit if onDataError is false, but also if it is not defined at all (i.e. this is the default behaviour if the request for data fails)

rubenpingol-xibo: Right. Done updating the PR.

dasgarner (approved):
Looks good to me!

PR #1948, opened 2023-07-28
Fixes/issues from 3.3.7

PeterMis:
In addition to 3.3.7 issues
- Player Software : validate fileName on save to avoid duplicates
- Schedule Grid : Fix command type event end date
- Members forms : Fix sorting by member column in all of them (in v3 this only affects Display Group members form)
- Folders : Fixed Root Folder title

### DIFF
--- a/lib/Controller/PlayerSoftware.php
+++ b/lib/Controller/PlayerSoftware.php
@@ -608,7 +608,7 @@ class PlayerSoftware extends Base
             // Add the Player Software record
             $playerSoftware = $this->getPlayerVersionFactory()->createEmpty();
             $playerSoftware->modifiedBy = $this->getUser()->userName;
-            $playerSoftware->fileName = $file->fileName;
+            $playerSoftware->fileName = urlencode($file->fileName);
             $playerSoftware->size = filesize($filePath);
             $playerSoftware->md5 = md5_file($filePath);
             $playerSoftware->decorateRecord();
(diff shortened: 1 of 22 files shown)


dasgarner (approved):

lib/Controller/PlayerSoftware.php
```diff
@@ -608,7 +608,7 @@ public function add(Request $request, Response $response)
             // Add the Player Software record
             $playerSoftware = $this->getPlayerVersionFactory()->createEmpty();
             $playerSoftware->modifiedBy = $this->getUser()->userName;
-            $playerSoftware->fileName = $file->fileName;
+            $playerSoftware->fileName = urlencode($file->fileName);
```
dasgarner, line 611: I think it might be better to replace spaces with `_` as I don't know whether SOC will support having % in filenames any better than it does spaces?

Perhaps it would be better to replace any non Alpha character with an _ for belt and braces?

We could also add a comment for why we're doing it?

PeterMis: We use the `_` in standard release names which then is parsed to "guess" version and revision etc, although that likely would not be correct with odd fileNames anyway 🤔 

Perhaps we could go with something like
```
preg_replace('/[^a-zA-Z0-9_.]+/', '-', $file->fileName);
```
so alphanum characters `_` and `.` are left alone and everything else replaced with `-` ? 🤔

Alternatively replace with `_`, but that could potentially cause the version/code to be set incorrectly 
(if say you add a word and space before standard release fileName for some reason)

dasgarner: Re-reading the comment from SOC team, I think using `-` will be fine!

PeterMis: Ok, the change in the commit below should work fine then 🚀

PR #1949, opened 2023-07-31
Release Prep - 3.3.7

dasgarner:
xibosignageltd/xibo-private#372


lib/Controller/Folder.php
```diff
@@ -2,7 +2,7 @@
 /*
  * Copyright (C) 2023 Xibo Signage Ltd
  *
- * Xibo - Digital Signage - https://xibosignage.com
+ * Xibo - Digital Signage - http://www.xibo.org.uk
```
PeterMis, line 5: I'm not sure if we want to change this 🤔 

Given the copyright update in here and in Handlers.php , perhaps it's not quite correctly configured?

PeterMis (dismissed):
Looks good!

Minor comment below about odd thing with copyright in php files

PeterMis (approved):

PR #1954, opened 2023-08-01
XMDS: Dependencies should come via HTTP if possible + some other minor updates

dasgarner:
Other minor fixes:

HLS widget upgrade
Primary key on lkdgdg
Fixes from 3.3.7
Naming things


PeterMis (approved):
🚀

PR #1956, opened 2023-08-02
Widget: webpage modeid should be non camel case

dasgarner:
In v3, although the form has `modeId`, the `setOption` is actually `modeId`, which means that all players have been looking for `modeid`.

fixes https://github.com/xibosignageltd/xibo-private/issues/382


maurofmferrao (approved):
Looks good to me!

PR #1958, opened 2023-08-03
Widgets Preview: player.js refactor with paged elements grouping and effects

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#365


modules/src/player.js
```diff
@@ -469,159 +508,329 @@ $(function() {
...
+                              parentId: groupContentID,
+                              id: groupContentID,
+                            },
+                            $groupContentItem.find(`.${groupContentID}--item`),
+                          );
+                        });
+                      }
+                    }
+                  });
+                  // $.each(dataItems, function(_key, item) {
```
maurofmferrao, line 762: Should we keep this commented code?

views/common.twig
```diff
@@ -525,20 +525,8 @@
                     save: "{{ "Save" |trans }}",
                     cancel: "{{ "Cancel" |trans }}",
                 },
-                positionTab: {
```
maurofmferrao, line 528: There was a merge conflict here, we need to keep the `positionTab` properties

maurofmferrao (changes requested):
All looks good to me, except 2 minor changes

modules/src/player.js
rubenpingol-xibo: My bad. I'll have this removed.

views/common.twig
rubenpingol-xibo: Strange, I checked the copy of westphal and it doesn't have the `positionTab` in it. I'll have this added back.

maurofmferrao (approved):
Looks good!

PR #1966, opened 2023-08-07
Cypress Test - Schedule

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/341


dasgarner:
Before we merge this and it gets into a release, we need to find a way to remove access to these special "test only" routes in the main release files.

I think the way we do that is have them in a separate file called `routes-cypress.php` which we include from routes.php only if it exists.

We can then delete the routes-cypress.php from inside the production Dockerfile, meaning they won't be available in the main build.

This PR can wait until after rc1 on that basis.

dasgarner:
> We can then delete the routes-cypress.php from inside the production Dockerfile, meaning they won't be available in the main build.

I don't see this in the updated PR - did it get done?

ifarzana:
> I don't see this in the updated PR - did it get done?

my mistake, that's added now.

PR #1973, opened 2023-08-08
XMDS font fixes & duration/numItems

dasgarner:
Fixed:
 - downloads for fonts via XMDS/CDN
 - property validation of the default value, and conditions which use the value in the condition for the test
 - numItems for dataset/calendar xibosignageltd/xibo-private#394


maurofmferrao (approved):
Looks good to me!

PR #1979, opened 2023-08-10
Dataset: Elements

rubenpingol-xibo:
relates to xibosignage/xibo#2747


dasgarner (approved):
The only thing I am not clear on is when this message is shown

> No field is available for that type of DataSet element.

If that is when they haven't selected a field yet, should we change it to 

> Please select a field for this element

rubenpingol-xibo:
This message is shown when there is no available field corresponding to the type of element added to the editor.

rubenpingol-xibo:
Basically, field here are columns that are available from the selected dataset. So, when a String element has been added to the editor, it checks from the columns if there is available String data type column, if not then show this message

dasgarner (approved):

PR #1981, opened 2023-08-11
Widgets: HTML sanitisation

dasgarner:
- DataSet correctly handle saving HTML in a HTML column type (sanitised for body)
 - RSS Ticker correctly handle strip tags (previously dropped content)


ifarzana (approved):
Looks good to me.

PR #1988, opened 2023-08-16
Mastodon username search

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/351

### DIFF
--- a/lib/Widget/MastodonProvider.php
+++ b/lib/Widget/MastodonProvider.php
@@ -46,21 +46,22 @@ class MastodonProvider implements WidgetProviderInterface
 
         try {
             $httpOptions = [
-               // 'timeout' => 20, // wait no more than 20 seconds
-                'query' => [
-                    'limit' => $dataProvider->getProperty('numItems', 15)
-                ]
+                'timeout' => 20, // wait no more than 20 seconds
+            ];
+
+            $queryOptions = [
+                'limit' => $dataProvider->getProperty('numItems', 15)
             ];
 
             if ($dataProvider->getProperty('searchOn', 'all') === 'local') {
-                $httpOptions['query']['local'] = true;
+                $queryOptions['local'] = true;
             } elseif ($dataProvider->getProperty('searchOn', 'all') === 'remote') {
-                $httpOptions['query']['remote'] = true;
+                $queryOptions['remote'] = true;
             }
 
             // Media Only
             if ($dataProvider->getProperty('onlyMedia', 0)) {
-                $httpOptions['query']['only_media'] = true;
+                $queryOptions['only_media'] = true;
             }
 
             if (!empty($dataProvider->getProperty('serverUrl', ''))) {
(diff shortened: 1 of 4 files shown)


lib/Widget/MastodonProvider.php
```diff
@@ -62,24 +64,35 @@ public function fetchData(DataProviderInterface $dataProvider): WidgetProviderIn
...
             $hashtag = trim($dataProvider->getProperty('hashtag', ''));
-            if (!empty($hashtag)) {
-                $uri = rtrim($uri, '/').'/api/v1/timelines/tag/'. trim($hashtag, '#');
+
+            // when username is provided do not search in public timeline
+            if (!empty($dataProvider->getProperty('userName', ''))) {
+                // username search: get account ID, always returns one record
+                $accountId = $this->getAccountId($uri, $dataProvider);
+                $httpOptions['query']['tagged'] = trim($hashtag, '#');
+                ;
```
dasgarner, line 78: ?

lib/Widget/MastodonProvider.php
```diff
@@ -157,4 +170,45 @@ public function getDataModifiedDt(DataProviderInterface $dataProvider): ?Carbon
     {
         return null;
     }
+
+    /**
+     * Get Mastodon Account Id from username
+     * @throws GuzzleException
+     */
+    private function getAccountId(mixed $uri, DataProviderInterface $dataProvider)
```
dasgarner, line 178: Can we make this function declaration clearer? isn't it `string $uri, string $userName` ?

lib/Widget/MastodonProvider.php
```diff
@@ -157,4 +170,45 @@ public function getDataModifiedDt(DataProviderInterface $dataProvider): ?Carbon
...
+        $this->getLog()->debug('Mastodon: getAccountId: ID ' . $result['id']);
+
+        return $result['id'];
+    }
+
+    /**
+     * Get Mastodon account public status
+     * @throws GuzzleException
+     */
+    private function getResult(mixed $uri, $accountId, $httpOptions)
```
dasgarner, line 203: Is this really `getResult` ?

It seems like this is only used for account search?

Is $uri really mixed?

lib/Widget/MastodonProvider.php
```diff
@@ -157,4 +170,45 @@ public function getDataModifiedDt(DataProviderInterface $dataProvider): ?Carbon
...
+    {
+        $uri = rtrim($uri, '/') . '/api/v1/accounts/' . $accountId . '/statuses?';
+
+        $client = new Client();
```
dasgarner, line 207: We should use `getGuzzleClient` because it comes configured with any global options (such as proxy)

dasgarner (changes requested):
Broadly looks good. I have some minor refactor requests.

dasgarner (approved):

PR #1991, opened 2023-08-17
Misc fixes

dasgarner:
* Phinx deprecation warning
* Should not be able to set expiry dates when uploading to the library
* Media manager dashboard


dasgarner:
phpcs is for unchanged lines, won't fix.

PeterMis (approved):
Looks good

PR #1996, opened 2023-08-18
Misc fixes 5

dasgarner:
A novel way to solve the issues returning data to the calendar between elements vs static. Also adding the language option.


maurofmferrao (approved):
Looks good to me!

PR #1998, opened 2023-08-21
Misc fixes for Elements/PlayerJS

dasgarner:
- Date format appears twice
 - Missing element for screen name
 - onParseData not run in viewer
 - extends broken in viewer
 - onTemplateRender not run for elements in player

xibosignageltd/xibo-private#419
xibosignageltd/xibo-private#420


maurofmferrao (approved):
Looks good to me!

PR #2004, opened 2023-08-22
Misc fixes 7

dasgarner:
Video: switch to using $ notation for client side variables in visibility. xibosignageltd/xibo-private#424
About: attempt to fix SHA xibosignageltd/xibo-private#425
Chore: renamed region timeline form to make it clear it's still used xibosignageltd/xibo-private#428


dasgarner:
phpcs on unrelated changes - ignoring.

maurofmferrao (approved):

PR #2011, opened 2023-08-25
Misc fixes 10

dasgarner:
Peter:
 * Widget cache: too many files were building up, so they are cleared daily
 * Help links: adjustments after testing
 * Regions to have an empty name by default
 * Broken connectors (installed but no class) shouldn't show

Mauro:
 * Uploader shouldn't show view library
 * Modules/Module Templates can have an icon
 * Swap to use fa classes directly
 * Convert all elements to use icons


PeterMis (approved):

maurofmferrao (approved):
Looks good to me!

PR #2014, opened 2023-08-29
Player: Get data refactor for static templates and elements

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#435


maurofmferrao (approved):
Looks good to me!

PR #2021, opened 2023-08-30
Required Elements and Weather fixes

dasgarner:
Implement required elements in the BE so that the "attribution" element must be provided when adding Weather. Also fixes for the weather icons not being included.


dasgarner:
phpcs is on unchanged lines, won't fix.

PR #2023, opened 2023-08-31
Alias along with its parameters

ifarzana:
- https://github.com/xibosignageltd/xibo-private/issues/447


maurofmferrao (approved):
All looks good to me!

PR #2024, opened 2023-08-31
Widgets: Fix currencies flags and icon

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#440

### DIFF
--- a/modules/templates/currency-elements.xml
+++ b/modules/templates/currency-elements.xml
@@ -98,7 +98,7 @@ if (elementData !== null && elementData.hasOwnProperty('ChangeStyle')) {
         </properties>
         <stencil>
             <hbs><![CDATA[
-<div class="arrow currency-elements-change-icon">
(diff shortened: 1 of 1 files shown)


dasgarner:
I think we may have already fixed this issue in this PR: https://github.com/xibosignage/xibo-cms/pull/2021

We had a similar issue with weather icons where the `autoInclude` property wasn't working as expected.

dasgarner:
See flags working without this PR already:

image

rubenpingol-xibo:
Yes, as I have updated my branch on this PR and tested it, I can confirm that the changes in this PR are not necessary anymore.

> I think we may have already fixed this issue in this PR: #2021
> 
> We had a similar issue with weather icons where the `autoInclude` property wasn't working as expected.

dasgarner:
Thanks Ruben, and sorry for the wasted effort. I'll close this out then.

PR #2029, opened 2023-08-31
Widgets and Player: isRepeatData and fixes to rendering, animation duration, and non-data elements

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#371 xibosignageltd/xibo-private#435 xibosignageltd/xibo-private#365

### DIFF
--- a/lib/Controller/Widget.php
+++ b/lib/Controller/Widget.php
@@ -349,7 +349,7 @@ class Widget extends Base
             'commonProperties' => [
                 'name' => $widget->getOptionValue('name', null),
                 'enableStat' => $widget->getOptionValue('enableStat', null),
-                'isRepeatData' => $widget->getOptionValue('isRepeatData', null),
+                'isRepeatData' => $widget->getOptionValue('isRepeatData', true),
                 'duration' => $widget->duration,
                 'useDuration' => $widget->useDuration
             ],
--- a/lib/Controller/Widget.php
+++ b/lib/Controller/Widget.php
@@ -349,7 +349,7 @@ class Widget extends Base
             'commonProperties' => [
                 'name' => $widget->getOptionValue('name', null),
                 'enableStat' => $widget->getOptionValue('enableStat', null),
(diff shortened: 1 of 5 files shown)


lib/Entity/Widget.php
```diff
@@ -213,6 +213,12 @@ class Widget implements \JsonSerializable
      */
     public $isValid = false;
 
+    /**
+     * Flag to indicate whether the widget data has to be repeated in the player
+     * @var bool
+     */
+    public $isRepeatData = true;
```
dasgarner, line 220: I think we are saving this in the widget options, so we shouldn't also need it here.

lib/Controller/Widget.php
```diff
@@ -349,7 +349,7 @@ public function editWidgetForm(Request $request, Response $response, $id)
             'commonProperties' => [
                 'name' => $widget->getOptionValue('name', null),
                 'enableStat' => $widget->getOptionValue('enableStat', null),
-                'isRepeatData' => $widget->getOptionValue('isRepeatData', null),
+                'isRepeatData' => $widget->getOptionValue('isRepeatData', true),
```
dasgarner, line 352: This was null before because it is not always required (we don't use it on widgets without data for example)

lib/Widget/Render/WidgetHtmlRenderer.php
```diff
@@ -458,6 +458,9 @@ private function render(
             // templateId or "elements"
             $templateId = $widget->getOptionValue('templateId', null);
 
+            // isRepeatData
+            $isRepeatData = $widget->getOptionValue('isRepeatData', true);
```
dasgarner, line 462: This makes sense, but you can do it inline below, I think?

```
'isRepeatData' => $widget->getOptionValue('isRepeatData', 1) === 1,
```

dasgarner (changes requested):

maurofmferrao (approved):

dasgarner (approved):

PR #2065, opened 2023-09-08
Player: Add xiboIC.expireNow for elements and call xiboIC.reportFault

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#264


dasgarner (approved):
I think this looks good!

PR #2068, opened 2023-09-08
Widget Fixes

dasgarner:
- Countdown widget warning on layout status (can't use duration in status rule)
 - Dashboards should have a shorter token TTL
 - RSS ticker should support custom image enclosures
 - RSS ticker upgrade from v3 fixed
 - Welcome page uses product_support_url for non Xibo themed, which isn't set.


PeterMis (approved):
Looks good 🚀

PR #2076, opened 2023-09-11
Widgets: Enable marquee for widget elements

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#438

### DIFF
--- a/ui/src/core/forms.js
+++ b/ui/src/core/forms.js
@@ -1878,6 +1878,9 @@ window.forms = {
 
       // Add the options
       $.each(effects, function(_index, element) {
+        if (effectsType !== 'all' && element.group !== effectsType) {
+          return;
+        }
         $el.append(
           $('<option value="' +
             element.effect +
--- a/ui/src/core/forms.js
+++ b/ui/src/core/forms.js
@@ -1878,9 +1878,6 @@ window.forms = {
 
       // Add the options
       $.each(effects, function(_index, element) {
-        // if (effectsType !== 'all' && element.group !== effectsType) {
-        //   return;
-        // }
         $el.append(
           $('<option value="' +
             element.effect +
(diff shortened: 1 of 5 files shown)


ui/src/core/forms.js
```diff
@@ -1878,9 +1878,6 @@ window.forms = {
 
       // Add the options
       $.each(effects, function(_index, element) {
-        if (effectsType !== 'all' && element.group !== effectsType) {
```
dasgarner, line 1881: I thought this condition was here so that the `variant` of any effects property filtered the effects returned.

rubenpingol-xibo: Yes, this condition should filter the effects based on the `variant` but having this line prevents us from showing marquee effects as previously we were setting the effects field manually for elements.

`properties-panel.js`
image

rubenpingol-xibo: Should we change it's `variant = all`?

dasgarner: Sorry I don't understand - does variant filtering still work without this condition (via some other mechanism)?

rubenpingol-xibo: No, it won't work anymore without this condition. However, this condition prevents showing marquee effects.

maurofmferrao (approved):

ui/src/core/forms.js
dasgarner: Breaking something else can't be the solution, so perhaps the variant logic needs some refactoring. We have to be able to control which effects are returned for the property - either all, paged only or non-paged only (marquee). Please can you discuss with Mauro?

rubenpingol-xibo: Yes, will do.

maurofmferrao (approved):

dasgarner (approved):

PR #2088, opened 2023-09-13
Various issues with widgets that supply a URL

dasgarner:
XLF: incorrect options output, incorrect renderAs for webpage, incorrect application of filters saving URLs
fixes xibosignage/xibo#3140


maurofmferrao (approved):
Looks good to me!

PR #2091, opened 2023-09-13
Elements: Fix for scroll horizontal and vertical effect

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#478


maurofmferrao (approved):
Looks good to me!

PR #2093, opened 2023-09-14
Widgets: Marquee fixes for marquee up/down

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#438

### DIFF
--- a/modules/src/xibo-elements-render.js
+++ b/modules/src/xibo-elements-render.js
@@ -136,7 +136,7 @@ jQuery.fn.extend({
         ((options.effect === 'marqueeLeft') ? 'left' : 'right');
 
       // Make sure the speed is something sensible
-      options.speed = (options.speed === 0) ? 1 : options.speed;
+      options.speed = (options.speed === 0) ? 1 : options.speed / 25 * 1000;
 
       // Add gap between
       if ($this.find('.scroll').length > 0) {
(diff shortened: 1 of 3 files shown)


modules/src/xibo-elements-render.js
```diff
@@ -136,7 +136,7 @@ jQuery.fn.extend({
         ((options.effect === 'marqueeLeft') ? 'left' : 'right');
 
       // Make sure the speed is something sensible
-      options.speed = (options.speed === 0) ? 1 : options.speed;
+      options.speed = (options.speed === 0) ? 1 : options.speed / 25 * 1000;
```
dasgarner, line 139: What is the reason for dividing by 25, and can be add a comment explaining that?

rubenpingol-xibo: This calculation gives us a result of `80` as a value for `options.speed`. This value is the speed we are already using from `xibo-text-render.js` which is I believe based on the **85 milliseconds** default scrollamount as pixels per second.

maurofmferrao (approved):

modules/src/xibo-elements-render.js
dasgarner: Understood! I think a comment explaining this for "future Ruben" would be awesome, but you can merge this for testing in the meantime.

dasgarner (approved):

modules/src/xibo-elements-render.js
rubenpingol-xibo: Yep, got it @dasgarner. I'll take this into account. Thanks!

maurofmferrao (approved):
Looks good to me!

PR #2094, opened 2023-09-14
Widget issues with data cache

dasgarner:
DataSet takes too long to update on the player after editing
 - add data last modified and active info to data set page for debugging
 - XMDS required files needs to push the `updated` attribute on resource for older players
 - DataSet notify should run the widget sync task immediately
 - Widget Sync task should notify now for older displays and should have some way to notify newer players that data has changed 
- fixes xibosignage/xibo#3145

PDF widget is broken because it expects data, fixes xibosignage/xibo#3147

### DIFF
--- a/modules/dashboard.xml
+++ b/modules/dashboard.xml
@@ -27,7 +27,7 @@
     <class>\Xibo\Widget\DashboardProvider</class>
     <dataCacheKey>%widgetId%_%displayId%</dataCacheKey>
     <type>dashboard</type>
-    <dataType></dataType>
+    <dataType>xibo-dashboard-service</dataType>
     <schemaVersion>1</schemaVersion>
     <assignable>1</assignable>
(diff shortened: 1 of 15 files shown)


modules/dashboard.xml
```diff
@@ -27,7 +27,7 @@
     <class>\Xibo\Widget\DashboardProvider</class>
     <dataCacheKey>%widgetId%_%displayId%</dataCacheKey>
     <type>dashboard</type>
-    <dataType></dataType>
+    <dataType>xibo-dashboard-service</dataType>
```
dasgarner, line 30: Previously we fixed an issue with the dashboard module not getting updated tokens by adding `$dataModule->isWidgetProviderAvailable()` to where we output required files for data. However, that breaks all other modules with a provider that do not have data (we get a required file node which can never be fulfilled).

The correct fix for dashboards is to add a data type.

PR #2098, opened 2023-09-15
Playlist Editor: some items are missing the file name

rubenpingol-xibo:
relates to xibosignage/xibo#3142

### DIFF
--- a/lib/Controller/Widget.php
+++ b/lib/Controller/Widget.php
@@ -348,6 +348,14 @@ class Widget extends Base
             $template->decorateProperties($widget);
         }
 
+        $widgetName = $widget->getOptionValue('name', null);
+
+        if ($media !== null) {
+            $widgetName = $media->name;
+        } else if ($widgetName === null) {
+            $widgetName = $module->name;
+        }
+
         // Pass to view
         $this->getState()->template = '';
         $this->getState()->setData([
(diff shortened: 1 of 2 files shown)


lib/Controller/Widget.php
```diff
@@ -457,13 +465,31 @@ public function editWidget(Request $request, Response $response, $id)
         }
 
         $module = $this->moduleFactory->getByType($widget->type);
+        $widgetName = $widget->getOptionValue('name', $module->name);
```
dasgarner, line 468: If we set this on edit, it means that when the user changes the name of an item in the library, it will not also change on the playlist editor (because the old item will be set here).

So, I don't think we need to do anything on edit?

lib/Controller/Widget.php
```diff
@@ -348,6 +348,14 @@ public function editWidgetForm(Request $request, Response $response, $id)
             $template->decorateProperties($widget);
         }
 
+        $widgetName = $widget->getOptionValue('name', null);
+
+        if ($media !== null) {
```
dasgarner, line 353: We should only override the `$widgetName` if it is empty, right?

dasgarner (changes requested):

lib/Controller/Widget.php
rubenpingol-xibo: @dasgarner The `widgetName` comes out to be still empty if I remove these changes here on edit.

rubenpingol-xibo: Okay, I think I got it now. I'll update the PR

dasgarner:
Fixed in a subsequent PR.

PR #2105, opened 2023-09-19
XMDS: get resource for older players using wrong module cache and data provider

dasgarner:
relates to xibosignage/xibo#3145


dasgarner:
phpcs on unrelated lines - wont fix
cypress flaking

PeterMis (approved):

PR #2110, opened 2023-09-20
Missing API docs

dasgarner:


dasgarner:
phpcs on unchanged code - won't fix

PeterMis (approved):

PR #2114, opened 2023-09-25
Cypress - saving artifacts and misc fixes

ifarzana:
Cypress should save screenshots in artifacts and stop


maurofmferrao (approved):
Looks good to me!

PR #2116, opened 2023-09-26
Show error message on Elements with bad configured widget

maurofmferrao:
relates to xibosignageltd/xibo-private#499

### DIFF
--- a/lib/Widget/IcsProvider.php
+++ b/lib/Widget/IcsProvider.php
@@ -51,7 +51,7 @@ class IcsProvider implements WidgetProviderInterface
         // Do we have a feed configured?
         $uri = $dataProvider->getProperty('uri');
         if (empty($uri)) {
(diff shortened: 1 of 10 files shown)


lib/Widget/IcsProvider.php
```diff
@@ -51,7 +51,7 @@ public function fetchData(DataProviderInterface $dataProvider): WidgetProviderIn
         // Do we have a feed configured?
         $uri = $dataProvider->getProperty('uri');
         if (empty($uri)) {
-            throw new InvalidArgumentException('Please enter a the URI to a valid ICS feed.', 'uri');
+            throw new InvalidArgumentException('Please enter the URI to a valid ICS feed.', 'uri');
```
dasgarner, line 54: This should be translated (although it wasn't before)

dasgarner (changes requested):

dasgarner (approved):

PR #2119, opened 2023-09-29
Widget elements: Mastodon shows username instead of post text and date is not formatted

rubenpingol-xibo:
relates to xibosignage/xibo#3138


maurofmferrao (approved):
Looks good to me!

PR #2122, opened 2023-10-02
Cypress - add/edit

ifarzana:
- Related to https://github.com/xibosignageltd/xibo-private/issues/500


maurofmferrao (commented):
Just a minor comment about the name of a single test, everything else looks good!

dasgarner:
Also a note from me to say that phpcs is on new lines, so we ought to solve that before merge. Thanks.

maurofmferrao (approved):
Looks good to me!

PR #2124, opened 2023-10-02
Bugfixes 4.0.4 issues

PeterMis:
- Video Upload : no thumbnail without manually pausing/seeking in the video preview. Relates to https://github.com/xibosignage/xibo/issues/3153
- DataSets : Fix formula validation for `[DisplayGeoLocation]` replacement, Fix sql function for geometry, Fix an issue with mysql 8 where we had more parameters than needed in the query, Add displayId to the cacheKey, make sure we have both lower and upper limit in the cacheKey (instead of lower limit twice). Relates to https://github.com/xibosignage/xibo/issues/3144
- Stocks : Fix AlphaVantage api endpoints, they have changed them around :) . Relates to https://github.com/xibosignage/xibo/issues/3168
- Status Dashboard : Fix sql error when ordering by Faults column in Display grid. Relates to https://github.com/xibosignage/xibo/issues/3170
- Schedule Calendar : Fix issues with different date formats, relates to https://github.com/xibosignage/xibo/issues/3149
- Schedule : Fix Sync event form layouts dropdown not returning results when CMS is installed in a sub folder. Relates to https://github.com/xibosignage/xibo/issues/3158
- Folders : Fix Folders page not displaying Folder details. Relates to https://github.com/xibosignage/xibo/issues/3171
- Fonts : Run regex to fix issues with spaces and special characters in the fileName on upload. Relates to https://github.com/xibosignage/xibo/issues/3152  (This may need some other adjustments down the line for existing files.)

### DIFF
--- a/lib/Controller/Folder.php
+++ b/lib/Controller/Folder.php
@@ -95,7 +95,12 @@ class Folder extends Base
             $this->folderFactory->decorateWithSharing($folder);
             $this->folderFactory->decorateWithUsage($folder);
 
-            return $response->withJson($folder);
+            return $response->withJson([
+                'folder' => $folder,
+                'homeFolderCount' => $folder->getUnmatchedProperty('homeFolderCount'),
+                'sharing' => $folder->getUnmatchedProperty('sharing'),
+                'usage' => $folder->getUnmatchedProperty('usage')
+            ]);
         } else {
             // Show a tree view of all folders.
             $rootFolder = $this->folderFactory->getById(1);
--- a/lib/Controller/Folder.php
+++ b/lib/Controller/Folder.php
@@ -95,12 +95,7 @@ class Folder extends Base
             $this->folderFactory->decorateWithSharing($folder);
             $this->folderFactory->decorateWithUsage($folder);
 
-            return $response->withJson([
-                'folder' => $folder,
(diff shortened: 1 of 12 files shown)


PeterMis:
PHPCS on unchanged lines.

lib/Controller/Folder.php
```diff
@@ -95,7 +95,12 @@ public function grid(Request $request, Response $response, $folderId = null)
             $this->folderFactory->decorateWithSharing($folder);
             $this->folderFactory->decorateWithUsage($folder);
 
-            return $response->withJson($folder);
+            return $response->withJson([
```
dasgarner, line 98: I don't think this, and the twig changes, should be necessary because unmatched properties should be output when the folder object is JSON serialised

dasgarner (changes requested):
Please can you have another look at the folders page - i'd expect the unmatched properties to be output as part of the object?

lib/Controller/Folder.php
PeterMis: Good catch! the issue was there because Folder Entity was not implementing JsonSerializable, therefore the unmatched properties were not on the Folder object - fixed in the commit below

dasgarner (approved):

PR #2128, opened 2023-10-03
Widget fixes for Dashboard and DataSet

dasgarner:
Several widget fixes/adjustments:

 - Add a dashboard template so that the dashboard widget can be added (now that it has a datatype) xibosignage/xibo#3176
 - Add the no data message to DataSet templates that need it xibosignage/xibo#3177
 - Add a "string" template for DataSet when used inside a playlist
 - Move the "slideshow" DataSet template inside a playlist
 - Fixes for showing an icon and for showIn playlist for templates


maurofmferrao (approved):
Looks good!

PR #2129, opened 2023-10-03
Bugifxes 4.0.4 issues part2

PeterMis:
- Schedule : Sync event, check and throw relevant error if not all Displays were assigned a Layout. Relates to https://github.com/xibosignage/xibo/issues/3150
- Maintenance : Remove no longer needed asset (weather-humidity-percent) from weather template. Relates to https://github.com/xibosignage/xibo/issues/3155
- Custom Install : Try to create fonts sub-folder under selected Library Directory on install. Relates to https://github.com/xibosignage/xibo/issues/3156
- Layout : Filter Resolutions available in Layout editor Resolution dropdown by enabled flag, but always include currently selected resolution regardless of enabled flag. Relates to https://github.com/xibosignage/xibo/issues/3173
- Library Upload : Protect against empty Default Resize Limit causing various errors on upload form. Relates to https://github.com/xibosignage/xibo/issues/3159

### DIFF
--- a/lib/Factory/ResolutionFactory.php
+++ b/lib/Factory/ResolutionFactory.php
@@ -168,16 +168,19 @@ class ResolutionFactory extends BaseFactory
            WHERE 1 = 1
         ';
 
-        if ($parsedFilter->getInt('enabled', ['default' => -1]) != -1) {
+        if ($parsedFilter->getInt('enabled', ['default' => -1]) != -1
+            && $parsedFilter->getInt('withCurrent') !== null
+        ) {
+            $body .= ' AND ( enabled = :enabled OR `resolution`.resolutionId = :withCurrent) ';
+            $params['enabled'] = $parsedFilter->getInt('enabled');
+            $params['withCurrent'] = $parsedFilter->getInt('withCurrent');
+        }
+
+        if ($parsedFilter->getInt('enabled', ['default' => -1]) != -1
+            && $parsedFilter->getInt('withCurrent') === null
+        ) {
             $body .= ' AND enabled = :enabled ';
             $params['enabled'] = $parsedFilter->getInt('enabled');
-
-            // for Layout Background form, always return Layout current resolution
-            // even if it is disabled
-            if ($parsedFilter->getInt('withCurrent') !== null) {
-                $body .= ' OR `resolution`.resolutionId = :withCurrent ';
-                $params['withCurrent'] = $parsedFilter->getInt('withCurrent');
-            }
         }
 
         if ($parsedFilter->getInt('resolutionId') !== null) {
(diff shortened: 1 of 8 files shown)


PeterMis:
PHPCS on unchanged lines 🤷

lib/Factory/ResolutionFactory.php
```diff
@@ -225,6 +225,13 @@ public function query($sortOrder = null, $filterBy = [])
                 $body .= ' AND intended_width > intended_height ';
             }
         }
+        
+        // for Layout Background form, always return Layout current resolution
+        // even if it is disabled
+        if ($parsedFilter->getInt('withCurrent') !== null) {
```
dasgarner, line 231: I think this might be problematic as it will negate and other clause filtered. Perhaps it would be better if this filter just disabled the `enabled` one?

lib/Helper/Install.php
```diff
@@ -414,6 +414,11 @@ public function step7(Request $request, Response $response) : Response
             $library_location = $library_location . '/';
         }
 
+        // Attempt to create fonts sub-folder in Library location
+        if (!mkdir($library_location . 'fonts', 0777, true)) {
```
dasgarner, line 418: This is a great addition here, but I think maybe `ensureLibraryExists()` should have it too?

dasgarner (changes requested):

lib/Helper/Install.php
dasgarner: OK cool, as you are then!

lib/Factory/ResolutionFactory.php
PeterMis: The only place withCurrent is used is for the layout background form, which only uses that and enabled flag (now anyway).

In general though, it makes sense to move it inside the enabled condition to avoid potential issues.

lib/Factory/ResolutionFactory.php
```diff
@@ -171,6 +171,13 @@ public function query($sortOrder = null, $filterBy = [])
         if ($parsedFilter->getInt('enabled', ['default' => -1]) != -1) {
             $body .= ' AND enabled = :enabled ';
             $params['enabled'] = $parsedFilter->getInt('enabled');
+
```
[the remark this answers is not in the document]
PeterMis: The layout background form that runs this query, always has `enabled = 1` and `withCurrent = <currentResolutionId>`
If I were to do what you suggested, it would always return only current resolution in that dropdown and nothing else.

Anyway, if we want to be strict about it, it needs to be a bit different, I will change it again.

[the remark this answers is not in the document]
dasgarner: So do we just need some brackets? `( AND enabled = :enabled OR `resolution`.resolutionId = :withCurrent )`

dasgarner (approved):

PR #2132, opened 2023-10-04
Widget: Weather background image choose from library

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#485


maurofmferrao (approved):
Looks good to me!

PR #2138, opened 2023-10-09
Refactor player.js

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#505

### DIFF
--- a/modules/src/player-bundle.js
+++ b/modules/src/player-bundle.js
@@ -40,6 +40,9 @@ import './xibo-webpage-render';
 import './xibo-worldclock-render';
 import './xibo-elements-render';
 import './editor-render';
+
+// Import PlayerHelper
+window.PlayerHelper = require('../../ui/src/helpers/player-helper.js');
 import './player';
 
 window.jQuery = window.$ = require('jquery');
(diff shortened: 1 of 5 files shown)


dasgarner:
Please can you briefly outline (at a high level) what changed you made here and why? Just a couple of bullet points will do. It will help to work out what to review.

rubenpingol-xibo:
The first thing done here is I have extracted most player methods into a separate helper object which is now `PlayerHelper`.
One of the large block of code here on the previous `player.js` is the composition of the `Element` object which produces bunch of loops and duplicate methods with little differences. So, to list it out, here are they.

* We have `PlayerHelper.init()` method where we handle widgets and elements including widgets data. This method prepares data and properties for each widget and it's corresponding elements.
* `PlayerHelper.init()` method also structures the elements to what widget it belongs to
* Abstracted duplicate block of code into `PlayerHelper.composeSlotsData()` that is used to manage paging with slots and `pinnedSlot` for both stencils and single elements
* Other small duplicate methods from `player.js` has been extracted into `PlayerHelper` so it's easy to maintain and manage.

`PlayerHelper` can still be improved but this can be a starting point where we can use this externally to manage player widgets and elements.

dasgarner (approved):

maurofmferrao (approved):
Looks good to me!

PR #2147, opened 2023-10-10
Media Element

dasgarner:
Parsing and output for the media element, related to https://github.com/xibosignageltd/xibo-private/issues/294

Saving elements links media correctly (adding and removing), the HTML cache looks right with the `[[mediaId=x]]` refs and the decorated output looks good in preview with the URL.

However, the override with isn't working.

The default image for global-image is dangerous as it is a Xibo logo - it should be something generic instead.

Maintaining media links isn't working for saving the widget - it will not remove references which have been removed. We need to way to easily clear out only the library references from inside properties.


maurofmferrao (approved):
Looks good to me!

PR #2158, opened 2023-10-17
Effects control: Set default speed based on selected option

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#522


maurofmferrao (approved):
Looks good to me! 👍🏻

PR #2176, opened 2023-10-20
Text flashes on-screen once and then disappears when a weather widget is also present

rubenpingol-xibo:
relates to xibosignage/xibo#3195


maurofmferrao (approved):
Looks good!

PR #2179, opened 2023-10-23
Layout Editor: Module and media names are lost after editing

maurofmferrao:
relates to xibosignageltd/xibo-private#536


dasgarner (approved):
Looks good!

PR #2181, opened 2023-10-24
Schedule grid filter improvements and Folder css adjustments

PeterMis:
- Folders : few css adjustments to the folder tree next to grids on relevant pages
- Schedule : Added two new filters directSchedule and sharedSchedule
-- Direct Schedule when enabled is pretty much they way it works now, when disabled we should also see for example: any events scheduled to the displayGroup selected Display is a member of
-- Shared Schedule should show only events that are scheduled to all selected Displays/Groups ie it requires for example at least two Displays to be selected to show any events that are scheduled to both of them.

### DIFF
--- a/lib/Factory/ScheduleFactory.php
+++ b/lib/Factory/ScheduleFactory.php
@@ -486,13 +486,41 @@ class ScheduleFactory extends BaseFactory
         // End both dates
 
         if ($parsedFilter->getIntArray('displayGroupIds') != null) {
+            // parameterize the selected display/groups and number of selected display/groups
+            $selectedDisplayGroupIds = implode(',', $parsedFilter->getIntArray('displayGroupIds'));
+            $numberOfSelectedDisplayGroups = count($parsedFilter->getIntArray('displayGroupIds'));
+
+            // build date filter for sub-queries for shared schedules
+            $sharedScheduleDateFilter = '';
+            if ($parsedFilter->getInt('futureSchedulesFrom') !== null
+                && $parsedFilter->getInt('futureSchedulesTo') === null
+            ) {
+                // Get schedules that end after this date, or that recur after this date
+                $sharedScheduleDateFilter .= ' AND (IFNULL(`schedule`.toDt, `schedule`.fromDt) >= :futureSchedulesFrom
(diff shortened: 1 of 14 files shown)


lib/Factory/ScheduleFactory.php
```diff
@@ -486,8 +486,91 @@ public function query($sortOrder = null, $filterBy = [])
...
+                $parsedFilter->getInt('directSchedule') === null
+            ) {
+                $body .= ' AND `schedule`.eventId IN (
+                    SELECT `lkscheduledisplaygroup`.eventId FROM `lkscheduledisplaygroup`
+                    WHERE displayGroupId IN (' . implode(',', $parsedFilter->getIntArray('displayGroupIds')) . ')
+                     ) ';
+            } else {
+                // Schedule grid query
+                // check what options we were provided with and adjust query accordingly.
+                if ($parsedFilter->getInt('sharedSchedule') === 1) {
```
dasgarner, line 500: Can we add a couple of comments to each of these conditions to describe what the SQL is supposed to do?

lib/Factory/ScheduleFactory.php
```diff
@@ -486,8 +486,91 @@ public function query($sortOrder = null, $filterBy = [])
...
+                            SELECT lkdgdg.childId FROM lkdgdg WHERE lkdgdg.parentId IN (' .
+                                implode(',', $parsedFilter->getIntArray('displayGroupIds')) .') 
+                                AND lkdgdg.depth > 0)
+                            GROUP BY eventId
+                            HAVING COUNT(DISTINCT `lkscheduledisplaygroup`.displayGroupId) = ' .
+                            count($parsedFilter->getIntArray('displayGroupIds')) . '    
+                        )
+                     ) ';
+                    }
+                } else {
```
dasgarner, line 542: I think this kind of wrapped `else if` is confusing to read?

dasgarner (changes requested):

lib/Factory/ScheduleFactory.php
PeterMis: I've changed the structure ever so slightly ;) and the sql a little bit as well.

Comments for each section as there now too, which should help, thanks

dasgarner (approved):

PR #2182, opened 2023-10-25
PlayerJS: Refactor to render module templates

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#505

### DIFF
--- a/modules/src/player.js
+++ b/modules/src/player.js
@@ -760,7 +760,13 @@ $(function() {
 
           const $content = $('#content');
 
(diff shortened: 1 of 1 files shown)


maurofmferrao (approved):
Looks good to me!

rubenpingol-xibo:
I have not refactored `initStaticTemplates` anymore by separating the ones for module as they almost have the same code to use. The only difference is just the `$template` to use when it's a module.

PR #2187, opened 2023-10-27
Weather widget: changing the language doesn't affect the whole template

rubenpingol-xibo:
relates to xibosignage/xibo#3130

### DIFF
--- a/lib/Helper/Translate.php
+++ b/lib/Helper/Translate.php
@@ -124,9 +124,9 @@ class Translate
      * Get translations for user selected language
      * @param $config
      * @param $language
-     * @return Translator
+     * @return Translator|null
      */
-    public static function getTranslationsFromLocale($config, $language = NULL): Translator
+    public static function getTranslationsFromLocale($config, $language): ?Translator
     {
         // The default language
         $default = ($language === null) ? $config->getSetting('DEFAULT_LANGUAGE') : $language;
@@ -140,25 +140,20 @@ class Translate
 
         // Try to get the local firstly from _REQUEST (post then get)
         if ($language != null) {
-            // Serve only the requested language
-            // Firstly, Sanitize it
-            self::$requestedLanguage = str_replace('-', '_', $language);
+            $parsedLanguage = str_replace('-', '_', $language);
 
             // Check its valid
-            if (in_array(self::$requestedLanguage . '.mo', $supportedLanguages)) {
-                $foundLanguage = self::$requestedLanguage;
+            if (in_array($parsedLanguage . '.mo', $supportedLanguages)) {
+                $foundLanguage = $parsedLanguage;
             }
         }
 
-        // Are we still empty, then default language from settings
+        // Are we still empty, then return null
         if ($foundLanguage == '') {
             // Check the default
             if (!in_array($default . '.mo', $supportedLanguages)) {
-                $default = 'en_GB';
+                return null;
             }
-
-            // The default is valid
-            $foundLanguage = $default;
         }
 
         // Load translations
--- a/lib/Helper/Translate.php
+++ b/lib/Helper/Translate.php
@@ -122,15 +122,11 @@ class Translate
 
     /**
      * Get translations for user selected language
-     * @param $config
      * @param $language
      * @return Translator|null
      */
-    public static function getTranslationsFromLocale($config, $language): ?Translator
+    public static function getTranslationsFromLocale($language): ?Translator
     {
-        // The default language
-        $default = ($language === null) ? $config->getSetting('DEFAULT_LANGUAGE') : $language;
-
         // Build an array of supported languages
         $localeDir = PROJECT_ROOT . '/locale';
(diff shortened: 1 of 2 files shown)


lib/Widget/Render/WidgetHtmlRenderer.php
```diff
@@ -743,7 +752,11 @@ private function decorateTranslations(string $content): string
             $translateTag = str_replace('||', '', $sub);
 
             // We have a valid translateTag to substitute
-            $replace = __($translateTag);
+            if ($translator instanceof Translator) {
```
dasgarner, line 755: I think this can just be a `null` check so that we avoid another `use` import?

lib/Widget/Render/WidgetHtmlRenderer.php
```diff
@@ -468,12 +469,19 @@ private function render(
                 $widget->isValid = 0;
             }
 
+            $properties = $module->getPropertyValues();
+            $translator = null;
+
+            if (array_key_exists('lang', $properties) && !empty($properties['lang'])) {
```
dasgarner, line 475: This relies on the property id being `lang` - could we be cleverer and look through the property object too?

This also means the language property needs to be on the module and not the module template - which might be right, we just need to make sure (perhaps we could add a comment to say what we're doing?)

lib/Helper/Translate.php
```diff
@@ -120,6 +120,55 @@ public static function InitLocale($config, $language = NULL)
...
+        $supportedLanguages = array_map('basename', glob($localeDir . '/*.mo'));
+
+        // Record any matching languages we find.
+        $foundLanguage = null;
+
+        // Try to get the local firstly from _REQUEST (post then get)
+        if ($language != null) {
+            // Serve only the requested language
+            // Firstly, Sanitize it
+            self::$requestedLanguage = str_replace('-', '_', $language);
```
dasgarner, line 145: We need to avoid changing any of the member variables, because they would also be used by the main CMS

lib/Helper/Translate.php
```diff
@@ -120,6 +120,55 @@ public static function InitLocale($config, $language = NULL)
...
+
+            // Check its valid
+            if (in_array(self::$requestedLanguage . '.mo', $supportedLanguages)) {
+                $foundLanguage = self::$requestedLanguage;
+            }
+        }
+
+        // Are we still empty, then default language from settings
+        if ($foundLanguage == '') {
```
dasgarner, line 154: Do we want to set a default? Perhaps it would be better to return NULL if that language isn't supported, so that we can handle it in the calling code (which you have already I think).

lib/Helper/Translate.php
```diff
@@ -120,6 +120,55 @@ public static function InitLocale($config, $language = NULL)
         self::$jsLocaleRequested = str_replace('_', '-', self::$requestedLanguage);
     }
 
+    /**
+     * Get translations for user selected language
+     * @param $config
+     * @param $language
+     * @return Translator
+     */
+    public static function getTranslationsFromLocale($config, $language = NULL): Translator
```
dasgarner, line 129: Why would language ever be null here?

dasgarner (changes requested):
This is going in the right direction, with some small adjustments

lib/Widget/Render/WidgetHtmlRenderer.php
```diff
@@ -468,6 +468,21 @@ private function render(
                 $widget->isValid = 0;
             }
 
+            $moduleLanguage = null;
+            $translator = null;
+
+            // Check if a language property is defined against the module
+            // Note: We are using the language defined against the module and not from the module template
+            foreach ($module->properties as $property) {
+                if ($property->id === 'lang' && !empty($property->value)) {
```
dasgarner, line 477: I still think this should be based on the variant `languageSelector` rather than the property ID, in case someone uses the language selector on a property with a different id.

lib/Helper/Translate.php
```diff
@@ -120,6 +120,50 @@ public static function InitLocale($config, $language = NULL)
         self::$jsLocaleRequested = str_replace('_', '-', self::$requestedLanguage);
     }
 
+    /**
+     * Get translations for user selected language
+     * @param $config
+     * @param $language
+     * @return Translator|null
+     */
+    public static function getTranslationsFromLocale($config, $language): ?Translator
```
dasgarner, line 129: Just to test this with you - this method is only ever going to be called with a provided `$language`, so why do we need the config object?

dasgarner (changes requested):

lib/Widget/Render/WidgetHtmlRenderer.php
rubenpingol-xibo: This makes sense.

lib/Helper/Translate.php
rubenpingol-xibo: Upon checking, this should not be here anymore as we agreed to have no default language set and `return null` when the language is not supported.

dasgarner (approved):

PR #2188, opened 2023-10-31
Cypress workflow change

ifarzana:
- Related https://github.com/xibosignageltd/xibo-private/issues/534

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -1,15 +1,44 @@
 name: Test Suite
 
 on:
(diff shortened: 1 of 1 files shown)


ifarzana:
TEST: displays.cy.js END

dasgarner:
Should we make sure that the workflow is restricted only to PR comments: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only

dasgarner (approved):

PR #2193, opened 2023-11-02
4.0.5 issues pt2

dasgarner:
- RSS date sorting (should be newest first)
 - HLS widget retest: should also send the mute parameter
 - Display page shouldn't have unnecessary requests for the map pins
 - Displays page should show when a player is unsupported (based on the CMS)
 - SAML dynamic property fix


PeterMis (approved):

dasgarner:
Israt confirms that the failing test is unrelated to these changes.
PHPCS is on unchanged lines.

PR #2197, opened 2023-11-06
4.0.5 issues pt4

PeterMis:


lib/Controller/Schedule.php
```diff
@@ -2292,8 +2292,18 @@ public function grid(Request $request, Response $response)
...
-                    $repeatsOn = $weekdays[$event->recurrenceRepeatsOn - 1];
+                    $repeatDays = explode(',', $event->recurrenceRepeatsOn);
+                    $i = 0;
+                    foreach ($repeatDays as $repeatDay) {
+                        // Carbon getDays starts with Sunday,
+                        // return first element from that array if in our array we have 7 (Sunday)
+                        $repeatDay = ($repeatDay == 7) ? 0 : $repeatDay;
+                        $repeatsOn .= $weekdays[$repeatDay];
+                        if ($i < count($repeatDays) - 1) {
+                            $repeatsOn .= ' and ';
```
dasgarner, line 2303: I don't think we can have the English word "and" in here - would a comma or & be better?

dasgarner (changes requested):

lib/Controller/Schedule.php
PeterMis: sure, updated in the commit below.

PR #2207, opened 2023-11-08
Substitutes parser doesn't handle custom date format for variables

rubenpingol-xibo:
relates to xibosignage/xibo#3192


maurofmferrao (approved):
Looks good to me!

PR #2208, opened 2023-11-08
Auth: serialise logic for oAuth objects in session xibosignage/xibo#3184

dasgarner:
The root cause for the "Authorisation parameters missing from session" is that the session isn't being updated in the database for the `GET /api/authorize` request. It seems to be either silently failing, or not updating at all (no log showing the SQL is recorded, but it is for other requests).

I think this might be due to us serialising the private member variables on the application/redirect/scope objects. This PR limits the data we serialise.

This is only a theory.

### DIFF
--- a/lib/Entity/Application.php
+++ b/lib/Entity/Application.php
@@ -1,8 +1,8 @@
 <?php
 /*
- * Copyright (c) 2022 Xibo Signage Ltd
+ * Copyright (C) 2023 Xibo Signage Ltd
  *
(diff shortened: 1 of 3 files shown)


PeterMis (approved):
Certainly worth to test that theory :)

dasgarner:
@ifarzana please can you confirm if this is a false positive?

ifarzana:
Yes, they are false positives.

PR #2214, opened 2023-11-09
Cypress: Workflow change

ifarzana:


ifarzana:
TEST: all

dasgarner (approved):
I have reviewed the source of both the new actions and they look OK. If this works we ought to fix those actions to a particular commit so we always know what we're running.

PR #2216, opened 2023-11-09
Test if workflow change triggers test-suite

ifarzana:


ifarzana:
TEST: all

ifarzana:
TEST: all

dasgarner:
This won't work until it gets into develop I think

ifarzana:
Can you please approve and merge? Thanks

dasgarner:
I think you still need to change the name of the status check that gets updated? Probably to "Build Containers and Run Tests" ?

.github/workflows/test-suite.yaml
```diff
@@ -1,4 +1,4 @@
-name: Test Suite
+name: Build Containers and Run Tests
```
dasgarner, line 1: I think this should be the same and we should change the status update places.

ifarzana: `name: Build Containers and Run Tests`

Is this right?

ifarzana: Ah, I understand now. I don't have permission to edit the branch protection rule.

PR #2222, opened 2023-11-13
Changing data slot shouldn't reload element data

maurofmferrao:
relates to xibosignage/xibo#3233


maurofmferrao:
TEST: layout_designer_empty.cy.js, layout_designer_populated_unchanged.cy.js, layout_designer_populated.cy.js END

dasgarner (approved):

PR #2224, opened 2023-11-14
Elements: Fix preview layout content flashing

rubenpingol-xibo:
relates to xibosignage/xibo#3195

### DIFF
--- a/modules/layout-preview.twig
+++ b/modules/layout-preview.twig
@@ -57,10 +57,9 @@
                 previewTranslations.target = "{{ "Target"|trans }}";
             {% endautoescape %}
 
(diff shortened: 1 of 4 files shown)


maurofmferrao (dismissed):
Looks good to me!

ifarzana:
TEST: /Layout END

dasgarner (changes requested):
Needs the changes we discussed

rubenpingol-xibo:
@dasgarner @maurofmferrao, I think we still need few changes here on `html-preview.js`. The one with adding back `iframe.src = iframe.src` and setting the iframe visibility as a fix for the flashing issue.

PR #2227, opened 2023-11-16
Widgets: Fix scaling issue for static widgets and elements

rubenpingol-xibo:
relates to xibosignage/xibo#3239

### DIFF
--- a/modules/clock-analogue.xml
+++ b/modules/clock-analogue.xml
@@ -207,6 +207,10 @@ if (typeof updateClock === 'function') {
     setInterval(updateClock, 1000);
 }
     ]]></onVisible>
+    <onRender><![CDATA[
+      // Scale the content
+      $(target).xiboLayoutScaler(properties);
+]]></onRender>
     <assets>
         <asset id="clock-analogue-thumb" type="path" cmsOnly="true" mimeType="image/png" path="/modules/assets/template-thumbnails/clock/clock-analogue-thumb.png" />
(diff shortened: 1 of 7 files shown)


rubenpingol-xibo:
TEST: all

modules/clock-analogue.xml
```diff
@@ -207,6 +207,10 @@ if (typeof updateClock === 'function') {
     setInterval(updateClock, 1000);
 }
     ]]></onVisible>
+    <onRender><![CDATA[
+      // Scale the content
+      $(target).xiboLayoutScaler(properties);
```
dasgarner, line 212: Should this be `globalOptions` instead of `properties` ?

rubenpingol-xibo: The `onRender` function receives `properties` as a parameter. Which includes `globalOptions` as defined in `player.js`
image

Should we proceed by just `globalOptions` OR make use of `properties`?

maurofmferrao: I think we should use the same variable as in the other files, and in player.js we are indeed merging `widget.properties` and `globalOptions`.

maurofmferrao (changes requested):
Everything looks good except a minor things about the `scale = 1` 👍🏻

maurofmferrao (approved):
Looks great!

dasgarner (approved):

dasgarner:
@ifarzana are these failures false positives?

PR #2234, opened 2023-11-21
Widgets: Add logic when to show sample data

rubenpingol-xibo:
relates to xibosignage/xibo#3224

### DIFF
--- a/modules/templates/article-static.xml
+++ b/modules/templates/article-static.xml
@@ -102,21 +102,6 @@
 {% if textDirection == "rtl" %}#content { direction: rtl; }{% endif %}
 {% if backgroundColor %}body { background-color: {{backgroundColor}}; }{% endif %}
 {{css}}
-
-.error-message {
-    position: absolute;
-    bottom: 0;
-    left: 0;
-    text-align: center;
-    width: 100%;
-    padding: 12px 0;
-    background-color: #d05454;
-    color: white;
-    z-index: 2;
-    font-weight: bold;
-    font-size: 1.1rem;
-    opacity: 0.85;
-}
 </style>
 {% endif %}
             ]]></twig>
@@ -237,21 +222,6 @@ body {
     transform: translateY(-50%);
 }
 {% endif %}
-
-.error-message {
-    position: absolute;
-    bottom: 0;
-    left: 0;
-    text-align: center;
-    width: 100%;
-    padding: 12px 0;
-    background-color: #d05454;
-    color: white;
-    z-index: 2;
-    font-weight: bold;
-    font-size: 1.1rem;
-    opacity: 0.85;
-}
 </style>
             ]]></twig>
         </stencil>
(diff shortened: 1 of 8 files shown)


ui/src/helpers/player-helper.js
```diff
@@ -3,6 +3,13 @@ const PlayerHelper = function() {
   const _self = this;
   const urlParams = new URLSearchParams(window.location.search);
   const isPreview = urlParams.get('preview') === '1';
+  const staticTemplateIds = [
```
dasgarner, line 6: This sort of hard coding can't be used - lets discuss the problem and see how we can come up with an alternative solution.

dasgarner (changes requested):

rubenpingol-xibo:
TEST: /Layout END

ui/src/helpers/player-helper.js
```diff
@@ -82,6 +83,22 @@ const PlayerHelper = function() {
     }
   });
 
+  /**
+   * Checks if playerjs is used for preview or in editor
+   * @return {boolean}
+   */
+  this.isOnPreview = () => {
```
dasgarner, line 90: Sorry but I am not clear why we've changed this. The new method (using window.parent) won't work in webOS/Tizen because there will be a CORS violation between us and the main application (the window.parent), and in android there is no window.parent because we aren't in any frame at all.

The way we had it previously relied on the URL params, which seems to make sense, but perhaps I have the wrong idea?

rubenpingol-xibo: The reason for changing this is that `isPreview` in these lines always evaluates to **true** even when the iframe is in the editor
```
const urlParams = new URLSearchParams(window.location.search);
const isPreview = urlParams.get('preview') === '1';
```

I have used `window.parent` since the structure of previewing the layout from the editor and on a separate window using the preview link, is the same. So, when we are on the editor and the layout is not being previewed, `isOnPreview()` is evaluated to **false**.

I was not aware that `window.parent` is not working for webOS/Tizen and android.

Maybe there's another way to identify when the layout is being previewed or not.

dasgarner: Ah OK, I think we will need to add an `isEditor=1` parameter to these iframes when they are loaded in the editor - Mauro has an idea for how to do this

rubenpingol-xibo:
TEST: /Layout END

modules/templates/article-static.xml
```diff
@@ -102,6 +102,21 @@
 {% if textDirection == "rtl" %}#content { direction: rtl; }{% endif %}
 {% if backgroundColor %}body { background-color: {{backgroundColor}}; }{% endif %}
 {{css}}
+
+.error-message {
```
maurofmferrao, line 106: Not sure about having the CSS for error-message to be in every template, we should have it on a single file, or maybe even when we're adding it (inline styling).

maurofmferrao (changes requested):
Apart from the CSS for error message, everything looks good!

modules/templates/article-static.xml
rubenpingol-xibo: Yeah, I think this makes sense as I was doubtful having it added for every template just like what we currently have for calendar widget. Let me have this changed.

rubenpingol-xibo:
TEST: /Layout END

dasgarner (approved):

maurofmferrao (approved):

PR #2235, opened 2023-11-22
Misc fixes for duration is per item

dasgarner:
DataSet: fix table durations per page (rowsPerPage)
MenuBoard: fix duration is per item (reverts prior fix).

relates to xibosignageltd/xibo-private#555


maurofmferrao (approved):
Looks good!

dasgarner:
TEST: /Layout END

PR #2240, opened 2023-11-27
Fix duplicated widget when moving

maurofmferrao:
relates to xibosignage/xibo#3244


maurofmferrao:
TEST: /Layout/Editor END

PR #2255, opened 2023-12-06
RSS Ticker: add a static template to easily design a marquee

maurofmferrao:
relates to xibosignage/xibo#3234


maurofmferrao:
TEST: /Layout/Editor END

PR #2256, opened 2023-12-06
RSS Ticker: Title and Description marquee needs to be able to render like in v3

maurofmferrao:
relates to xibosignage/xibo#3256


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2257, opened 2023-12-07
4.0.6 issues pt3

PeterMis:


PeterMis:
TEST: /Administration/applications.cy.js END

views/schedule-page.twig
```diff
@@ -1261,7 +1261,12 @@
           .on('change', function (e) {
             // Refresh the calendar view
             setTimeout(calendar.view(), 1000);
-          });
+          })
+          .on('select2:open', function(event) {
+            setTimeout(function() {
+              $(event.target).data('select2').dropdown?.$search.get(0).focus();
```
dasgarner, line 1267: The chaining operator has good support, but there might be cases where it isn't. It would be better not to use it until we can run this through babel.

PeterMis:
TEST: /Administration/applications.cy.js END

dasgarner (approved):

PR #2263, opened 2023-12-08
Misc issues 4.0.6

dasgarner:
- Hide canvas region from the thumbnail unless it is the only region. Ensure the thumbnail is generated on publish.
 - Format menu board price with 2 decimals on the data entry and widget
 - Fix an error on the user grid when a user has a null homepage
 - XMDS report fault should protect against super long reasons
 - Fix Ad Campaigns add/edit so that impression type can be used

### DIFF
--- a/lib/Controller/Campaign.php
+++ b/lib/Controller/Campaign.php
@@ -527,7 +527,7 @@ class Campaign extends Base
      *  @SWG\Parameter(
      *      name="targetType",
      *      in="formData",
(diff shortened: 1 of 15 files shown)


dasgarner:
TEST: all

dasgarner:
TEST: all END

ifarzana (approved):
Looks good to me. However, we might want to show `Impressions` instead of imp in grid (Target Type)

ifarzana (approved):

dasgarner:

PR #2265, opened 2023-12-11
XMDS: fix data output for multiple canvas widgets for older players

dasgarner:
relates to xibosignageltd/xibo-private#559


ifarzana (approved):
🛠️

PR #2267, opened 2023-12-11
Elements: Pin slot with mix of single elements and element groups

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#565


rubenpingol-xibo:
TEST: /Layout END

maurofmferrao (approved):
Looks good to me!

PR #2274, opened 2023-12-15
Elements: Fix pin slot with mix of elements and and element groups

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#565


rubenpingol-xibo:
TEST: /Layout END

dasgarner (approved):

PR #2276, opened 2023-12-17
Bugfix/v4 stock symbol search error and autosuggestion not working

rubenpingol-xibo:


rubenpingol-xibo:
TEST: /Layout END

dasgarner:
I think you're going to need to take me through this one 😄

lib/Connector/AlphaVantageConnector.php
```diff
@@ -255,16 +255,19 @@ private function getStockResults(DataProviderInterface $dataProvider): void
 
                 $item = [];
 
-                foreach ($result['Time Series (Daily)'] as $series) {
+                $isFromCache = !array_key_exists('Time Series (Daily)', $result);
```
dasgarner, line 258: This doesn't make sense - the `getStockQuote()` should always return data in the same format/structure.

lib/Connector/AlphaVantageConnector.php
```diff
@@ -327,6 +331,10 @@ protected function getStockQuote(string $symbol, ?int $isPaidPlan): array
                 $this->getLogger()->debug('AlphaVantage Connector : getStockQuote is served from the cache.');
             }
 
+            if (is_object($data) && $data->data) {
```
dasgarner, line 334: Again, why are we modifying the cached data - it should always return the same.

modules/templates/stock-static.xml
```diff
@@ -26,6 +26,9 @@
         <showIn>none</showIn>
         <title>Stocks Custom HTML</title>
         <properties>
+            <property id="templateType" type="hidden">
```
dasgarner, line 29: A hidden property which changes the behaviour of a different field doesn't seem right to me.

dasgarner (changes requested):

modules/templates/stock-static.xml
rubenpingol-xibo: At the moment, this is the way I can identify that it's coming from the static template or not.

lib/Connector/AlphaVantageConnector.php
rubenpingol-xibo: Yes, supposedly the cache should always return the same format/structure but apparently in this case, it's not. Upon checking the format/structure that is being saved into the cache is the formatted output already.

```

                foreach ($results as $series) {
                    $item = [
                        'Name' => $name,
                        'Symbol' => $symbol,
                        'time' => $isFromCache ? $series['time'] : $result['Meta Data']['3. Last Refreshed'],
                        'LastTradePriceOnly' => $isFromCache ? $series['LastTradePriceOnly'] : round($series['4. close'], 4),
                        'RawLastTradePriceOnly' => $isFromCache ? $series['RawLastTradePriceOnly'] : $series['4. close'],
                        'YesterdayTradePriceOnly' => $isFromCache ? $series['YesterdayTradePriceOnly'] : round($series['1. open'], 4),
                        'RawYesterdayTradePriceOnly' => $isFromCache ? $series['RawYesterdayTradePriceOnly'] : $series['1. open'],
                        'TimeZone' => $isFromCache ? $series['TimeZone'] : $result['Meta Data']['5. Time Zone'],
                        'Currency' => $currency
                    ];

                    $item['Change'] = round($item['RawLastTradePriceOnly'] - $item['RawYesterdayTradePriceOnly'], 4);
                    $item['SymbolTrimmed'] = explode('.', $item['Symbol'])[0];
                    $item = $this->decorateWithReplacements($item);
                    break;
                }

```

modules/templates/stock-static.xml
dasgarner: Why does that matter?

lib/Connector/AlphaVantageConnector.php
dasgarner: So the question in that case is why?

We are in control of saving to the cache, and returning from the cache, so we ought to be able to make sure they are the same?

modules/templates/stock-static.xml
rubenpingol-xibo: The behavior of the script for autosuggestion won't work for properly for both element and static template. If you see my changes on `xibo-cms.js` there are configuration that works for element but not for static template and vice versa. I can't explain it yet why it does that

lib/Connector/AlphaVantageConnector.php
rubenpingol-xibo: I'll try to clear my local cache for this module and see maybe the one I'm getting is an old cache? But yes, we should have control of what we want to save to the cache.

rubenpingol-xibo:
TEST: /Layout END

maurofmferrao (approved):
Looks good to me!

dasgarner (approved):

PR #2279, opened 2023-12-20
Real time data pt4

dasgarner:
This PR tidies up the user experience for editing data connector scripts, and allows a data connector to proxy its requests via the CMS (because CORS). Players will also proxy.

It also creates a very simple real time data widget which provides to allow a DataSet to be selected.

It also creates a temporary Module Template editor so that a widget can be designed to take advantage of the real time data widget.


dasgarner:
TEST: /Schedule, /Campaign END

PeterMis (approved):
Looks good

PR #2287, opened 2024-01-04
PlayerJS: Refactor with access to widget for each canvas

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#567


rubenpingol-xibo:
TEST: /Layout END

modules/src/player.js
```diff
@@ -19,799 +19,5 @@
...
-              widget.data,
-              widget.showError,
-              widget.errorMessage,
-              widget.data,
-            );
-          }
-        });
-      }
-    });
+  XiboPlayer.init(widgetData, elements);
```
maurofmferrao, line 22: Do we still need this file? Or could we just use `xibo-player.js` to initialize instead?

maurofmferrao (commented):
Everything looks good to me, I just have a small question about the structure :)

modules/src/player.js
rubenpingol-xibo: I've thought about this as well 😅 Should we just add it after the `module.exports` on `xibo-player.js`?

rubenpingol-xibo:
TEST: /Layout END

maurofmferrao (approved):
Looks good to me!

PR #2302, opened 2024-01-11
Mastodon widgets / templates no date formatting

maurofmferrao:
relates to xibosignage/xibo#3281


dasgarner:
Quick question - why does removing the date format help?

maurofmferrao:
> Quick question - why does removing the date format help?

We had extra Date Format in some templates that don't have date in the provider data object ( Currency and Stocks ), should we keep Date Format on those? 🤔

rubenpingol-xibo (approved):

maurofmferrao:
TEST: /Layout/Editor END

mrdanleey:
I installed Xibo 4.0.7 with this fix, but still seeing no date formatting when previewing in the cms. Is this correct? Did not test it on a player yet.

PR #2304, opened 2024-01-12
Canvas region sharing and undo button invalid step removal

maurofmferrao:
- Add sharing for Canvas region
- When we revert/undo a widget save that is invalid ( like reverting Calendar widget to without URL ), remove that step from the history manager to allow more "Undos"


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2310, opened 2024-01-16
Layout Editor: Elements with names

maurofmferrao:
relates to xibosignage/xibo#3304


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2324, opened 2024-01-23
Editor: Publish modal button needs loading status

maurofmferrao:
relates to xibosignage/xibo#3296

### DIFF
--- a/ui/src/layout-editor/layout.js
+++ b/ui/src/layout-editor/layout.js
@@ -468,8 +468,11 @@ Layout.prototype.publish = function() {
       } else {
         toastr.error(res.message);
 
-        // Close dialog
-        bootbox.hideAll();
+        // Remove loading icon from publish dialog
+        $(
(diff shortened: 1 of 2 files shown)


maurofmferrao:
> If publish fails for some reason, would the loading spinner be removed?

I don't think so unfortunately, the problem is that with the current architecture we have there, the button will trigger a portion of code with eval ( which I'm not really a fan ) to another method that doesn't return the request state, so adding the cog works, but there's no way to remove it based on the request result without changing a bunch of things. 

The only way I can think of now is to have a timer to remove the cog, but that doesn't seem right.

Also, I'm not sure if we show a message if the publish fails, for example. I might test that by publishing in a different tab to see what happens. I'll try to improve it. 👍🏻

maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2325, opened 2024-01-23
Misc fixes related to sharing in 4.0

dasgarner:
This PR makes a change to the way we serve images/video thumbnails from the /library/download endpoint, so that if we say we're in preview mode (`preview=1`) we return a 1px placeholder when access is denied.

It also makes a change to the way module files have their permissions checked so that when data is returned users are granted permissions to any linked modules files. That means these files can be checked in the same way as any other file.

It also links together the sharing options applied to a Region and its Region Playlist, which means if we add a playlist (which adds a playlist region), we can set sharing on it.

### DIFF
--- a/lib/Controller/Layout.php
+++ b/lib/Controller/Layout.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2023 Xibo Signage Ltd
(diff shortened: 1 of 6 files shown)


dasgarner:
TEST: /Layout END

PeterMis (approved):

PR #2326, opened 2024-01-24
Schedule grid fixes

PeterMis:
- Fix js error (and PHP deprecated use of explode) when recurring event is set with recurrenceRepeatsOn null.
Relates to https://github.com/xibosignage/xibo/issues/3301
- Fix two issues with switching from Campaign to Interrupt event type - campaign should not be allowed, parentCampaignId should be cleared.
Relates to https://github.com/xibosignage/xibo/issues/3298

### DIFF
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -1776,6 +1776,20 @@ class Schedule extends Base
                     );
                 }
             ]);
+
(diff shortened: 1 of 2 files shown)


PeterMis:
TEST: /Schedule END

dasgarner (approved):

PR #2327, opened 2024-01-24
Widget: v4 calendar missing an option from v3 to "show only current"

maurofmferrao:
relates to xibosignage/xibo#3307

### DIFF
--- a/modules/calendar.xml
+++ b/modules/calendar.xml
@@ -185,43 +185,46 @@
         return false;
     }
 ]]></onDataError>
-    <onParseData><![CDATA[
-// item - element to be parsed
-// properties - object containing the values to be evaluated/used
-
-// If item comes as empty from data, skip parsing
-if (item === null) {
-    return null;
-}
+    <onDataLoad><![CDATA[
+// items: The items to render
+// meta: Metadata
+// properties: The properties for the widget
 
 var now = moment();
-var startDate = moment(item.startDate);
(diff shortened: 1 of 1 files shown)


maurofmferrao:
TEST: /Layout/Editor END

modules/calendar.xml
```diff
@@ -183,6 +188,12 @@
     <onParseData><![CDATA[
 // item - element to be parsed
 // properties - object containing the values to be evaluated/used
+
+// If item comes as empty from data, skip parsing
+if (item === null) {
```
dasgarner, line 193: The flow diagram doesn't have on it a case where onParseData can return null - is this a mistake or is the flow diagram incorrect?

maurofmferrao: Should those values be filtered in `onDataLoad` instead @dasgarner ? 🤔

dasgarner: It makes logical sense to me that if you want to remove data you do it in onDataLoad, and we make it so that onParseData must return an item (and if it doesn't we fail safe and use the original)?

maurofmferrao: Yeah, I think that makes total sense. I already moved that filtering logic to `onDataLoad` and it's working :)

maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2332, opened 2024-01-29
Widget : Add a condition to usage of upsize constraint.

PeterMis:
- Widget : Add a condition to use upsize constraint - compare requested dimensions with CMS resize limit.


PeterMis:
TEST: /Layout END

PR #2333, opened 2024-01-29
Fix for Video In duration getting set to 1 and XLF output tidying

dasgarner:
This PR fixes https://github.com/xibosignage/xibo/issues/3284.

It also tidies up the contents of the XLF where messages were being output for native widgets.


dasgarner:
TEST: /Layout END

PeterMis (approved):

PR #2335, opened 2024-01-29
Widget: clock widget upgrade issue (type changing back to analogue)

dasgarner:
This PR makes sure that when we upgrade a widget we only assert the module type if the widget compatibility class has not already changed it.
fixes xibosignage/xibo#3275


dasgarner:
TEST: /Layout END

ifarzana (approved):

PR #2336, opened 2024-01-29
Editor Layers: Send to back and update other element's layers

maurofmferrao:
- Update other object layers even if target's layer doesn't change
- Don't update other's layers if not needed
- Save all regions with one call


dasgarner (approved):

maurofmferrao:
TEST: /Layout/Editor END

PR #2337, opened 2024-01-30
Playlist Timeline: Owner filter showing empty in the dropdown list

maurofmferrao:
relates to xibosignage/xibo#3302

### DIFF
--- a/lib/Controller/Playlist.php
+++ b/lib/Controller/Playlist.php
@@ -141,6 +141,7 @@ class Playlist extends Base
         // Call to render the template
         $this->getState()->template = 'playlist-page';
         $this->getState()->setData([
+            'users' => $this->userFactory->query(),
             'modules' => $moduleFactory->getAssignableModules()
         ]);
 


maurofmferrao:
TEST: /Layout/Editor END

lib/Controller/Playlist.php
```diff
@@ -141,6 +141,7 @@ public function displayPage(Request $request, Response $response)
         // Call to render the template
         $this->getState()->template = 'playlist-page';
         $this->getState()->setData([
+            'users' => $this->userFactory->query(),
```
dasgarner, line 144: This approach doesn't scale on CMS instances with 100s of users - we need to use select2 instead.

dasgarner (changes requested):

lib/Controller/Playlist.php
maurofmferrao: I'll do that for this case then and also the Layout Editor page where we use the same approach.

PR #2340, opened 2024-01-30
Fix to save target regions on send to back

maurofmferrao:
relates to xibosignage/xibo#3320


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2341, opened 2024-01-31
Analogue/Flip Clock Widgets: layout scaler should not be called - v3 …

maurofmferrao:
…did its own scaling

relates to xibosignageltd/xibo-private#595


maurofmferrao:
TEST: /Layout/Editor END

rubenpingol-xibo (approved):

dasgarner (approved):

PR #2343, opened 2024-02-02
Templates which use properties in HBS should use styles instead

maurofmferrao:
relates to xibosignageltd/xibo-private#523


dasgarner (approved):

maurofmferrao:
TEST: /Layout/Editor END

PR #2348, opened 2024-02-05
Editor owner filter as select2

maurofmferrao:
relates to xibosignage/xibo#3302


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2356, opened 2024-02-07
Remove console.log from xiboIC

maurofmferrao:
relates to xibosignageltd/xibo-private#608

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,40 +1,155 @@
 {
   "name": "xibo-cms",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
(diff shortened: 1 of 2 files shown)


maurofmferrao:
TEST: /Layout/Editor END

dasgarner (changes requested):
Looks like you've got an enormous delta in package.lock again - are you running npm install with the right version of node?

maurofmferrao:
> Looks like you've got an enormous delta in package.lock again - are you running npm install with the right version of node?

I think node's version might be the issue... What version should I be using?

dasgarner:
Node 12

maurofmferrao:
TEST: /Layout/Editor END

dasgarner (approved):

PR #2360, opened 2024-02-08
Currencies and Stocks duration issue

dasgarner:
This PR fixes an issue with currencies and stocks not calculating their duration correctly as they did not know how many items were available.

It also adds a change which will ignore the `[0]` library reference in JavaScript, which seems to be the most common array accessor.


dasgarner:
TEST: /Layout/Editor END

maurofmferrao (approved):
Looks good to me!

PR #2361, opened 2024-02-08
PlayerJS: fix for date rendering on overridden elements.

dasgarner:
This PR fixes an issue with the date element not outputting formatted when used as an overridden element (i.e. data set date).

It also fixes an issue I discovered in testing where I had missed a slot (I had slot 1 and 3) and that caused an exception.


dasgarner:
TEST: /Layout/Editor END

maurofmferrao (approved):
Looks good!

dasgarner:
Superceeded by https://github.com/xibosignage/xibo-cms/pull/2363

PR #2362, opened 2024-02-08
Connectors: Xibo Exchange simple switch to v4 layouts.json

dasgarner:
This PR makes a simple quick fix to the Xibo Exchange endpoint so that we can support v4 only layouts.


dasgarner:
TEST: /Layout/Editor END

PeterMis (approved):

PR #2366, opened 2024-02-09
Resolve handling for onTemplateDataLoad and ensure data items are passed correctly

dasgarner:
PlayerJS: make sure we pass dataItems into both data load methods and hook those up on the back end. 
PlayerJS: Fix v3 players isDataReady.
PlayerJS: call expire now on static widgets where data is not ready
Modules: remove old onDataError.

relates to xibosignageltd/xibo-private#611

### DIFF
--- a/lib/Entity/Module.php
+++ b/lib/Entity/Module.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2023 Xibo Signage Ltd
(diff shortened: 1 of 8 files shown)


dasgarner:
TEST: /Layout END

maurofmferrao (approved):
Looks good to me!

PR #2374, opened 2024-02-13
PlayerJS: Fix undefined repeat data slot and add duration props to groups

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#606


rubenpingol-xibo:
TEST: /Layout END

maurofmferrao (approved):

PR #2376, opened 2024-02-16
Layout Editor: add playlist as a toolbar option rather than a widget

maurofmferrao:
relates to xibosignageltd/xibo-private#631


dasgarner:
This looks good, but before we merge it we need to decide where it goes (i.e. are we ready to make a release40)

PR #2390, opened 2024-02-27
Elements: add an API check to ensure we aren't trying to save a widget with no elements

dasgarner:
related to xibosignageltd/xibo-private#617

I elected to throw an error instead of deleting the widget as we shouldn't make a DELETE inside a POST API call. Hopefully the error will give us an example for where this happens in the FE which we can diagnose.


dasgarner:
TEST: /Layout/Editor END

PR #2394, opened 2024-02-27
Widget: calendar widget event filtering only ever return events 5 days in the past.

dasgarner:
fixes xibosignage/xibo#3350


dasgarner:
TEST: /Layout/Editor END

maurofmferrao (approved):
Looks good to me!

PR #2395, opened 2024-02-28
PlayerJS: Performance and memory leak improvements

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#624


maurofmferrao (dismissed):
Looks good to me!

modules/src/xibo-finance-render.js
```diff
@@ -33,6 +33,8 @@ jQuery.fn.extend({
       previewHeight: 0,
       scaleOverride: 0,
     };
+    const urlParams = new URLSearchParams(window.location.search);
```
dasgarner, line 36: URLSearchParams is not available in older versions of webOS/Tizen.

We this polyfilled in xiboIC, so we should use that to determine whether we're the editor or not?

dasgarner (changes requested):

modules/src/xibo-finance-render.js
rubenpingol-xibo: The thing is, `xiboIC.checkIsPreview()` will always return **true** even when we are in the editor since it first evaluates the `preview` param.

```
checkIsPreview: function() {
      // If we don't have URLSearchParams defined, we're also in preview
      if (typeof(URLSearchParams) === 'undefined') {
        _lib.isPreview = true;
        return true;
      }

      // Check if we have the preview flag in URL
      const searchParams = new URLSearchParams(window.location.search);
      if (searchParams.has('preview') && searchParams.get('preview') == 1) {
        _lib.isPreview = true;
        return true;
      }

      // For the widget preview in viewer
      if (searchParams.has('isEditor') &&
        searchParams.get('isEditor') === '1'
      ) {
        _lib.isPreview = true;
        return true;
      }

      return false;
    }
```

rubenpingol-xibo: Should we update `xiboIC.checkIsPreview()` to first evaluate for `isEditor` param?

rubenpingol-xibo: This makes more sense than updating the xiboIC.checkIsPreview(). Let me add it to xiboIC then

maurofmferrao (approved):

dasgarner (approved):

rubenpingol-xibo:
TEST: /Layout END

dasgarner (approved):

ifarzana (approved):
Approving to build PR container

PR #2399, opened 2024-02-29
Layout: sync together canvas and layout sharing.

dasgarner:
Updating layout sharing should update any Canvas regions with the same. Adding a new canvas region should adopt the layout sharing that already exists. A double-lock.

In addition, trying to set canvas permissions separately will throw an error (@maurofmferrao I think you will remove that option once this is merged?)


PeterMis (approved):

ifarzana:
TEST: all

ifarzana (approved):

PR #2400, opened 2024-02-29
QA: Streamline build process for easier manual and automated tested

dasgarner:
Build on PR approval. test using build container.


ifarzana (approved):

dasgarner:
TEST: all END

dasgarner:
The test suite workflow from develop gets used because its a comment workflow, so we need to merge this first without having the test suite run (as we've deleted the Dockerfile.ci it uses to build)

PR #2404, opened 2024-03-01
DataSet: Add oauth2 client_credentials support for access_token auth

harbormw:
Implements basic framework for oauth2.0 authentication for remote datasets.

+ Added 4 fields to dataset table
+ Updated dataset entity/controller/factory to support new fields
+ Updated `views` dataset twigs for new fields
+ Updated dataset factory logic to interact with the `oauth2Url` and pass token as `Bearer` in header to request going to the remote dataset api endpoint


dasgarner:
There have been some other requests for this feature, so I've opened an issue to track it: https://github.com/xibosignage/xibo/issues/3566

I think this PR is a good foundation, but needs to changes, happy to elaborate on those if you're still keen on working on it.

PR #2414, opened 2024-03-05
Elements, Widgets, and Regions need to obey sharing

maurofmferrao:
relates to xibosignage/xibo#3353

### DIFF
--- a/lib/Controller/Layout.php
+++ b/lib/Controller/Layout.php
@@ -1542,6 +1542,9 @@ class Layout extends Base
                         // Augment with deletable flag
                         $widget->setUnmatchedProperty('isDeletable', $this->getUser()->checkDeleteable($widget));
 
+                        // Augment with viewable flag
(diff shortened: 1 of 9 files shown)


lib/Controller/Layout.php
```diff
@@ -1542,6 +1542,9 @@ public function grid(Request $request, Response $response)
                         // Augment with deletable flag
                         $widget->setUnmatchedProperty('isDeletable', $this->getUser()->checkDeleteable($widget));
 
+                        // Augment with viewable flag
+                        $widget->setUnmatchedProperty('isViewable', $this->getUser()->checkViewable($widget));
```
dasgarner, line 1546: These are actually always viewable because they are returned by the factory which does a view check by default - so we could just set these to true

dasgarner: Sorry, I mean the playlist one!

ifarzana:
TEST: all

ifarzana:

PR #2418, opened 2024-03-05
xfA V3 Player: shows items more frequently than desired

rubenpingol-xibo:
relates to xibosignage/xibo#3343


dasgarner:
As this comes from your fork, the container to build for testing the PR will not have permission to push to the registry.

I suggest in this case that we merge the PR to develop and test this using the old method (it is all player testing anyway, which we know will be harder to do locally). @ifarzana do you agree?

PR #2431, opened 2024-03-14
Editor toolbar: Load preferences for a new user is broken

maurofmferrao:
relates to xibosignage/xibo#3371


dasgarner (approved):

ifarzana:
TEST: /Administration, login.cy.js, unauthed.cy.js END

PR #2432, opened 2024-03-14
Fix JS error and moveable when adding from playlist tab

maurofmferrao:
relates to xibosignageltd/xibo-private#639


dasgarner (approved):

ifarzana:
TEST: /Layout END

ifarzana:
The Cypress test for "adding a clock widget" failed because the toolbar was not initially present in the container.

PR #2437, opened 2024-03-19
Template Editor : add missing features

PeterMis:
Relates to https://github.com/xibosignageltd/xibo-private/issues/643


dasgarner (approved):
We should add some phpdocs to the controller methods when we do the finishing touches. This looks OK to merge now.

PR #2453, opened 2024-03-26
Misc sanitisation improvements v3

dasgarner:


PeterMis (approved):

ifarzana:
TEST: all

ifarzana:
I am not sure if we can run cypress tests for this @dasgarner

dasgarner:
> I am not sure if we can run cypress tests for this @dasgarner

No we won't be able to do unfortunately. I think for this small case we can ignore that - I will merge.

PR #2460, opened 2024-03-28
Widget: DataSet issues with duration is per item and freshness timeout

dasgarner:
This PR addresses two issues:
 - The freshness timeout property in DataSet does not work for Elements
 - The validation for ensuring we have a number of items makes more sense on numItems vs upperLimit


maurofmferrao (approved):
Looks good to me!

gracec20:
TEST: /Layout, /Library END

PR #2461, opened 2024-03-28
User: update help text to be clearer

dasgarner:
fixes xibosignage/xibo#3376


PeterMis (approved):

dasgarner:
@ifarzana if you are in agreement I don't think there is an QA to do here?

ifarzana:
I agree, no QA is required as it is a small change in text, Tx

PR #2462, opened 2024-03-28
Widget: Stocks and Currencies invalid duration when items per page is used

dasgarner:
Widget uses maxItemsPerPage instead of itemsPerPage. Also improve error responses from AlphaVantageConnector.

 fixes xibosignage/xibo#3368


maurofmferrao (approved):
Looks good to me!

ifarzana:
TEST: /Layout END

PR #2466, opened 2024-04-02
Global Elements: Add Date

maurofmferrao:
relates to xibosignageltd/xibo-private#662

### DIFF
--- a/modules/templates/global-elements.xml
+++ b/modules/templates/global-elements.xml
@@ -193,12 +193,13 @@
         <title>Date</title>
         <type>element</type>
         <dataType>global</dataType>
+        <showIn>none</showIn>
         <canRotate>true</canRotate>
         <thumbnail>date-thumb</thumbnail>
-        <startWidth>380</startWidth>
+        <startWidth>300</startWidth>
         <startHeight>80</startHeight>
         <properties>
-            <property id="date" type="date" format="DD/MM/Y HH:mm:ss" variant="dateTime">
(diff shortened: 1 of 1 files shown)


maurofmferrao:
> How does this know to show the current date and keep it updated? I can't find where it would do that?

Oh I didn't remember it was supposed to do that, I'll add a checkbox to "Show current Date" then, that hides the date input. 👍🏻

modules/templates/global-elements.xml
```diff
@@ -362,6 +362,230 @@ $(target).find('.date').each(function(_idx, dateEl){
...
+    // Update every second
+    window['updateInterval_' + id] = setInterval(function() {
+        // If element was removed, clear interval
+        if(!$dateEl.is(':visible')) {
+            clearInterval(window['updateInterval_' + id]);
+            return;
+        }
+
+        // Update date value
+        var currentDate = moment().add(offset, "m");
```
dasgarner, line 553: `offset` isn't defined?

dasgarner (changes requested):

modules/templates/global-elements.xml
maurofmferrao: `offset` is defined on line 538, but since it's only used on the `setInterval`, I can always move it to that scope.

dasgarner: No that's ok, I understand now! 👍

ifarzana:
TEST: /Layout/Editor/layout_editor_empty.cy.js END

PR #2479, opened 2024-04-10
Widget: Fix marquee effect implementation from xiboTextRender

rubenpingol-xibo:
relates to xibosignage/xibo#3388

### DIFF
--- a/modules/src/editor-render.js
+++ b/modules/src/editor-render.js
@@ -29,10 +29,12 @@ $(function() {
       globalOptions.originalWidth = e.data.options.originalWidth;
       globalOptions.originalHeight = e.data.options.originalHeight;
 
+      // Used in case we are on the editor to set pauseEffectOnStart
+      const isEditor = xiboIC.checkIsEditor();
       // Set the pause state for animation to false
       // To start right after the render effects are generated
       globalOptions.pauseEffectOnStart =
-        e.data.options.pauseEffectOnStart ?? false;
+        e.data.options.pauseEffectOnStart ?? !isEditor;
 
       // Arguments for both reRender
       const args = (typeof widget != 'undefined') ? [
--- a/modules/src/editor-render.js
+++ b/modules/src/editor-render.js
@@ -29,12 +29,10 @@ $(function() {
       globalOptions.originalWidth = e.data.options.originalWidth;
       globalOptions.originalHeight = e.data.options.originalHeight;
 
-      // Used in case we are on the editor to set pauseEffectOnStart
-      const isEditor = xiboIC.checkIsEditor();
       // Set the pause state for animation to false
       // To start right after the render effects are generated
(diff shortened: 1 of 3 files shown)


maurofmferrao (approved):

rubenpingol-xibo:
I believe that the deleted lines here are not necessary anymore since we have moved the template animation being handled by `onTemplateVisible` using `xiboLayoutAnimate`. But I'm not certain yet if we still need those deleted lines here for module only.

rubenpingol-xibo:
Can you confirm @maurofmferrao?

maurofmferrao:
> Can you confirm @maurofmferrao?

I think that makes sense, but we need just to make sure all modules/templates that were using `xiboTextRender` also run `xiboLayoutAnimate`. 

What happens if we have a template that needs to be animated on `onTemplateRender` and not on `onTemplateVisible`? Do we ever want to do that @dasgarner ?

maurofmferrao:
We also need to test if the behaviour change on the editor, it might not start the animation on resizing :/

rubenpingol-xibo:
> We also need to test if the behaviour change on the editor, it might not start the animation on resizing :/

Yes, I think this makes sense and I can see it's not animating now on resize. I can add `isEditor` to its condition

maurofmferrao:
> > We also need to test if the behaviour change on the editor, it might not start the animation on resizing :/
> 
> Yes, I think this makes sense and I can see it's not animating now on resize. I can add `isEditor` to its condition

Not sure we want that, what if in a different widget we want to run on load and not on visible, even on a player?

The question is: why is `pauseEffectOnStart` false and running those lines since its default value is true?

rubenpingol-xibo:
> > > We also need to test if the behaviour change on the editor, it might not start the animation on resizing :/
> > 
> > 
> > Yes, I think this makes sense and I can see it's not animating now on resize. I can add `isEditor` to its condition
> 
> Not sure we want that, what if in a different widget we want to run on load and not on visible, even on a player?
> 
> The question is: why is `pauseEffectOnStart` false and running those lines since its default value is true?

I think `pauseEffectOnStart` is being set to **false** on `editor-render.js` since it's being set to have a default **false** value

rubenpingol-xibo:
No, I think it's also being set to have default value of **false** on `xibo-player`.

maurofmferrao:
Yeah, it's being set as false on those two places, but maybe the fix is just to set it as `true` by default?

I think that we want the render to always pause on start except when we force it to start right away, on a specific module or on the editor. So we still need to check if it's the editor to start the animation, maybe do that in `editor-render.js` instead of in `xiboTextRender`?

rubenpingol-xibo:
Yes, I think we should proceed with this.

maurofmferrao (approved):
Looks good to me!

dasgarner:
TEST: /Layout END

PR #2544, opened 2024-06-02
Folders: menuboard add ignores the provided folderId.

dasgarner:
Adjust the feature check to be inline with other controllers add function
fixes xibosignage/xibo#3416

### DIFF
--- a/lib/Controller/MenuBoard.php
+++ b/lib/Controller/MenuBoard.php
@@ -38,34 +38,14 @@ use Xibo\Support\Exception\NotFoundException;
 class MenuBoard extends Base
 {
     /**
-     * @var MenuBoardFactory
-     */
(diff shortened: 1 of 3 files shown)


PeterMis (commented):
This would certainly fix the folderId issue, however I think there might be one more thing we should address as well - with the permissionsFolderId.

Could we perhaps : 
- Look up Folder `$folder` by id
- Get permissionsFolderId via `$folder->getPermissionFolderIdOrThis()`
- Pass the permissionsFolderId to the create function (and set it on the object there)

Just to ensure that is set correctly and does not always default to Root (1) no matter what :)

PeterMis (approved):

PeterMis (approved):

ifarzana:
TEST: /Layout END

PR #2549, opened 2024-06-04
Module: cache issues with weather and others

dasgarner:
Fix an issue where the data provider `useDisplayLocation` variable in the cache key is not substituted correctly https://github.com/xibosignage/xibo/issues/3422

Also a potential solution for "cache not ready" on display specific caches, caused by not supplying the data provider with lat/long/displayId in all cases.


PeterMis (approved):

ifarzana:
TEST: /Layout END

ifarzana:

PR #2550, opened 2024-06-04
Library: replaced images are not replaced in layouts when they are used as elements

dasgarner:
This PR implements element checking on replacement of an image in the Library.
fixes xibosignage/xibo#3418


PeterMis (approved):

gracec20:
TEST: /Layout, /Library END

PR #2551, opened 2024-06-05
Fix - PlayerJS: Pinned slot skipping first item on next page

rubenpingol-xibo:
relates to xibosignage/xibo#3420


maurofmferrao (approved):
Looks good to me!

PR #2554, opened 2024-06-06
Library upload API returns plain/text instead of application/json (fixed)

nadzpogi:
relates to xibosignage/xibo#3427

- Removed the setting of the content-type header to application/json in some of the modified methods where it was previously set to json.


PeterMis (approved):

ifarzana:
TEST: /Library END

PR #2558, opened 2024-06-10
Module: data provider lat/lng might return null if it has not ever need set on a display

dasgarner:
xibosignage/xibo#3428

The prior fix of this issue cause this error:

```
Xibo\Widget\Provider\DataProvider::getDisplayLatitude(): Return value must be of type float, null returned
```

This is because some displays will have `null` in their lat/lng. It is possible we would have seen this error before, but we were not reliably using the lat/lng in all cases.


ifarzana:
TEST: /Layout END

ifarzana:

PR #2559, opened 2024-06-10
Module: data provider: make sure there is always a lat/lon

dasgarner:
The root cause of our issue is that we did not provide the CMS lat/lon when the display one was null/empty.

xibosignage/xibo#3428


ifarzana (approved):

ifarzana:
TEST: /Layout END

PR #2561, opened 2024-06-11
Module: HTML cache can be generated via XMDS which fails for HLS.

dasgarner:
In this PR we add the url_decode filter to the XMDS entrypoint so that if we need to generate widget HTML (which uses a twig file) we have that filter available.

xibosignage/xibo#3433


nadzpogi (approved):
Looks good to me!

ifarzana:
TEST: login.cy.js END

PR #2562, opened 2024-06-11
Merge 4.0.12 into 4.1.0-alpha2

dasgarner:
@maurofmferrao I have requested your review here as there were some tricky merges around the image element used in forecast/global


maurofmferrao (approved):
Looks good to me! (🤞🏻)

PR #2564, opened 2024-06-12
Sync Group: Add Switch Delay and Video Pause Delay

mgbaybay:
## Changes:
- Added Switch Delay and Video Pause Delay in Sync Group

Relates to: https://github.com/xibosignage/xibo/issues/3200

### DIFF
--- a/db/migrations/20240612112949_add_switch_delay_and_video_pause_delay_to_sync_group.php
+++ b/db/migrations/20240612112949_add_switch_delay_and_video_pause_delay_to_sync_group.php
@@ -1,0 +1,40 @@
+<?php
+/*
+ * Copyright (C) 2024 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use Phinx\Migration\AbstractMigration;
+
+class AddSwitchDelayAndVideoPauseDelayToSyncGroup extends AbstractMigration
+{
+
+    /**
+     * Added syncSwitchDelay and syncVideoPauseDelay columns to Sync Groups
+     * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+     */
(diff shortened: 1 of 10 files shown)


lib/Xmds/Soap5.php
```diff
@@ -476,6 +476,16 @@ public function RegisterDisplay(
             $value = $return->createTextNode($syncGroup->syncPublisherPort ?? 9590);
             $syncPublisherPortNode->appendChild($value);
             $displayElement->appendChild($syncPublisherPortNode);
+
+            $syncSwitchDelayNode = $return->createElement('syncSwitchDelay');
+            $value = $return->createTextNode($syncGroup->syncSwitchDelay ?? '');
```
dasgarner, line 481: We should return the default values for these options, rather than empty. I think it would be handled correctly on the player as-is, but its clearer to always have them set.

views/syncgroup-page.twig
```diff
@@ -130,6 +130,8 @@
             { "data": "owner", responsivePriority: 3 },
             { "data": "modifiedByName", responsivePriority: 4 },
             { "data": "syncPublisherPort", responsivePriority: 3 },
+            { "data": "syncSwitchDelay", responsivePriority: 3 },
```
dasgarner, line 133: You will also need to add some table headings for these fields in the HTML output above

db/migrations/20240612112949_add_switch_delay_and_video_pause_delay_to_sync_group.php
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use Phinx\Migration\AbstractMigration;
+
+class AddSwitchDelayAndVideoPauseDelayToSyncGroup extends AbstractMigration
```
dasgarner, line 25: These PHPCS errors can both be addressed please. For the namespace issue we can suppress it in a class comment

```
@phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
```

You can run PHPCS locally which would highlight both of these ahead of creating a PR.

dasgarner (changes requested):
Overall this looks good - I have a few comments/suggestions

views/syncgroup-page.twig
mgbaybay: Thanks again for the review! Table headings has been added on this [commit](https://github.com/xibosignage/xibo-cms/pull/2564/commits/d2bf44bfdc0bae292e60291730cf72b9b009b411).

db/migrations/20240612112949_add_switch_delay_and_video_pause_delay_to_sync_group.php
mgbaybay: Finally, PHPCS errors has also been addressed on this [commit](https://github.com/xibosignage/xibo-cms/pull/2564/commits/d2bf44bfdc0bae292e60291730cf72b9b009b411). Thank you again for the review!

dasgarner (approved):

ifarzana:
TEST: login.cy.js END

PR #2565, opened 2024-06-12
Module: webpage widget should use original width/height

dasgarner:
These parameters represent region width/height and should be used rather than determining it client side.

fixes xibosignage/xibo#3424


maurofmferrao (approved):

ifarzana:
TEST: /Layout END

PR #2566, opened 2024-06-13
Module: improve the way widget data is cached and returned

dasgarner:
https://github.com/xibosignage/xibo/issues/3434

This PR recognises that when using Stash the OLD invalidation method, it will only return old data if another cache is currently regenerating the cache and has called lock. This means that we were very unlikely to return old cache via XMDS.

This PR extends the expiry time on all caches by 50% or sets it to 15 minutes, whichever is longer. The original expiry information is stored inside the cache, so that we can decide ourselves whether that data is a miss or valid.

The upshot is that we should avoid any getData errors once this PR is included, although there will be a possibility of returning old data for widgets which are set to update data very frequently.

### DIFF
--- a/lib/Widget/Render/WidgetDataProviderCache.php
+++ b/lib/Widget/Render/WidgetDataProviderCache.php
@@ -216,11 +216,10 @@ class WidgetDataProviderCache
         }
 
         // Set some cache dates so that we can track when this data provider was cached and when it should expire.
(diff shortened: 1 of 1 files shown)


mgbaybay (approved):
Excellent! The approach is straightforward and is easy to follow. Comments were also helpful in providing context to the logic

ifarzana:
TEST: /Layout END

PR #2567, opened 2024-06-13
XMDS: optimise the way notifications are inserted.

dasgarner:


dasgarner:
@ifarzana Nadz has reviewed these changes and says they are OK from a syntax perspective and that there are no obvious errors - given the scope of the changes I think this will have to be sufficient (in Peter's absence).

If you're OK with that, please can you approve so the PR is built?

nadzpogi (approved):
Looks Good!

PR #2568, opened 2024-06-14
Widgets: Add the ability to filter datasets by display tag

nadzpogi:
relates to xibosignage/xibo#3062

### DIFF
--- a/lib/Entity/DataSet.php
+++ b/lib/Entity/DataSet.php
@@ -524,8 +524,10 @@ class DataSet implements \JsonSerializable
 
                 // Loop through each tag and get the actual tag value from the database
                 foreach ($displayTags as $tag) {
-                    $tagName = $tag['tagName'];
-                    $defaultTagValue = $tag['defaultValue'];
+                    $tagSanitizer = $this->getSanitizer($tag);
+
+                    $tagName = $tagSanitizer->getString('tagName');
+                    $defaultTagValue = $tagSanitizer->getString('defaultValue');
                     $tagString = $tag['tagString'];
 
                     $query = 'SELECT `lktagdisplaygroup`.`value` AS tagValue
--- a/lib/Entity/DataSet.php
+++ b/lib/Entity/DataSet.php
@@ -494,47 +494,69 @@ class DataSet implements \JsonSerializable
 
         // Fetch display tag value/s
         if ($filter != '' && $displayId != 0) {
-            $displayTags = [];
 
-            // Define the regular expression to match [Tag:tagNameHere]
-            $pattern = '/\[Tag:(\w+)\]/';
+            // Define the regular expression to match [Tag:...]
+            $pattern = '/\[Tag:[^]]+\]/';
 
-            // find all matches
+            // Find all instances of [Tag:...]
             preg_match_all($pattern, $filter, $matches);
 
             // Check if matches were found
-            if (!empty($matches[1])) {
-                // Iterate through the matches and add them to the displayTags array
-                foreach ($matches[1] as $tag) {
-                    $displayTags[] = $tag;
+            if (!empty($matches[0])) {
+                $displayTags = [];
+
+                // Iterate through the matches and process each tag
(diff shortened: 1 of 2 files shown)


lib/Entity/DataSet.php
```diff
@@ -492,6 +492,52 @@ public function getData($filterBy = [], $options = [])
             'connection' => 'default'
         ], $options);
 
+        // Fetch display tag value/s
+        if ($filter != '' && $displayId != 0) {
+            $displayTags = [];
+
+            // Define the regular expression to match [Tag:tagNameHere]
+            $pattern = '/\[Tag:(\w+)\]/';
```
dasgarner, line 500: Does this work with all allowed tag names? I.e. if I have a tag called `$storeNo` ?

lib/Entity/DataSet.php
```diff
@@ -492,6 +492,52 @@ public function getData($filterBy = [], $options = [])
...
+            if (!empty($matches[1])) {
+                // Iterate through the matches and add them to the displayTags array
+                foreach ($matches[1] as $tag) {
+                    $displayTags[] = $tag;
+                }
+            }
+
+            // Loop through each tag and get the actual tag value from the database
+            foreach ($displayTags as $tag) {
+                $query = "
```
dasgarner, line 515: String should be enclosed in `'` instead of `"` please

lib/Entity/DataSet.php
```diff
@@ -492,6 +492,52 @@ public function getData($filterBy = [], $options = [])
...
+                $params = [
+                    'displayId' => $displayId,
+                    'tag' => $tag
+                ];
+
+                // Execute the query
+                $results = $this->getStore()->select($query, $params);
+
+                // Check if results were found and get the tag value
+                $tagValue = !empty($results) ? $results[0]['tagValue'] : '';
```
dasgarner, line 534: I don't think this caters for the case where the display has the tag, but no value has been set?

dasgarner (changes requested):
Looks good overall, but I think there are a few minor adjustments to make.

We also spoke about an extension to allow a default value to be provided along with the syntax, e.g. `[Tag:tagValue:defaultValue]` - we'd use the default value if the tag was not present on the display. Will you be adding that also?

lib/Entity/DataSet.php
nadzpogi: Done!

lib/Entity/DataSet.php
nadzpogi: I've replaced it now with a more generic regular expression to match `[Tag:...]` instead.

lib/Entity/DataSet.php
```diff
@@ -494,47 +494,69 @@ public function getData($filterBy = [], $options = [])
...
-                $results = $this->getStore()->select($query, $params);
-
-                // Check if results were found and get the tag value
-                $tagValue = !empty($results) ? $results[0]['tagValue'] : '';
-
-                // Replace [Tag:anyTagNameHere] in the $filter with the actual tag value
-                $filter = preg_replace("/\[Tag:$tag\]/", $tagValue, $filter);
+                // Loop through each tag and get the actual tag value from the database
+                foreach ($displayTags as $tag) {
+                    $tagName = $tag['tagName'];
```
dasgarner, line 527: This is user input, so we need to load the parsed out $tag array into a sanitizer and ensure the values we've received are strings as expected. Have a think about where these values are used to see what I mean.

dasgarner (changes requested):
Overall this is looking great - I have one final comment

lib/Entity/DataSet.php
nadzpogi: Thanks for pointing that out. I've now loaded the $tag array into a sanitizer and made sure that the tagName and defaultValue that we're getting are strings.

dasgarner (approved):

ifarzana:
TEST: /Library, /Layout END

PR #2571, opened 2024-06-14
Dashboard: remove unused code to improve page load performance.

dasgarner:
This PR adjusts two parts of the status dashboard logic.

1. we queried and looped over display groups for no apparent reason (likely this is legacy code left by mistake) - removed
2. we uses a count() method on an array of users, which has been switched to a direct query approach instead

fixes xibosignage/xibo#3440

### DIFF
--- a/lib/Factory/UserFactory.php
+++ b/lib/Factory/UserFactory.php
@@ -415,4 +415,46 @@ class UserFactory extends BaseFactory
 
         return $entries;
     }
+
+    /**
+     * Get a count of users, with respect to the logged-in user
+     * @return int
+     */
+    public function count(): int
+    {
+        $params = [];
+        $sql = '
+        SELECT COUNT(DISTINCT `user`.userId) AS countOf
+          FROM `user`
+            INNER JOIN `lkusergroup`
+            ON `lkusergroup`.userId = `user`.userId
+            INNER JOIN `folder`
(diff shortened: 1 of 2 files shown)


lib/Factory/UserFactory.php
```diff
@@ -415,4 +415,46 @@ public function query($sortOrder = [], $filterBy = [])
...
+            INNER JOIN `group`
+            ON `group`.groupId = `lkusergroup`.groupId
+              AND `group`.isUserSpecific = 1
+        ';
+
+        // Super admins should get a count of all users in the system.
+        if (!$this->getUser()->isSuperAdmin()) {
+            // Non-super admins should only get a count of users in their group
+            $sql .= '
+                WHERE user.userId IN (
```
nadzpogi, line 442: `user.userId` should be \`user\`.userId instead (with backticks).

nadzpogi (changes requested):
Just need a very minor change. Everything else looks good to me!

lib/Factory/UserFactory.php
dasgarner: Thanks, good catch.

nadzpogi (approved):
All good!

ifarzana:
TEST: dashboard.cy.js END

PR #2574, opened 2024-06-17
Display: register should process operating system if sent by the player

mgbaybay:
## Changes:
- Added osVersion, osSdk, manufacturer, brand, model in `Display` 

Relates to: https://github.com/xibosignageltd/xibo-private/issues/725

### DIFF
--- a/lib/Xmds/Soap4.php
+++ b/lib/Xmds/Soap4.php
@@ -50,17 +50,12 @@ class Soap4 extends Soap
      * @param string $macAddress
      * @param null $xmrChannel
      * @param null $xmrPubKey
-     * @param string $osVersion
-     * @param string $osSdk
-     * @param string $manufacturer
-     * @param string $brand
-     * @param string $model
      * @return string
      * @throws GeneralException
      * @throws NotFoundException
      * @throws \SoapFault
      */
-    public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientType, $clientVersion, $clientCode, $operatingSystem, $macAddress, $xmrChannel = null, $xmrPubKey = null, $osVersion, $osSdk, $manufacturer, $brand, $model)
+    public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientType, $clientVersion, $clientCode, $operatingSystem, $macAddress, $xmrChannel = null, $xmrPubKey = null)
     {
         $this->logProcessor->setRoute('RegisterDisplay');
 
@@ -75,11 +70,6 @@ class Soap4 extends Soap
             'macAddress' => $macAddress,
             'xmrChannel' => $xmrChannel,
             'xmrPubKey' => $xmrPubKey,
-            'osVersion' => $osVersion,
-            'osSdk' => $osSdk,
-            'manufacturer' => $manufacturer,
-            'brand' => $brand,
-            'model' => $model,
         ]);
 
         // Sanitize
@@ -89,12 +79,9 @@ class Soap4 extends Soap
         $clientType = $sanitized->getString('clientType');
         $clientVersion = $sanitized->getString('clientVersion');
         $clientCode = $sanitized->getInt('clientCode');
-        $osVersion = $sanitized->getString('osVersion');
-        $osSdk = $sanitized->getString('osSdk');
-        $manufacturer = $sanitized->getString('manufacturer');
-        $brand = $sanitized->getString('brand');
-        $model = $sanitized->getString('model');
+        $macAddress = $sanitized->getString('macAddress');
         $clientAddress = $this->getIp();
+        $operatingSystem = $sanitized->getString('operatingSystem');
 
         // Check the serverKey matches
         if ($serverKey != $this->getConfig()->getSetting('SERVER_KEY')) {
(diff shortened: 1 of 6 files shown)


lib/Xmds/Soap4.php
```diff
@@ -50,12 +50,17 @@ class Soap4 extends Soap
...
+     * @param string $manufacturer
+     * @param string $brand
+     * @param string $model
      * @return string
      * @throws GeneralException
      * @throws NotFoundException
      * @throws \SoapFault
      */
-    public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientType, $clientVersion, $clientCode, $operatingSystem, $macAddress, $xmrChannel = null, $xmrPubKey = null)
+    public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientType, $clientVersion, $clientCode, $operatingSystem, $macAddress, $xmrChannel = null, $xmrPubKey = null, $osVersion, $osSdk, $manufacturer, $brand, $model)
```
dasgarner, line 63: Just a little hint while you're working on this still (I realise its a draft) - we're not adding additional parameters to this API, instead the details will appear in the existing `operatingSystem` parameter (either as JSON or plain string) as described in the issue.

lib/Xmds/Soap4.php
```diff
@@ -271,6 +285,11 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
         $display->clientType = $clientType;
         $display->clientVersion = $clientVersion;
         $display->clientCode = $clientCode;
+        $display->osVersion = $osVersion;
+        $display->osSdk = $osSdk;
+        $display->manufacturer = $manufacturer;
+        $display->brand = $brand;
+        $display->model = $model;
         //$display->operatingSystem = $operatingSystem;
```
dasgarner, line 293: As per the above comment, this is where we'd parse out the existing parameter.

lib/Xmds/Soap4.php
mgbaybay: Hi, Dan! I really appreciate the review. Could you please take a look at the latest [commit ](https://github.com/xibosignage/xibo-cms/pull/2574/commits/48e7c9c0fab2c49afeb7846b0a06ac615c016a81) and let me know your thoughts? Thank you so much!

lib/Xmds/Soap4.php
mgbaybay: Based on the latest [commit](https://github.com/xibosignage/xibo-cms/pull/2574/commits/48e7c9c0fab2c49afeb7846b0a06ac615c016a81), requesting re-review here as well, please. Thank you!

lib/Xmds/Soap4.php
```diff
@@ -272,6 +273,16 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
         $display->clientVersion = $clientVersion;
         $display->clientCode = $clientCode;
         //$display->operatingSystem = $operatingSystem;
+
+        // Parse operatingSystem data
+        $operatingSystem = json_decode($operatingSystem, true);
```
dasgarner, line 278: Newer versions of players will submit a JSON string to this endpoint, and in that case this code looks good. However older players do not know how to do that yet, and so the JSON decode may fail.

We should handle that case, and instead take the original $operatingSystem as provided by the player, and submit it to osVersion.

lib/Xmds/Soap4.php
```diff
@@ -272,6 +273,16 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
         $display->clientVersion = $clientVersion;
         $display->clientCode = $clientCode;
         //$display->operatingSystem = $operatingSystem;
```
dasgarner, line 275: We could remove this comment

lib/Xmds/Soap4.php
mgbaybay: Thank you again for the review! Feedback has been addressed [here](https://github.com/xibosignage/xibo-cms/pull/2574/commits/d783e966a3136dce6a1006af8fdca48b4cfe7c0f). Requesting re-review, please. Thanks!

lib/Xmds/Soap4.php
```diff
@@ -271,7 +272,18 @@ public function RegisterDisplay($serverKey, $hardwareKey, $displayName, $clientT
...
         $display->clientVersion = $clientVersion;
         $display->clientCode = $clientCode;
-        //$display->operatingSystem = $operatingSystem;
+
+        // Parse operatingSystem data
+        $operatingSystem = json_decode($operatingSystem, false);
+
+        // Newer version of players will return a JSON value, but for older version, it will return a string.
+        // In case the json decode fails, use the operatingSystem string value as the default value for the osVersion.
+        $display->osVersion = isset($operatingSystem) ? $operatingSystem->osVersion : $operatingSystem;
```
dasgarner, line 281: If json_decode fails, the operating system string value will be `false` (because the return of json_decode will be false overwriting the variable.

You'll probably need another temporary variable 🙂

dasgarner (changes requested):

lib/Xmds/Soap4.php
mgbaybay: Thank you again for the review! Please see the latest changes [here](https://github.com/xibosignage/xibo-cms/pull/2574/commits/803b8ba158a782468b7a6cb41692978118c396fc).

dasgarner (approved):

ifarzana (approved):
Approving a minor change and rebuild PR container

ifarzana:
TEST: /Display END

ifarzana:
TEST: /Display/displaysettings.cy.js END

PR #2591, opened 2024-06-28
Custom schedule criteria (via connectors)

nadzpogi:
relates to xibosignageltd/xibo-private#704

### DIFF
--- a/lib/Connector/OpenWeatherMapConnector.php
+++ b/lib/Connector/OpenWeatherMapConnector.php
@@ -26,6 +26,8 @@ use Carbon\Carbon;
 use GuzzleHttp\Exception\RequestException;
 use Illuminate\Support\Str;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Xibo\Event\ScheduleCriteriaRequestEvent;
+use Xibo\Event\ScheduleCriteriaRequestInterface;
 use Xibo\Event\WidgetDataRequestEvent;
 use Xibo\Support\Exception\GeneralException;
 use Xibo\Support\Sanitizer\SanitizerInterface;
@@ -58,6 +60,7 @@ class OpenWeatherMapConnector implements ConnectorInterface
     public function registerWithDispatcher(EventDispatcherInterface $dispatcher): ConnectorInterface
     {
         $dispatcher->addListener(WidgetDataRequestEvent::$NAME, [$this, 'onDataRequest']);
+        $dispatcher->addListener(ScheduleCriteriaRequestEvent::$NAME, [$this, 'onScheduleCriteriaRequest']);
         return $this;
     }
 
@@ -637,5 +640,51 @@ class OpenWeatherMapConnector implements ConnectorInterface
             'W' => [247.5, 292.5],
             'NW' => [292.5, 337.5]
         ];
+    }
+
+    public function onScheduleCriteriaRequest(ScheduleCriteriaRequestInterface $event): void
+    {
+        // Initialize Open Weather Schedule Criteria parameters
+        $event->addType('open_weather_map', __('Open Weather Map'))
+                ->addMetric('owm_weather_condition', __('Weather Condition'))
+                    ->addValues('dropdown', [
+                        'Clear Sky',
+                        'Few Clouds',
+                        'Scattered Clouds',
+                        'Broken Clouds',
+                        'Shower Rain',
+                        'Rain',
+                        'Thunderstorm',
+                        'Snow',
+                        'Mist'
+                    ])
+                ->addMetric('owm_atmosphere', __('Atmosphere'))
+                    ->addValues('dropdown', [
+                        'List',
+                        'Smoke',
+                        'Haze',
+                        'Dust',
+                        'Fog',
+                        'Sand',
+                        'Dust',
+                        'Ash',
+                        'Squall',
+                        'Tornado'
+                    ])
+                ->addMetric('owm_temperature', __('Temperature'))
+                    ->addValues('number', [])
+                ->addMetric('own_wind_speed', __('Wind Speed'))
+                    ->addValues('number', [])
+                ->addMetric('owm_wind_direction', __('Wind Direction'))
+                    ->addValues('number', [])
+                ->addMetric('owm_humidity', __('Humidity'))
+                    ->addValues('number', [])
+                ->addMetric('owm_pressure', __('Pressure'))
+                    ->addValues('number', [])
+                ->addMetric('owm_visibility', __('Visibility'))
+                    ->addValues('number', []);
+
+
+        $this->getLogger()->debug('>> onScheduleCriteriaRequest method called');
     }
 }
--- a/lib/Connector/OpenWeatherMapConnector.php
+++ b/lib/Connector/OpenWeatherMapConnector.php
@@ -651,43 +651,43 @@ class OpenWeatherMapConnector implements ConnectorInterface
     public function onScheduleCriteriaRequest(ScheduleCriteriaRequestInterface $event): void
     {
         // Initialize Open Weather Schedule Criteria parameters
-        $event->addType('open_weather_map', __('Open Weather Map'))
-                ->addMetric('owm_weather_condition', __('Weather Condition'))
+        $event->addType('weather', __('Weather'))
+                ->addMetric('weather_condition', __('Weather Condition'))
                     ->addValues('dropdown', [
-                        'Clear Sky',
-                        'Few Clouds',
-                        'Scattered Clouds',
-                        'Broken Clouds',
-                        'Shower Rain',
-                        'Rain',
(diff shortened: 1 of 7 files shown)


lib/Connector/OpenWeatherMapConnector.php
```diff
@@ -638,4 +642,52 @@ private static function cardinalDirections()
...
+     * @return void
+     * @throws ConfigurationException
+     */
+    public function onScheduleCriteriaRequest(ScheduleCriteriaRequestInterface $event): void
+    {
+        // Initialize Open Weather Schedule Criteria parameters
+        $event->addType('open_weather_map', __('Open Weather Map'))
+                ->addMetric('owm_weather_condition', __('Weather Condition'))
+                    ->addValues('dropdown', [
+                        'Clear Sky',
```
dasgarner, line 657: These values should be id/title and the title translated.

lib/Connector/OpenWeatherMapConnector.php
```diff
@@ -638,4 +642,52 @@ private static function cardinalDirections()
...
+
+    /**
+     * @param ScheduleCriteriaRequestInterface $event
+     * @return void
+     * @throws ConfigurationException
+     */
+    public function onScheduleCriteriaRequest(ScheduleCriteriaRequestInterface $event): void
+    {
+        // Initialize Open Weather Schedule Criteria parameters
+        $event->addType('open_weather_map', __('Open Weather Map'))
```
dasgarner, line 654: I can see the reasoning for having `open_weather_map` as the type, however this would make it harder to replace Open Weather Map with a different service.

If we made the type `weather`, then another service could also implement these criteria via another Connector, allowing Open Weather Map to be replaced.

lib/Controller/Schedule.php
```diff
@@ -773,6 +775,15 @@ public function addForm(Request $request, Response $response, ?string $from, ?in
...
         $defaultLong = (float)$this->getConfig()->getSetting('DEFAULT_LONG');
 
+        // Dispatch an event to retrieve criteria for scheduling from the OpenWeatherMap connector.
+        $event = new ScheduleCriteriaRequestEvent();
+        $this->getDispatcher()->dispatch($event, ScheduleCriteriaRequestEvent::$NAME);
+
+        // Retrieve the data from the event
+        $criteria = $event->getCriteria();
+
+        // maybe add a scheduleCriteria and add that in the $addFormData instead of adding the type, metric and value individually
```
dasgarner, line 785: Maybe? I think you have done this?

lib/Event/ScheduleCriteriaRequestEvent.php
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace Xibo\Event;
+
+use Xibo\Support\Exception\ConfigurationException;
+
+class ScheduleCriteriaRequestEvent extends Event implements ScheduleCriteriaRequestInterface
```
dasgarner, line 27: Let's have a comment describing what this class is used for? Something simple, not as much detail as on the interface.

lib/Event/ScheduleCriteriaRequestEvent.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $metric = [
+            'id' => $id,
+            'name' => $name,
+            'values' => null
+        ];
+
+        // Add the metric to the current type
+        if (isset($this->criteria['types'][$this->currentTypeIndex])) {
+            $this->criteria['types'][$this->currentTypeIndex]['metrics'][] = $metric;
```
dasgarner, line 64: `['metrics']` might be undefined?

lib/Event/ScheduleCriteriaRequestEvent.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function addValues(string $inputType, array $values): self
+    {
+        // Restrict input types to 'dropdown', 'number', 'text' and 'date'
+        $allowedInputTypes = ['dropdown', 'number', 'text', 'date'];
+        if (!in_array($inputType, $allowedInputTypes)) {
+            throw new ConfigurationException('Invalid input type.');
```
dasgarner, line 81: Although these are configuration errors on the server, we should still translate the message.

lib/Event/ScheduleCriteriaRequestEvent.php
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        } else {
+            throw new ConfigurationException('Current type is not set.');
+        }
+
+        return $this;
+    }
+
+    public function getCriteria(): array
```
dasgarner, line 102: This needs a phpdoc

lib/Event/ScheduleCriteriaRequestInterface.php
```diff
@@ -0,0 +1,68 @@
...
+namespace Xibo\Event;
+
+use Xibo\Support\Exception\ConfigurationException;
+
+/**
+ * Interface for managing schedule criteria types, metrics, and values.
+ *
+ * Allows the addition of types, metrics, and values in a chained manner:
+ * - Start with addType() to add a new type. Call addType() multiple times to add multiple types.
+ * - Follow with addMetric() to add metrics under the specified type. Call addMetric() multiple times to add multiple metrics to the current type.
```
dasgarner, line 32: phpcs error on line length - do you have your IDE configured to show you these in real time?

ui/src/core/xibo-calendar.js
```diff
@@ -40,6 +40,133 @@ $(document).ready(function() {
         });
     });
 
+    // Set a listener for type change event
+    $('body').on('change', 'select[name="criteria_type[]"]', function() {
```
dasgarner, line 44: Should we think about scoping this to the add/edit form? As it stands any select field with this name is going to be bound.

dasgarner (changes requested):
Overall this looks good!

I have a few suggestions for very minor changes - no logic changes, just some housekeeping.

@maurofmferrao please can you review the JS?

lib/Controller/Schedule.php
nadzpogi: Sorry about this one. This was actually a note for myself while I was still working on the logic. I was supposed to remove it, but it seems I forgot to do so.

ui/src/core/xibo-calendar.js
```diff
@@ -40,6 +40,133 @@ $(document).ready(function() {
...
+        // Get the row where the type was changed
+        var $row = $(this).closest('.form-group');
```
maurofmferrao, line 46: Instead of using this repeatedly in the event handler function, I would instead capture the event in the anonymous function:
```
function(e) {
(...)
```

And then create a simple variable from the target at the top of the function:
```
var $target = $(e.target);
```

To use instead of `$(this)`. This avoids any unwanted scope issues that might come with the word `this`.

ui/src/core/xibo-calendar.js
```diff
@@ -40,6 +40,133 @@ $(document).ready(function() {
...
+
+        if (scheduleCriteria) {
+            if (selectedType === 'custom') {
+                // Use a text input for metrics
+                updateMetricsFieldAsText($row);
+
+                // Use a text input for values
+                updateValueFieldAsText($row);
+            } else if (scheduleCriteria) {
+
```
maurofmferrao, line 60: This is a minor suggestion, since you'll probably update this file, it would be good to remove unnecessary padding from blocks :) (https://eslint.org/docs/latest/rules/padded-blocks)

maurofmferrao (changes requested):
Just some minor JS code convention fixes. Everything else looks good to me!

maurofmferrao (approved):
Looks good to me!

lib/Event/ScheduleCriteriaRequestEvent.php
nadzpogi: `['metrics']` is already initialized when a new type is added. And it will throw a `ConfigurationException` if `addMetrics()` is called without adding a new type.

lib/Event/ScheduleCriteriaRequestInterface.php
nadzpogi: I'm sorry, I overlooked the exceeded line length. But, its already configured in my IDE. I'll keep this in mind moving forward.

dasgarner (approved):

ifarzana:
TEST: /Schedule END

ifarzana:

maurofmferrao (approved):
Looks good to me!

ifarzana:

ifarzana:

ifarzana:
TEST: login.cy.js END

PR #2596, opened 2024-07-01
Bandwidth Report: Fixed Table/Chart is not showing any data

mgbaybay:
## Changes:
- Fixes error in displaying the table/chart in Bandwidth Report
- RCA: When the data returned is empty, infinite value is calculated in the $base, hence, causes the type error

Relates to: https://github.com/xibosignage/xibo/issues/3438


dasgarner (approved):

ifarzana:
TEST: /Reporting END

PR #2598, opened 2024-07-01
SSP Connector: Add activity report summary and chart

mgbaybay:
## Changes:
- Added activity report summary table and chart
- Modified display filter from input to select option based on the available displays
- Modified partner filter from input to select option based on the available partners
- Added dynamic filter to create summary per display/hour/hour and display 

Relates to: https://github.com/xibosignageltd/xibo-private/issues/720

### DIFF
--- a/lib/Connector/ConnectorInterface.php
+++ b/lib/Connector/ConnectorInterface.php
@@ -52,4 +52,5 @@ interface ConnectorInterface
     public function getSettingsFormTwig(): string;
     public function getSettingsFormJavaScript(): string;
     public function processSettingsForm(SanitizerInterface $params, array $settings): array;
+    public function getAvailablePartners(bool $isThrowError = false, ?string $withApiKey = null);
 }
--- a/lib/Connector/ConnectorInterface.php
+++ b/lib/Connector/ConnectorInterface.php
@@ -52,5 +52,4 @@ interface ConnectorInterface
     public function getSettingsFormTwig(): string;
     public function getSettingsFormJavaScript(): string;
     public function processSettingsForm(SanitizerInterface $params, array $settings): array;
-    public function getAvailablePartners(bool $isThrowError = false, ?string $withApiKey = null);
 }
--- a/lib/Controller/Applications.php
+++ b/lib/Controller/Applications.php
@@ -115,14 +115,22 @@ class Applications extends Base
     {
         // Load all connectors and output any javascript.
         $connectorJavaScript = [];
+        $partnersArr = [];
+
         foreach ($this->connectorFactory->query(['isVisible' => 1]) as $connector) {
             try {
                 // Create a connector, add in platform settings and register it with the dispatcher.
                 $connectorObject = $this->connectorFactory->create($connector);
 
                 $settingsFormJavaScript = $connectorObject->getSettingsFormJavaScript();
+                $partners = $connectorObject->getAvailablePartners() ?? [];
+
                 if (!empty($settingsFormJavaScript)) {
                     $connectorJavaScript[] = $settingsFormJavaScript;
+                }
+
+                if (!empty($partners)) {
+                    $partnersArr[] = $partners;
                 }
             } catch (\Exception $exception) {
                 // Log and ignore.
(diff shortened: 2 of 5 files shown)


lib/Connector/ConnectorInterface.php
```diff
@@ -52,4 +52,5 @@ public function isProviderSetting($setting): bool;
     public function getSettingsFormTwig(): string;
     public function getSettingsFormJavaScript(): string;
     public function processSettingsForm(SanitizerInterface $params, array $settings): array;
+    public function getAvailablePartners(bool $isThrowError = false, ?string $withApiKey = null);
```
dasgarner, line 55: This method is specific to the SSP connector and wouldn't be applicable to all connectors, so it isn't appropriate to add it to the interface.

However, we can use the proxy route to add a method specifically to our SSP connector. Have a look at the `connector.edit.form.proxy` route in routes-web.

lib/Controller/Applications.php
```diff
@@ -115,15 +115,23 @@ public function displayPage(Request $request, Response $response)
     {
         // Load all connectors and output any javascript.
         $connectorJavaScript = [];
+        $partnersArr = [];
```
dasgarner, line 118: This is specific to the SSP connector and therefore it isn't appropriate to add it to the applications page. We're using this as a source of data for the partners dropdown. Once we have implemented the proxy route above, we can fetch the partners on form load and populate the list, or even use a search query from the select2 itself (if we format our method to return in "grid" format accordingly).

dasgarner (changes requested):
I have some changes requested on the BE.

dasgarner:
@maurofmferrao please could you review the JavaScript?

views/xibo-ssp-connector-form-javascript.twig
```diff
@@ -153,8 +155,172 @@
...
+          let filteredData = filterData(json.data, filter);
+
+          drawSummaryTable(filteredData, filter);
+      }, true);
     });
+
+    // Watch for filter option changes in Summary tab
+    dialog.find('select[name="campaignFilter"]').on('change', function() {
+      table.ajax.reload(function(json) {
+        let filter = $('select[name="campaignFilter"]').val();
```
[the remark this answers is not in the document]
mgbaybay: Hi, @maurofmferrao, thank you very much for the review! Requesting re-review on this [commit](https://github.com/xibosignage/xibo-cms/pull/2598/commits/ef4450b9eab2453b94b1d69c2b609a12761d6c14), please. Thanks again!

views/xibo-ssp-connector-form-javascript.twig
```diff
@@ -153,8 +155,172 @@
     // Find the apply button
     dialog.find('a[data-apply-button="true"]').on('click', function() {
       $(this).addClass('disabled').append('<span class="saving fa fa-cog fa-spin p-1"></span>');
-      table.ajax.reload(null, true);
+      table.ajax.reload(function(json) {
+          let filter = $('select[name="campaignFilter"]').val();
```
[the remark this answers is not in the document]
mgbaybay: Thanks again for the review! Feedback has been addressed on this [commit](https://github.com/xibosignage/xibo-cms/pull/2598/commits/ef4450b9eab2453b94b1d69c2b609a12761d6c14) :)

views/xibo-ssp-connector-form-javascript.twig
```diff
@@ -69,6 +69,9 @@
 
     dialog.closest('.modal').addClass('modal-big');
 
+    // Get the filter value
+    let filter = $('select[name="campaignFilter"]').val();
```
maurofmferrao, line 73: Is the field `'select[name="campaignFilter"]'` inside of the `dialog`? If so, maybe it's better to use the query on that dialog instead with `dialog.find('select[name="campaignFilter"]')`?

maurofmferrao (changes requested):

views/xibo-ssp-connector-form-javascript.twig
mgbaybay: Thanks for this! Could you please take a look at this again based on this [commit](https://github.com/xibosignage/xibo-cms/pull/2598/commits/02e8d1a991179bed685ec48c20388056ec7d3b2e)?

maurofmferrao (approved):
Looks good to me!

lib/Controller/Applications.php
mgbaybay: Removed changes on this as well, thank you!

dasgarner (approved):

ifarzana:
TEST: /Administration END

PR #2599, opened 2024-07-02
Data Widget: increasing/decreasing the number of slots does not update the duration when durationPerItem is checked

dasgarner:
This PR fixes two problems:
 - the duration was not being calculated when the number of unique slots (pages) was changed
 - the UI was not refreshing layout status after saving elements


PeterMis (approved):

ifarzana:

PR #2600, opened 2024-07-02
Schedule: fix monthly repeat and from/to dates for time zone.

dasgarner:
The from/to date are being shown with respect to CMS timezone, instead of timezone agnostic. In addition the recurrence detail uses the timezone specific date during formatting.

xibosignage/xibo#3429


PeterMis (approved):

ifarzana:
TEST: /Schedule END

ifarzana:
TEST: /Schedule/dayparts.cy.js END

PR #2602, opened 2024-07-02
Media: Fixed image not updating values when resizing

mgbaybay:
## Changes:
- Fixed image not updating values when resizing

Relates to: https://github.com/xibosignage/xibo/issues/3443

### DIFF
--- a/lib/Entity/Media.php
+++ b/lib/Entity/Media.php
@@ -837,14 +837,16 @@ class Media implements \JsonSerializable
      * @param $md5
      * @param $fileSize
      */
-    public function release($md5, $fileSize)
+    public function release($md5, $fileSize, $height, $width)
     {
         // Update the MD5 and fileSize
-        $this->getStore()->update('UPDATE `media` SET md5 = :md5, fileSize = :fileSize, released = :released, modifiedDt = :modifiedDt WHERE mediaId = :mediaId', [
+        $this->getStore()->update('UPDATE `media` SET md5 = :md5, fileSize = :fileSize, released = :released, height = :height, width = :width, modifiedDt = :modifiedDt WHERE mediaId = :mediaId', [
             'fileSize' => $fileSize,
             'md5' => $md5,
             'released' => 1,
             'mediaId' => $this->mediaId,
+            'height' => $height,
+            'width' => $width,
             'modifiedDt' => Carbon::now()->format(DateFormatHelper::getSystemFormat())
         ]);
         $this->getLog()->debug('Updating image md5 and fileSize. MediaId '. $this->mediaId);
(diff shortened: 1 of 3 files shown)


lib/XTR/ImageProcessingTask.php
```diff
@@ -97,7 +97,9 @@ private function runImageProcessing()
             $count++;
 
             // Release image and save
-            $media->release(md5_file($filePath), filesize($filePath));
+            list($updatedImgWidth, $updatedImgHeight) = @getimagesize($filePath);
```
dasgarner, line 100: Have we explored if there is a more efficient way to do this? For example, does the image processing service know how it has resized the image? Could it return the new width/height for us to use?

dasgarner (commented):
I think this would work fine as-is, but I had a question to see if it could be more efficient (maybe not)

lib/XTR/ImageProcessingTask.php
dasgarner: Sure, but can we avoid calling `getimagesize` twice? If the resizeImage function knows what it resized it to, could that be returned instead?

mgbaybay: Hi again, Dan! I've made a few changes and would like to request a re-review, please. Thanks again :)

lib/Entity/Media.php
```diff
@@ -834,23 +834,21 @@ private function assessDimensions(): void
 
     /**
      * Release an image from image processing
-     * @param $md5
-     * @param $fileSize
+     * @param $img
      */
-    public function release($md5, $fileSize, $height, $width)
+    public function release($img)
```
dasgarner, line 839: Sorry if my comments have confused you. It is better to have explicit parameters as you had it before (you could even add types to them if you wanted).

lib/Service/ImageProcessingService.php
```diff
@@ -62,13 +62,21 @@ public function resizeImage($filePath, $width, $height)
             $img->resize($width, $height, function ($constraint)  {
                 $constraint->aspectRatio();
             });
+
+            // Get the updated height and width
```
dasgarner, line 66: Great 👍

dasgarner (changes requested):
The return from resizeImage is exactly what I hoped! That's great.

lib/Entity/Media.php
mgbaybay: Thanks for this :) I've added a new commit [here ](https://github.com/xibosignage/xibo-cms/pull/2602/commits/542d406e34ee2b22954a661dc8cb7a74531a672e). Requesting review again, please. Thank you!

dasgarner (approved):

ifarzana:
TEST: /Library END

PR #2606, opened 2024-07-04
Widget: Marquee up/down scrolling on android

dasgarner:
The marquee sets the contents of scroller to float:left, which on android causes the containing div height to go from 100% to 0 (as the things inside it are all floated). Fixing a pixel height to the content div makes the container for the scroller have a height.

fixes xibosignage/xibo#3388


rubenpingol-xibo (approved):

ifarzana:
TEST: /Layout END

PR #2607, opened 2024-07-04
Stocks and Currency stencils

maurofmferrao:
relates to xibosignageltd/xibo-private#769


modules/assets/template-thumbnails/currency/elements/stocks1.xcf
dasgarner, line 1: Should this file be here?

dasgarner (changes requested):

modules/assets/template-thumbnails/currency/elements/stocks1.xcf
maurofmferrao: Good catch, that was a file from Gimp to create the thumbnails. 🤦🏻 I'm going to remove it on another commit.

dasgarner (approved):

gracec20:
TEST: /Layout END

PR #2617, opened 2024-07-10
Dataset Add Column: Update help text to warn that column names with spaces cannot be used

nadzpogi:
relates to xibosignageltd/xibo-private#791


PeterMis (approved):

ifarzana:
TEST: /Library END

PR #2634, opened 2024-07-15
Schedule Edit: Preview button on Edit Event form requires Campaign features to be enabled

nadzpogi:
relates to xibosignageltd/xibo-private#539

- Did not need to provide a dedicated preview button for `layout` Event Type, as the current preview button already accommodates both campaign and layout previews.

### DIFF
--- a/views/schedule-form-add.twig
+++ b/views/schedule-form-add.twig
@@ -152,10 +152,12 @@
                         {{ forms.inputFullScreenSchedule('playlist', title, "", helpText, "playlist-control full-screen-control", playlistId, "", "", readonlySelect) }}
                         {{ forms.hidden('fullScreenCampaignId') }}
 
-                        <div class="form-group row preview-button-container">
-                            <div class="offset-md-2 col-md-10">
-                                <a id="previewButton" class="btn btn-success" target="_blank" data-url="{{ url_for("campaign.preview", {id: ':id'}) }}">{% trans "Preview" %} <span class="fa fa-tablet"></span></a>
-                                <small class="form-text text-muted">{% trans "Preview your selection in a new tab" %}</small>
+                        <div style="{% if not currentUser.featureEnabled('campaign.view') %}display: none;{% endif %}">
+                            <div class="form-group row preview-button-container">
+                                <div class="offset-md-2 col-md-10">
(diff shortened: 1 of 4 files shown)


views/schedule-form-add.twig
```diff
@@ -152,10 +152,12 @@
                         {{ forms.inputFullScreenSchedule('playlist', title, "", helpText, "playlist-control full-screen-control", playlistId, "", "", readonlySelect) }}
                         {{ forms.hidden('fullScreenCampaignId') }}
 
-                        <div class="form-group row preview-button-container">
-                            <div class="offset-md-2 col-md-10">
-                                <a id="previewButton" class="btn btn-success" target="_blank" data-url="{{ url_for("campaign.preview", {id: ':id'}) }}">{% trans "Preview" %} <span class="fa fa-tablet"></span></a>
-                                <small class="form-text text-muted">{% trans "Preview your selection in a new tab" %}</small>
+                        <div style="{% if not currentUser.featureEnabled('campaign.view') %}display: none;{% endif %}">
```
PeterMis, line 155: I'm not quite sure if that solves the problem.

Wasn't the intention to allow users with `layout.view` but without `campaign.view` feature to preview the Layout, if Layout eventType is selected?
With this change the button would just not be there without `campaign.view` feature.

I also *think* that the way `campaign.preview` route is validated makes it so it won't be available without `campaign.view` feature, as we check them one by one the first check will return `Feature not enabled`

PeterMis (changes requested):
I *think* we may need to adjust this a little bit.
Please see the comment below for details, if anything is unclear let me know

views/schedule-form-add.twig
nadzpogi: I see. I hadn't thought of it that way, and it looks like I missed one of the main points of the issue. This makes more sense. I'll modify it according to your suggestion. Thank you! 😁

nadzpogi:
The force-pushed commit only contains an amended commit message and title.

PeterMis (dismissed):

PeterMis (approved):

ifarzana:
TEST: /Schedule, /Layout END

PR #2637, opened 2024-07-15
Editor Property Panel: Keep same tab when moving between objects

maurofmferrao:
relates to xibosignageltd/xibo-private#775


dasgarner (approved):

gracec20:
TEST: /Layout END

gracec20:

PR #2639, opened 2024-07-16
DataSet: fix regression in filter clause matching for uniqueKey and data entry filtering

dasgarner:
relates to xibosignage/xibo#3457


PeterMis (approved):

ifarzana:
TEST: /Library END

PR #2642, opened 2024-07-17
Report: Add Session History

mgbaybay:
## Changes:
- Added `session` option and `session history table` in Sessions report 

Relates to: https://github.com/xibosignageltd/xibo-private/issues/747

### DIFF
--- a/lib/Report/SessionHistory.php
+++ b/lib/Report/SessionHistory.php
@@ -294,7 +294,7 @@ class SessionHistory implements ReportInterface
                 $rows,
                 count($rows),
             );
-        } else {
+        } else if ($type === 'debug') {
             $params = [
                 'fromDt' => $fromDt->format(DateFormatHelper::getSystemFormat()),
                 'toDt' => $toDt->format(DateFormatHelper::getSystemFormat()),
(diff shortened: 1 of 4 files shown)


lib/Report/SessionHistory.php
```diff
@@ -361,6 +361,87 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+            $sortOrder = $this->gridRenderSort($sanitizedParams);
+
+            if (is_array($sortOrder)) {
+                $sql .= ' ORDER BY ' . implode(',', $sortOrder);
+            }
+
+            $rows = [];
+            foreach ($this->store->select($sql, $params) as $row) {
+                $sessionRecord = $this->logFactory->createEmpty()->hydrate($row, ['htmlStringProperties' => ['message']]);
+                $endTime = isset($row['session_expiration']) ? date("Y-m-d H:i:s", $row['session_expiration']) : null;
```
dasgarner, line 412: There are some stray `"` here.

Also it would be better to use the `getSystemFormat` function in DateFormatHelper

lib/Report/SessionHistory.php
```diff
@@ -361,6 +361,87 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+             `session_history`.`userAgent`,
+             `session_history`.`ipAddress`,
+             `user`.`userName`,
+             `usertype`.`userType`,
+             `session`.`session_expiration`
+                FROM `session_history`
+                    LEFT OUTER JOIN `user` ON `user`.`userId` = `session_history`.`userId`
+                    LEFT OUTER JOIN `usertype` ON `usertype`.`userTypeId` = `user`.`userTypeId`
+                    LEFT OUTER JOIN `session` ON `session`.`session_data` 
+                        LIKE CONCAT("sessionHistoryId|i:", session_history.sessionId, "%")
```
dasgarner, line 388: Needing to do this sort of join points to a data model problem, it also means that as the sessions are removed from the session table, we lose the `session_expiration` and therefore the end date and duration from the report.

Ideally we would have a `lastUsedTime` field in the session_history table which is updated each time an existing session is touched, and would then accurately represent the time that session was active.

That is outside the scope of the initial issue, but please you could investigate adding this column to the table and making sure it gets updated when the session is accessed?

dasgarner (changes requested):
The report looks good! There are some structure changes I think we need to look at.

dasgarner (approved):

ifarzana (approved):

ifarzana:
TEST: /Reporting END

PR #2644, opened 2024-07-17
Add user: Select/Deselect all folders

maurofmferrao:
relates to xibosignageltd/xibo-private#782


gracec20:
TEST: /Administration END

PR #2648, opened 2024-07-22
Real-time data: provide the data connector JS via Connectors

nadzpogi:
relates to xibosignageltd/xibo-private#753

### DIFF
--- a/lib/Event/DataConnectorSourceRequestEvent.php
+++ b/lib/Event/DataConnectorSourceRequestEvent.php
@@ -37,9 +37,17 @@ class DataConnectorSourceRequestEvent extends Event implements DataConnectorSour
     /**
      * @var array
      */
-    private $dataConnectorSources = [
-        ['id' => 'user_defined', 'name' => 'User Defined JavaScript']
-    ];
+    private $dataConnectorSources = [];
+
+    /**
+     * Initializes the dataConnectorSources with default value.
+     */
+    public function __construct() {
+        $this->dataConnectorSources[] = [
+            'id' => 'user_defined',
+            'name' => __('User-Defined JavaScript')
+        ];
+    }
 
     /**
      * @inheritDoc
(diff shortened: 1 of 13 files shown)


lib/Event/DataConnectorSourceRequestEvent.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+class DataConnectorSourceRequestEvent extends Event implements DataConnectorSourceProviderInterface
+{
+    public static $NAME = 'data.connector.source.request';
+
+    /**
+     * @var array
+     */
+    private $dataConnectorSources = [
+        ['id' => 'user_defined', 'name' => 'User Defined JavaScript']
```
dasgarner, line 41: Can we translate the name?

views/dataset-data-connector-page.twig
```diff
@@ -67,7 +67,7 @@
                                 </div>
                             </div>
 
-                            {{ forms.button("Save"|trans, "submit", null, null, null, "btn-success") }}
+                            {{ forms.button("Save"|trans, "submit", null, null, null, "btn-success " ~ (dataSet.dataConnectorSource != 'user_defined' ? 'disabled' : ''), disabled) }}
```
dasgarner, line 70: Should we hide it as well?

dasgarner (changes requested):
This is great - I only have a very minor comment regarding translating.

lib/Event/DataConnectorSourceRequestEvent.php
nadzpogi: Noted. Sorry, the translation slipped my mind. 😅

views/dataset-data-connector-page.twig
nadzpogi: Oh, it's already within the div that gets hidden if the source is not user-defined, so there's no need to explicitly hide the button within that div.

But I forgot to remove the disabled parameter (it was a test to see if it works), so I'll remove that one.
` ''), disabled) }}`

dasgarner (approved):

PR #2652, opened 2024-07-25
Permissions: only update permission record in the DB if necessary 

dasgarner:
This PR tries to reduce the number of unnecessary SQL updates which would run when saving permissions if there are no changes to the existing records.

Inserting new records is unaffected. Inserting existing records only happens if a property has changed.

relates to xibosignageltd/xibo-private#797

### DIFF
--- a/lib/Entity/Permission.php
+++ b/lib/Entity/Permission.php
@@ -1,8 +1,8 @@
 <?php
 /*
- * Copyright (c) 2022 Xibo Signage Ltd
+ * Copyright (C) 2024 Xibo Signage Ltd
  *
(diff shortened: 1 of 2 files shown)


mgbaybay (approved):

gracec20:
TEST: /Layout END

gracec20:
TEST: /Layout/Editor/layout_editor_dataset.cy.js, /Layout/Editor/layout_editor_rss_ticker.cy.js END

gracec20:
TEST: /Layout/Editor/layout_editor_rss_ticker.cy.js END

PR #2653, opened 2024-07-25
Media: when deleting a media record reinstate its parent if it has one

dasgarner:
fixes to xibosignage/xibo#3454


PeterMis (approved):

ifarzana:
TEST: /Library END

PR #2654, opened 2024-07-25
XMDS: use a sub-query to filter out draft layouts before outer join

dasgarner:
This PR updates the query for returning the schedule to XMDS so that layouts which have an associated draft are not returned twice in the schedule.

fixes to xibosignage/xibo#3437


mgbaybay (approved):

ifarzana:
TEST: /Schedule END

PR #2659, opened 2024-07-29
Reporting: Add Availability and Grouping Option

mgbaybay:
## Changes:
Part 1:
- Implement `Group By` filter option 
- Added `Availability Percentage` column
- Implement dynamic table columns display

Relates to: https://github.com/xibosignage/xibo/issues/3357

### DIFF
--- a/lib/Report/TimeDisconnectedSummary.php
+++ b/lib/Report/TimeDisconnectedSummary.php
@@ -342,21 +342,26 @@ class TimeDisconnectedSummary implements ReportInterface
         $displayBody = 'FROM `display` ';
 
         if ($groupBy === 'displayGroup') {
-            $displaySelect .= ', displaygroup.displayGroup, displaygroup.displayGroupId ';
+            $displaySelect .= ', displaydg.displayGroup, displaydg.displayGroupId ';
         }
 
-        if ($tags != '' || $groupBy === 'displayGroup') {
+        if ($tags != '') {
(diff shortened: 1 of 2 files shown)


lib/Report/TimeDisconnectedSummary.php
```diff
@@ -340,12 +341,21 @@ public function getResults(SanitizerInterface $sanitizedParams)
...
+
+        if ($tags != '' || $groupBy === 'displayGroup') {
             $displayBody .= 'INNER JOIN `lkdisplaydg`
                         ON lkdisplaydg.DisplayID = display.displayid
                      INNER JOIN `displaygroup`
-                        ON displaygroup.displaygroupId = lkdisplaydg.displaygroupId
-                         AND `displaygroup`.isDisplaySpecific = 1 ';
+                        ON displaygroup.displaygroupId = lkdisplaydg.displaygroupId ';
+
+            if ($groupBy === 'displayGroup') {
```
dasgarner, line 354: If I filter by a tag which happens to be assigned to several displays, I will get a those displays as my only results. If I then group by display group without changing my tag filter, will I get other displays included in the results because of this clause?

dasgarner (changes requested):
I'm not sure this is right yet, lets have a chat about it.

dasgarner (approved):
This looks good to me!

gracec20:
TEST: /Reporting END

PR #2662, opened 2024-07-30
Dataset: Importing a CSV file with an empty line fails via remote dataset and uploading the file

nadzpogi:
relates to xibosignage/xibo#3456

- `ini_set('auto_detect_line_endings', true);` is deprecated.

### DIFF
--- a/lib/Helper/DataSetUploadHandler.php
+++ b/lib/Helper/DataSetUploadHandler.php
@@ -97,12 +97,9 @@ class DataSetUploadHandler extends BlueImpUploadHandler
             while (($data = fgetcsv($handle)) !== FALSE ) {
                 $i++;
 
-                // Filter out rows that are entirely empty
-                $filteredData = array_filter($data, function($row) {
-                    // Check if the row is empty (all elements are empty or null)
-                    return array_filter($row, function($value) {
-                        return !empty($value);
-                    });
+                // remove any elements that doesn't contain any value from the array
+                $filteredData = array_filter($data, function($value) {
+                    return !empty($value);
                 });
 
                 // Skip empty lines without any delimiters or data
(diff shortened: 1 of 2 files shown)


PeterMis (changes requested):
From the issue I'd assume this should be about remote dataSet.

I don't think that would fix remote dataSet import - shouldn't we look in `DataSetFactory -> callRemoteService()` or `processCsvEntries()` etc  instead ?

DataSetUploadHandler is for manual importing data to dataset.

I think that would fix manual csv import to dataSet for csv that contains empty rows ( just a small comment on that below ) - manual import also has an issue whereby only `,` separator works -  but perhaps that's outside of the scope of this issue

lib/Helper/DataSetUploadHandler.php
```diff
@@ -90,16 +90,26 @@ protected function handleFormData($file, $index)
...
-
             $firstRow = true;
             $i = 0;
 
             $handle = fopen($controller->getConfig()->getSetting('LIBRARY_LOCATION') . 'temp/' . $fileName, 'r');
             while (($data = fgetcsv($handle)) !== FALSE ) {
                 $i++;
 
+                // Filter out rows that are entirely empty
+                $filteredData = array_filter($data, function($row) {
```
PeterMis, line 101: Are we sure this should be double `array_filter` ?

and not just
```
$filteredData = array_filter($data, function($value) {
    return !empty($value);
});
```

I think the `$data` here is already an array containing single row data

Should this filter also be after the firstRow check? - no need to filter first row if its going to be ignored anyway

PeterMis (changes requested):
I think this is almost there!

The remote dataSet part looks good to me

The manual csv import I think was more correct previously, please see comments below

lib/Helper/DataSetUploadHandler.php
nadzpogi: Your right, I was trying to remember why I did that in the first place but i can't seem to fully remember it. 😅

I think I may have done that to avoid something like `,,,,` or basically lines with no values but has multiple delimiters. And I might have been confused thinking that the `$data` contains all the rows from the CSV when in fact, it only contains a single row per iteration.

Thank you for pointing that out to me.

PeterMis (approved):

ifarzana:
TEST: /Administration END

PR #2663, opened 2024-07-30
Upgrade Node.js and npm packages

maurofmferrao:
relates to xibosignageltd/xibo-private#792


package.json
```diff
@@ -86,42 +80,38 @@
...
-    "monaco-editor": "^0.46.0",
-    "moveable": "^0.48.0",
-    "pdfjs-dist": "2.4.456",
-    "persian-date": "^0.3.1-b",
+    "moment": "~2.30.1",
+    "moment-jalaali": "~0.10.1",
+    "moment-timezone": "^0.5.45",
+    "monaco-editor": "^0.50.0",
+    "moveable": "^0.53.0",
+    "pdfjs-dist": "^4.5.136",
```
dasgarner, line 107: When we looked at a v4 upgrade before we ran into errors because the accompanying asset wasn't the same `/modules/assets/pdfjs/pdf.worker.js`. Have you managed to cater for that?

dasgarner (changes requested):
If we're upgrading node, we are also going to want to upgrade the node version in the Dockerfile's used to build the containers.

This should be simple by changing `FROM node:12 AS webpack` in the 3 Dockerfiles

Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN find -type d -name '.git' -exec rm -r {} + && \
 
 # Stage 2
 # Run webpack
-FROM node:12 AS webpack
+FROM node:lts AS webpack
```
dasgarner, line 28: We should use a specific version of node so that it doesn't change without our awareness.

dasgarner (changes requested):

dasgarner (approved):

PR #2664, opened 2024-07-30
Exchange: layouts added from exchange should not have the imported tag

dasgarner:
This is because they are applied templates and the parent tags win.
xibosignageltd/xibo-private#811

This PR makes sure that the "imported" tag is not applied when applying a template from a provider (in this case the Xibo Exchange).


mgbaybay (approved):

gracec20:
TEST: /Layout/Editor/layout_editor_mastodon.cy.js END

PR #2665, opened 2024-07-31
DataSet: trying to import without any columns logs an unhelpful error

nadzpogi:
relates to xibosignage/xibo#3458

### DIFF
--- a/lib/Helper/DataSetUploadHandler.php
+++ b/lib/Helper/DataSetUploadHandler.php
@@ -62,6 +62,12 @@ class DataSetUploadHandler extends BlueImpUploadHandler
                 throw new AccessDeniedException();
             }
 
+            // Check if there are any columns defined in the dataset
+            if (count($dataSet->getColumn()) === 0) {
+                $controller->getLog()->error('Import failed: No columns defined in the dataset.');
+                throw new InvalidArgumentException('Import failed: No columns defined in the dataset.');
+            }
+
             // We are allowed to edit - pull all required parameters from the request object
             $overwrite = $sanitizer->getCheckbox('overwrite');
             $ignoreFirstRow = $sanitizer->getCheckbox('ignorefirstrow');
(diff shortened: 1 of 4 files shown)


PeterMis (changes requested):
I think this looks fine for manual csv import, however looking at the issue it should also cover remote datasets.

ie processing json/csv from remote dataSet source should also check for no columns - in this case perhaps also conditionally for no `remote` columns.

The message on the form could also be adjusted for remote dataSet to check the remote columns - the message for no columns at all that you have is fine, but in case of the remote dataSet we are more interested in remote columns instead.

lib/Helper/DataSetUploadHandler.php
```diff
@@ -62,6 +62,20 @@ protected function handleFormData($file, $index)
                 throw new AccessDeniedException();
             }
 
+            // Get all columns
+            $columns = $dataSet->getColumn();
+
+            // Filter columns where dataSetColumnType is "Value"
+            $filteredColumns = array_filter($columns, function ($column) {
+                return $column->dataSetColumnType == 'Value';
```
PeterMis, line 70: Might be better to rely on `dataSetColumnTypeId` instead? ( result should be the same )

lib/Helper/DataSetUploadHandler.php
```diff
@@ -62,6 +62,20 @@ protected function handleFormData($file, $index)
...
+            });
+
+            // Check if there are any value columns defined in the dataset
+            if (count($filteredColumns) === 0) {
+                $controller->getLog()->error('Import failed: No value columns defined in the dataset.');
+                throw new InvalidArgumentException('Import failed: No value columns defined in the dataset.');
```
PeterMis, line 76: Should we translate the exception message?

lib/XTR/RemoteDataSetFetchTask.php
```diff
@@ -99,6 +99,20 @@ public function run()
                     continue;
                 }
 
+                // Get all columns
+                $columns = $dataSet->getColumn();
+
+                // Filter columns where dataSetColumnType is "Remote"
+                $filteredColumns = array_filter($columns, function ($column) {
+                    return $column->dataSetColumnType == 'Remote';
```
PeterMis, line 107: Similar to previous comment, perhaps rely on `dataSetColumnTypeId` ?

PeterMis (commented):
Looks good to me, just a very minor comments below.

PeterMis (approved):

ifarzana:
TEST: /Administration END

PR #2679, opened 2024-08-02
Reports : API Requests new report type, various adjustments.

PeterMis:
relates to xibosignage/xibo#3362

- Requests : New API requests report type that does not join on log/audit tables
- Various adjustments to the report, forms etc to suit the new report type
- Change the url added to database to correctly include route parameters ( previously things like `{id}` would be saved)
- Separate sql connection and commit to ensure we get the error requests in the table
- Added small new Factory and Entity for API requests


dasgarner (approved):

ifarzana:
TEST: /Reporting END

PR #2683, opened 2024-08-05
Content-Length header check with newer Apache versions

mariosteinbacher:
Apache 2.4.44 and newer removed the Content-Length header.

So third party media via the ticker module for example can't be loaded, and throws an exception from guzzlehttp 'An error was encountered during the on_headers event'

this is just a quick and dirty fix. maybe there is a better solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tests/http-client.private.env.json
 
 # Hide module/template translate file
 locale/moduletranslate.php
+/nbproject/private/
--- a/lib/Factory/MediaFactory.php
+++ b/lib/Factory/MediaFactory.php
@@ -271,12 +271,17 @@ class MediaFactory extends BaseFactory
                                 $this->getLog()->debug('processDownloads: successful, headers = '
                                     . var_export($response->getHeaders(), true));
 
-                                // Get the content length
-                                $contentLength = $response->getHeaderLine('Content-Length');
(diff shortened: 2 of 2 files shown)


dasgarner:
Thank you for your contribution - if I understand the code correctly this will disable the Content-Length header check if request we're trying to download doesn't have it?

If that is the case, then unfortunately we wouldn't be able to merge this because the check is in place to prevent this sort of request from being accepted. It is possible to create a denial of service situation by effectively downloading never ending streams and tying up the download process indefinitely.

I've had a brief read through the associated links in your commit, and it seems the Apache change has caused quite a kerfuffle... the suggested fix seems to be to set `ap_trust_cgilike_cl` - is that something which would work as an alternative?

mariosteinbacher:
Hello Dan!
I am as surprised as you when I have read that apache removed this header.
I am happy with any solution. 
so the solution would be to set this variable on the third party server and the php script must add the Content-Length header. As far as I have understood it.
And maybe I have missed the section where the content-length is checked against the actual file size.

dasgarner:
Hello! Xibo checks the Content-Length header provided against the max file size allowed on the CMS, to make sure that we're not attempting to download files which are larger than allowed. Xibo doesn't check the actual file size.

I think the solution is to apply `ap_trust_cgilike_cl` to the source server as you've suggested.

PR #2686, opened 2024-08-06
Schedule Criteria: CMS API + XMR update to displays

nadzpogi:
relates to xibosignageltd/xibo-private#807

### DIFF
--- a/lib/Controller/DisplayGroup.php
+++ b/lib/Controller/DisplayGroup.php
@@ -21,6 +21,7 @@
  */
 namespace Xibo\Controller;
 
+use Psr\Http\Message\ResponseInterface;
 use Slim\Http\Response as Response;
 use Slim\Http\ServerRequest as Request;
 use Xibo\Entity\Display;
@@ -36,6 +37,7 @@ use Xibo\Factory\ModuleFactory;
 use Xibo\Factory\TagFactory;
 use Xibo\Service\PlayerActionServiceInterface;
 use Xibo\Support\Exception\AccessDeniedException;
+use Xibo\Support\Exception\ControllerNotImplemented;
 use Xibo\Support\Exception\GeneralException;
 use Xibo\Support\Exception\InvalidArgumentException;
 use Xibo\Support\Exception\NotFoundException;
@@ -43,7 +45,9 @@ use Xibo\XMR\ChangeLayoutAction;
 use Xibo\XMR\CollectNowAction;
 use Xibo\XMR\CommandAction;
 use Xibo\XMR\OverlayLayoutAction;
+use Xibo\XMR\PlayerActionException;
 use Xibo\XMR\RevertToSchedule;
+use Xibo\XMR\ScheduleCriteriaUpdateAction;
 use Xibo\XMR\TriggerWebhookAction;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,90 @@ public function triggerWebhook(Request $request, Response $response, $id)
             'id' => $displayGroup->displayGroupId
         ]);
 
+        return $this->render($request, $response);
+    }
+
+    /**
+     * @param Request $request
```
dasgarner, line 2880: Please can the swagger doc be above these?

lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,90 @@ public function triggerWebhook(Request $request, Response $response, $id)
...
+     * @param Response $response
+     * @param null $displayGroupId
```
dasgarner, line 2882: null or int?

lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,90 @@ public function triggerWebhook(Request $request, Response $response, $id)
...
+     * @return ResponseInterface|Response
+     * @throws ControllerNotImplemented
+     * @throws GeneralException
+     * @throws NotFoundException
+     * @throws PlayerActionException
+     * @SWG\Post(
+     *   path="/displaygroup/criteria[/{displayGroupId}]",
+     *   operationId="ScheduleCriteriaUpdateActionPushCriteriaUpdate",
```
dasgarner, line 2890: I think we can shorten this a bit, perhaps `ScheduleCriteriaUpdate` ?

lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,90 @@ public function triggerWebhook(Request $request, Response $response, $id)
...
+            $displayGroup = $this->displayFactory->getByDisplayGroupId($displayGroupId);
+        } else {
+            // fetch all displays regardless of Display Group
+            $displayGroup = $this->displayFactory->query();
+        }
+
+        // Create and send the player action
+        $this->playerAction->sendAction(
+            $displayGroup,
+            (new ScheduleCriteriaUpdateAction())->setCriteriaUpdates($criteriaUpdates)
```
dasgarner, line 2949: We should sanitize the input in the array before sending it, to make sure someone doesn't try to send unwanted parameters to the player

lib/XMR/ScheduleCriteriaUpdateAction.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return $this
+     * @throws PlayerActionException
+     */
+    public function setCriteriaUpdates(array $criteriaUpdates)
+    {
+        foreach ($criteriaUpdates as $criteria) {
+            // Ensure each criterion has metric, value, and ttl
+            if (empty($criteria['metric']) || empty($criteria['value']) || empty($criteria['ttl'])) {
+                // Throw an exception if any of the required fields are missing or empty
+                throw new PlayerActionException(__('Invalid criteria format. Metric, value, and ttl must all be present and not empty.'));
```
dasgarner, line 53: This line fails phpcs check as it is too long.

dasgarner: It is good to have this validation here, but we could also validate in the Controller

dasgarner (changes requested):
I think the logic you have is good. I have a few suggestions and changes requested.

lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,120 @@ public function triggerWebhook(Request $request, Response $response, $id)
...
+        foreach ($criteriaUpdates as $criteria) {
+            $criteriaSanitizer = $this->getSanitizer($criteria);
+
+            // Sanitize and retrieve the metric, value, and ttl
+            $metric = $criteriaSanitizer->getString('metric');
+            $value = $criteriaSanitizer->getString('value');
+            $ttl = $criteriaSanitizer->getInt('ttl');
+
+            // Ensure each criterion has metric, value, and ttl
+            if (empty($metric) || empty($value) || empty($ttl)) {
```
dasgarner, line 2953: empty will return true for `0`, which I think we want to be a valid value for TTL (i.e. unlimited)

lib/Controller/DisplayGroup.php
```diff
@@ -2869,6 +2873,120 @@ public function triggerWebhook(Request $request, Response $response, $id)
...
+
+            // Add sanitized criteria
+            $sanitizedCriteriaUpdates[] = [
+                'metric' => $metric,
+                'value' => $value,
+                'ttl' => $ttl
+            ];
+        }
+
+        if ($displayGroupId !== null) {
```
dasgarner, line 2968: Now that I see this, it does worry me that this could potentially be 1000s of displays. Perhaps it would be more sensible to make that parameter not optional after all?

dasgarner (changes requested):

lib/Controller/DisplayGroup.php
nadzpogi: I think your right now that I think about it. Then, I'll make it a required parameter instead of it being optional

dasgarner (approved):

ifarzana:
TEST: /Display END

PR #2691, opened 2024-08-07
Shrunken login page

maurofmferrao:
relates to xibosignageltd/xibo-private#827


ifarzana:
TEST: login.cy.js END

PR #2706, opened 2024-08-13
Modules: Disable library media preview for generic file

mgbaybay:
## Changes
- Modules: Disable library media preview for generic file

Relates to: https://github.com/xibosignageltd/xibo-private/issues/802


dasgarner (changes requested):
This looks great - one final change I think we could make is to disable the "preview enabled" option on the module settings form, so that it is clear it isn't supported for any module with allowPreview set to 0

dasgarner (approved):

ifarzana:
TEST: /Library, /Administration END

PR #2707, opened 2024-08-14
Bugfix/android player not rendering elements

rubenpingol-xibo:
relates to xibosignageltd/xibo-private#810

### DIFF
--- a/modules/layout-renderer.twig
+++ b/modules/layout-renderer.twig
@@ -56,7 +56,7 @@
             {% set xlrOptions = previewOptions|merge({inPreview: true}) %}
             (function($){
                 $(document).ready(function(){
-                    let xiboLayoutRenderer = new XiboLayoutRenderer({{ layoutObj|json_encode()|raw }}, {{ xlrOptions|json_encode()|raw }});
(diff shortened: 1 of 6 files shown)


modules/layout-renderer.twig
```diff
@@ -31,14 +31,10 @@
         <meta name="public-path" content="{{ theme.rootUri() }}"/>
         
         <link rel="shortcut icon" href="{{ theme.uri("img/favicon.ico") }}" />
-        <link rel="stylesheet" type="text/css" href="{{ theme.rootUri() }}dist/preview/html-preview.css" />
-        <link rel="stylesheet" type="text/css" href="{{ theme.uri("css/html-preview.css") }}" />
```
dasgarner, line 35: This file is used during the white label process to replace the Xibo logo with the white label logo equivalent, which I guess is something we need to handle in XLR?

dasgarner (commented):
I have a question which might result in a follow on issue. For now, this is good to test and see if it solves the main issue.

dasgarner (dismissed):

modules/layout-renderer.twig
rubenpingol-xibo: Yes, I guess we can have this handled by XLR instead. I just need to know what data determines if it's white label or not.

dasgarner (approved):

ifarzana:
TEST: /Layout END

PR #2710, opened 2024-08-21
Ad Campaigns: unable to set a start/end date

maurofmferrao:
relates to xibosignageltd/xibo-private#835

### DIFF
--- a/ui/src/core/xibo-cms.js
+++ b/ui/src/core/xibo-cms.js
@@ -276,7 +276,18 @@ function XiboInitialise(scope, options) {
     // NOTE: The validation plugin does not like binding to multiple forms at once.
     $(scope + ' .XiboForm').validate({
         submitHandler: XiboFormSubmit,
+        // Ignore the date picker helpers
+        ignore: '.datePickerHelper',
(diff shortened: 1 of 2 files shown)


dasgarner:
I'm not clear how the proposed changes fix the issue? I can see that the formatting of the error message will be better, but I don't see a change which prevents the error from showing?

maurofmferrao:
> I'm not clear how the proposed changes fix the issue? I can see that the formatting of the error message will be better, but I don't see a change which prevents the error from showing?

The error was a hard one to catch, so Flatpckr creates a aux input for the formatted date and that was the one throwing the error on validation, not the input that gets submitted with the form. By using the ignore on the validation method [here](https://github.com/xibosignage/xibo-cms/pull/2710/files#diff-4d00470992e5172618a886d5e9a249d5d5efeb38fc14825af2cd23c37ef2b9cfR216) for the date picker helper, we prevent that error to happening. At least in theory ( it was working locally 🤞🏻).

dasgarner (approved):

ifarzana:
TEST: /Campaign END

PR #2711, opened 2024-08-22
Playlists: Fix for disable manual changes to media assignments on dynamic playlists

mgbaybay:
## Changes
- Disable manual changes to media assignments on dynamic playlists

Relates to: https://github.com/xibosignage/xibo/issues/3480

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -1168,10 +1168,12 @@ class Library extends Base
             $folderId = $this->getUser()->homeFolderId;
         }
 
-        $playlist = $this->playlistFactory->getById($parsedBody->getInt('playlistId'));
+        if ($parsedBody->getInt('playlistId') !== null) {
+            $playlist = $this->playlistFactory->getById($parsedBody->getInt('playlistId'));
 
-        if ($playlist->isDynamic === 1) {
-            throw new InvalidArgumentException(__('This Playlist is dynamically managed so cannot accept manual assignments.'), 'isDynamic');
+            if ($playlist->isDynamic === 1) {
+                throw new InvalidArgumentException(__('This Playlist is dynamically managed so cannot accept manual assignments.'), 'isDynamic');
+            }
         }
 
         $options = array_merge([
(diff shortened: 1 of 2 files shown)


lib/Controller/Library.php
```diff
@@ -1168,6 +1168,12 @@ public function add(Request $request, Response $response)
             $folderId = $this->getUser()->homeFolderId;
         }
 
+        $playlist = $this->playlistFactory->getById($parsedBody->getInt('playlistId'));
```
dasgarner, line 1171: If we're uploading media not on a playlist, then this parameter would be empty and I think throw a not found?

lib/Controller/Widget.php
```diff
@@ -1638,6 +1666,10 @@ public function widgetSetRegion(Request $request, Response $response, $id)
             throw new InvalidArgumentException(__('This Layout is not a Draft, please checkout.'), 'layoutId');
         }
 
+        if ($playlist->isDynamic === 1) {
```
dasgarner, line 1669: Dynamic playlists are never region specific (they are always created on the playlist page, not the layout editor), so I think this can't be triggered)

lib/Controller/Widget.php
```diff
@@ -1720,6 +1752,10 @@ public function saveElements(Request $request, Response $response, $id)
...
```
dasgarner, line 1755: Dynamic playlists (as per prior comments can only be created from the Playlist page and therefore `isRegionPlaylist` = false, so we can't get here.

dasgarner (changes requested):

lib/Controller/Library.php
mgbaybay: @dasgarner Thank you for the review! I've updated the logic to address this, please see latest [commit ](https://github.com/xibosignage/xibo-cms/pull/2711/commits/da53951c8848138bdba4936f28bcd1cf28101c3e). Thanks again!

dasgarner (approved):

lib/Controller/Widget.php
mgbaybay: Also removed this based on this [commit ](https://github.com/xibosignage/xibo-cms/pull/2711/commits/da53951c8848138bdba4936f28bcd1cf28101c3e) :)

ifarzana:
TEST: /Library END

PR #2712, opened 2024-08-22
Layout Editor: we should not show the playlist editor for dynamic playlists

maurofmferrao:
relates to xibosignageltd/xibo-private#838

### DIFF
--- a/ui/src/layout-editor/main.js
+++ b/ui/src/layout-editor/main.js
@@ -3609,6 +3609,13 @@ lD.openContextMenu = function(obj, position = {x: 0, y: 0}) {
     layoutObject.playlists.widgets.length > 0
   );
 
+  // Check if it's a dynamic playlist
+  layoutObject.isDynamicPlaylist =
+    (
+      layoutObject.isPlaylist &&
+      $(obj).find('.dynamic-playlist-icon').length === 1
+    );
+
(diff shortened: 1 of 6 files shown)


ui/src/layout-editor/main.js
```diff
@@ -3609,6 +3609,13 @@ lD.openContextMenu = function(obj, position = {x: 0, y: 0}) {
     layoutObject.playlists.widgets.length > 0
   );
 
+  // Check if it's a dynamic playlist
+  layoutObject.isDynamicPlaylist =
+    (
+      layoutObject.isPlaylist &&
+      $(obj).find('.dynamic-playlist-icon').length === 1
```
dasgarner, line 3616: Would it be better to add something more "permanent" to this DOM element? Perhaps a data attribute or similar. I worry that if we change the icon for some reason we create a hard-to-find bug?

maurofmferrao: Yeah, you're right. I will add a class to the main object and also for the convert playlist button on top. Thanks!

dasgarner (approved):

dasgarner (approved):

ifarzana:
TEST: /Layout END

ifarzana:

PR #2714, opened 2024-08-23
Display: Fix for updating detail of an alert when submitting the end

mgbaybay:
## Changes
- Display: Update detail of an alert when submitting the end

Relates to: https://github.com/xibosignage/xibo/issues/3478

### DIFF
--- a/lib/Entity/DisplayEvent.php
+++ b/lib/Entity/DisplayEvent.php
@@ -167,13 +167,13 @@ class DisplayEvent implements \JsonSerializable
 
         // When updating the event end, concatenate the end message to the current message
         $this->getStore()->update(
-            'UPDATE `displayevent` SET 
+            "UPDATE `displayevent` SET 
                       `end` = :toDt, 
-                      `detail` = CONCAT_WS(". ", NULLIF(CONCAT_WS(".", NULLIF(`detail`, "")), ""), :detail)
+                      `detail` = CONCAT_WS('. ', NULLIF(`detail`, ''), :detail)
                       WHERE displayId = :displayId 
                         AND `end` IS NULL 
                         AND eventTypeId = :eventTypeId
-                        AND refId = :refId',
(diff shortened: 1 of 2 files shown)


lib/Xmds/Soap.php
```diff
@@ -3202,7 +3202,7 @@ private function createDisplayAlert(\DomElement $alertNode)
             $eventTypeId = $displayEvent->getEventIdFromString($eventType);
             empty($refId)
                 ? $displayEvent->eventEnd($this->display->displayId, $eventTypeId, $date)
-                : $displayEvent->eventEndByReference($this->display->displayId, $eventTypeId, $refId);
+                : $displayEvent->eventEndByReference($this->display->displayId, $eventTypeId, $refId, $detail);
```
dasgarner, line 3205: Is it possible to also do it if there is no reference?

lib/Entity/DisplayEvent.php
```diff
@@ -165,8 +165,11 @@ public function eventEndByReference(int $displayId, int $eventTypeId, int $refId
             )
         );
 
+        // When updating the event end, concatenate the end message to the current message
         $this->getStore()->update(
-            'UPDATE `displayevent` SET `end` = :toDt
+            'UPDATE `displayevent` SET 
+                      `end` = :toDt, 
+                      `detail` = CONCAT_WS(". ", NULLIF(CONCAT_WS(".", NULLIF(`detail`, "")), ""), :detail)
```
dasgarner, line 172: Can we use single quotes for consistency?

dasgarner: Also can you explain why the double concat_ws in a comment so its easy to refer to later?

dasgarner (changes requested):

dasgarner (approved):

ifarzana:
TEST: /Display END

PR #2718, opened 2024-08-27
Display: Allow limited view access to non-destructive edit-only features

mgbaybay:
## Changes
- Display: Allow limited view access to non-destructive edit-only features

Relates to: https://github.com/xibosignage/xibo/issues/3479


lib/Controller/DisplayGroup.php
```diff
@@ -460,11 +460,16 @@ public function grid(Request $request, Response $response)
                 ];
             }
 
-            if ($this->getUser()->featureEnabled('displaygroup.modify')
-                && $this->getUser()->checkEditable($group)
+            // Check if limited view access is allowed
+            if (($this->getUser()->featureEnabled('displaygroup.modify') && $this->getUser()->checkEditable($group))
+                || $this->getUser()->featureEnabled('displays.limitedView')
```
dasgarner, line 465: To do these things at the group level feels like something beyond what you might expect for the displays.limitedView feature, however we could add a second new feature to allow this at the group level?

dasgarner (changes requested):
Much more inline with the intension! great.

I have one comment regarding the display group access

lib/Controller/DisplayGroup.php
mgbaybay: Thanks, @dasgarner :) Added a new feature for the group level in this [commit  ](https://github.com/xibosignage/xibo-cms/pull/2718/commits/30095c0a731a2a1a3c645dbbaf36e317764dc4ab)

dasgarner (approved):

ifarzana:
TEST: /Administration END

ifarzana:
TEST: /Administration END

PR #2726, opened 2024-09-02
Fix validation for hidden inputs

maurofmferrao:
relates to xibosignageltd/xibo-private#835


ifarzana:
TEST: /Administration END

PR #2729, opened 2024-09-06
Pixabay: Get 1080p videos by default instead of 4k

mgbaybay:
## Changes
BE:
- Get 1080p videos by default instead of 4k

Relates to: https://github.com/xibosignageltd/xibo-private/issues/840

### DIFF
--- a/lib/Connector/PixabayConnector.php
+++ b/lib/Connector/PixabayConnector.php
@@ -193,30 +193,25 @@ class PixabayConnector implements ConnectorInterface
                     $searchResult->thumbnail = $result->videos->tiny->url;
                     $searchResult->duration = $result->duration;
 
-                    if ($this->isDefaultVideoResolution($result, 'large')) {
-                        $searchResult->download = $result->videos->large->url;
-                        $searchResult->width = $result->videos->large->width;
-                        $searchResult->height = $result->videos->large->height;
-                        $searchResult->fileSize = $result->videos->large->size;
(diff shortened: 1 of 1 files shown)


lib/Connector/PixabayConnector.php
```diff
@@ -221,10 +226,6 @@ public function onLibraryProvider(LibraryProviderEvent $event)
                             $result->picture_id,
                             'https://i.vimeocdn.com/video/pictureId_960x540.png'
                         );
-                    } else {
```
dasgarner, line 224: I am not clear why we moved from using the medium thumbnail with tiny thumbnail fallback, to using the thumbnail for each size of video. Aren't the "large" ones going to be too big for our usage of a thumbnail?

mgbaybay: Thanks for this! I was thinking that there must be some sort of consistency (ie if we use large video, then we should use large thumbnail as well) haha 😄 but you're right, some thumbnails might be too large for our use case. 

Reverted the changes :)

dasgarner (approved):

ifarzana (approved):
Approving to build container

ifarzana (approved):
Approving to build container

ifarzana:
TEST: /Library END

ifarzana:
The error was expected due to a bug. We will run another test to ensure the PR gets the green light.

ifarzana:
TEST: login.cy.js END

PR #2733, opened 2024-09-11
Auditing: adding/deleting a widget should log the mediaId being modified if there is one

nadzpogi:
relates to xibosignageltd/xibo-private#856

### DIFF
--- a/lib/Entity/Widget.php
+++ b/lib/Entity/Widget.php
@@ -1104,8 +1104,12 @@ class Widget implements \JsonSerializable
             ['widgetId' => $this->widgetId]
         );
         
-        // Unlink Media
         $this->mediaIds = [];
+
+        // initialize media Ids to unlink
+        $mediaIdsToUnlink = $this->getMediaIdsToUnlink();
+
+        // Unlink Media
         $this->unlinkMedia();
 
         // Delete any fallback data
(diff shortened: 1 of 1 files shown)


lib/Entity/Widget.php
```diff
@@ -1269,6 +1269,9 @@ private function linkMedia()
                 'mediaId' => $mediaId,
                 'mediaId2' => $mediaId
             ));
+
+            // audit the media being added
+            $this->getLog()->audit('Media', $mediaId, 'Media Added', ['mediaId' => $mediaId, 'widgetId' => $this->widgetId]);
```
dasgarner, line 1274: This will get us a message whenever a media item is linked to a widget, which is a good start. I think the message could make that clearer - i.e. "Media added to Widget".

lib/Entity/Widget.php
```diff
@@ -1296,6 +1299,9 @@ private function unlinkMedia()
             $i++;
             $sql .= ',:mediaId' . $i;
             $params['mediaId' . $i] = $mediaId;
+
+            // audit the media being deleted
+            $this->getLog()->audit('Media', $mediaId, 'Media Removed', ['mediaId' => $mediaId, 'widgetId' => $this->widgetId]);
```
dasgarner, line 1304: This makes sense - although same as above regarding the message.

It would also be useful to log any mediaId's removed during Widget delete on the existing audit message for widgets that have been deleted.

dasgarner (changes requested):

dasgarner (approved):

ifarzana (approved):
Approving to build container

ifarzana:
TEST: login.cy.js END

PR #2736, opened 2024-09-12
Locale: fix error with translations generation script.

dasgarner:
relates to xibosignage/xibo#3500

### DIFF
--- a/bin/locale.php
+++ b/bin/locale.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2023 Xibo Signage Ltd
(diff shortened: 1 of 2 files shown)


ifarzana (approved):

dasgarner:
@ifarzana with your approval I do not think there is anything specific to test here in a PR container. The process for generating those files is only run during release, so any further issues should be spotted then. If you agree I will merge.

ifarzana (approved):
Please merge this PR.

ifarzana:
TEST: login.cy.js END

PR #2739, opened 2024-09-17
Campaign: Layout assignment filtering improvements

nadzpogi:
relates to xibosignage/xibo#3482


PeterMis (approved):
Looks good to me.

I was slightly concerned about possible duplicate ids ( there is folderId for the campaign too), but I think we handle that on submit to avoid issues, so should be all good.

ifarzana:
TEST: /Campaign END

PR #2740, opened 2024-09-18
Schedule Criteria: weather "clear sky" condition does not match

nadzpogi:
relates to xibosignage/xibo#3487

- Modified metric IDs to better align with JSON responses from the OWM API.
- Added a prefix to avoid conflicts with other potential weather schedule criteria.
- Removed metrics that are not present in OpenWeather API responses.


PeterMis (dismissed):

PeterMis (dismissed):

dasgarner (changes requested):
The OWM prefix doesn't make sense - criteria should be service agnostic so a prefix of `weather_` would make more sense.

dasgarner (approved):

dasgarner (approved):

ifarzana:
TEST: /Schedule END

PR #2748, opened 2024-09-25
Reports: Add Proof of Play MongoDB grouping

mgbaybay:
## Changes
FE:
- Fixed column duration column calculation

BE:
- [MongoDB] Added data calculation for accumulating the values according to grouping option

Relates to: https://github.com/xibosignage/xibo/issues/3356

### DIFF
--- a/lib/Report/ProofOfPlay.php
+++ b/lib/Report/ProofOfPlay.php
@@ -1255,20 +1255,20 @@ class ProofOfPlay implements ReportInterface
     /**
      * Group by display group in MongoDB
      * @param array $rows
-     * @param array $displayIds
+     * @param array $filteredDisplayGroupIds
      * @return array
      * @throws NotFoundException
      */
-    private function groupByDisplayGroupMongoDb(array $rows, array $displayIds = []) : array
+    private function groupByDisplayGroupMongoDb(array $rows, array $filteredDisplayGroupIds) : array
     {
         $data = [];
+        $displayInfoArr = $this->displayGroupFactory->query();
 
         // Get the display groups
         foreach ($rows as $row) {
-            foreach ($this->displayGroupFactory->query(null, ['displayId' => $row['displayId']]) as $dg) {
-
+            foreach ($displayInfoArr as $dg) {
                 // Do we have a filter?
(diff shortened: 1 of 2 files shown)


lib/Report/ProofOfPlay.php
```diff
@@ -1228,4 +1251,115 @@ private function groupByTagType(string $tagType) : string
...
+     * @return array
+     * @throws NotFoundException
+     */
+    private function groupByDisplayGroupMongoDb(array $rows, array $displayIds = []) : array
+    {
+        $data = [];
+
+        // Get the display groups
+        foreach ($rows as $row) {
+            foreach ($this->displayGroupFactory->query(null, ['displayId' => $row['displayId']]) as $dg) {
```
dasgarner, line 1268: Should we consider caching the display somewhere so that we don't hit the database for each stat record?

lib/Report/ProofOfPlay.php
```diff
@@ -1228,4 +1251,115 @@ private function groupByTagType(string $tagType) : string
...
+
+                // Do we have a filter?
+                if (!$displayIds || in_array($dg->displayGroupId, $displayIds)) {
```
dasgarner, line 1271: We're checking here whether the displayGroupId is in an array of displayIds (for filtering purposes), but will that ever be true?

lib/Report/ProofOfPlay.php
```diff
@@ -1228,4 +1251,115 @@ private function groupByTagType(string $tagType) : string
...
+    {
+        $tags = [];
+        $filter = match ($tagsType) {
+            'media' => 'Media',
+            'layout' => 'Layout',
+            'dg' => 'Display',
+        };
+
+        // What type of tags are looking for?
+        foreach ($this->tagFactory->query() as $tag) {
```
dasgarner, line 1353: This is always looking at all possible tags, which feels like it could be optimised, but I am not sure why we look for them all?

dasgarner (changes requested):

dasgarner (approved):

ifarzana:
TEST: /Reporting END

PR #2762, opened 2024-10-10
Elements need to obey sharing permissions

maurofmferrao:
relates to xibosignage/xibo#3516

 - Prevent elements and groups to be deleted if parent widget isn't editable
 - Failed to create widget when adding widget due to permissions lock adding fix
 - Hide element group edit button if it's not editable


dasgarner (approved):

ifarzana:
TEST: /Layout END

ifarzana:

PR #2764, opened 2024-10-11
Layout/Playlist Editor: Add dismiss-able notice for minimum resolution

maurofmferrao:
relates to xibosignage/xibo#3512

### DIFF
--- a/ui/src/editor-core/common.js
+++ b/ui/src/editor-core/common.js
@@ -227,18 +227,24 @@ module.exports = {
   /**
    * Handle minimum dimensions for the editor
    * @param {object} editor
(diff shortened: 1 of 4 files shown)


dasgarner (approved):

maurofmferrao:
@dasgarner , the zoom level was being used for the calculation of the minimum screen size, which isn't really compatible with the message, and also in a lower window size, decreasing the zoom level will make it usable. So can you please review the newest commit? 🙇🏻

dasgarner (approved):

gracec20:
TEST: /Layout END

gracec20:

PR #2767, opened 2024-10-14
XMDS: spread collection interval with a small offset

dasgarner:
This PR applies a small offset (+/- 10 seconds) to the collection interval so that displays which are started at the same time, i.e. at the start of day, or rebooted via command) do not keep connecting concurrently. Instead they spread out after the first connection.

 fixes to xibosignage/xibo#3530


nadzpogi (approved):

ifarzana:
TEST: login.cy.js END

PR #2792, opened 2024-11-05
Calendar widget issues with triggers and number of items

dasgarner:
This PR addresses two issues:
 - https://github.com/xibosignage/xibo/issues/3509
 - https://github.com/xibosignage/xibo/issues/3515

I don't think it was strictly necessary to remove the `typeof` check,  but the new version does make it easier to understand what is happening, so I kept it in. Triggering using xiboIC with a target of 0 disables the source checking for the trigger.

onDataLoad handles numItems and that has been removed from the one static template that referenced it. The provider always returns all items, so we handle it client side.

### DIFF
--- a/modules/calendar.xml
+++ b/modules/calendar.xml
@@ -185,6 +185,10 @@
 // meta: Metadata
 // properties: The properties for the widget
 
+// Track whether there is an event currently ongoing
+window.ongoingEvent = false;
+
+// Process events
 var now = moment();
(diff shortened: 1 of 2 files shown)


maurofmferrao (dismissed):
Looks good to me!

ifarzana:
TEST: /Layout/Editor/layout_editor_clock.cy.js END

PR #2793, opened 2024-11-05
Editor: Clear layout button

maurofmferrao:
relates to xibosignageltd/xibo-private#866


views/layout-form-clear.twig
```diff
@@ -0,0 +1,27 @@
...
+{% block formButtons %}
+    {% trans "No" %}, XiboDialogClose()
+    {% trans "Yes" %}, $("#layoutClearForm").submit()
+{% endblock %}
+
+{% block formHtml %}
+    <div class="row">
+        <div class="col-md-12">
+            <form id="layoutClearForm" class="XiboForm form-horizontal" method="clear" action="{{ url_for("layout.clear", {id: layout.layoutId}) }}">
+                {% set message %}{% trans "Are you sure you want to clear this item?" %}{% endset %}
```
dasgarner, line 19: I think we might want to specifically say Layout and maybe call out that it will be the Draft that is cleared?

dasgarner (approved):

PR #2799, opened 2024-11-08
Elements: Fix cast to int for number properties

maurofmferrao:
relates to xibosignage/xibo#3515


dasgarner (commented):
Does this work if the property is empty as well as if it is 0?

maurofmferrao:
> Does this work if the property is empty as well as if it is 0?

It was working for empty/undefined because it wasn't entering the if statement, but for 0 ( not `"0"` ) or `false` it would also skip the first condition, so I've updated the code to check for undefined only.

ifarzana:
TEST: login.cy.js END

PR #2821, opened 2024-11-29
User: force password change shouldn't be accessible unless set.

dasgarner:
fixes xibosignage/xibo#3555


PeterMis (approved):

ifarzana:
TEST: login.cy.js END

PR #2823, opened 2024-12-02
User: force password change shouldn't be accessible unless set.

dasgarner:
fixes xibosignage/xibo#3555


dasgarner:
@ifarzana if you agree this could be closed as the Kopff PR has been tested: https://github.com/xibosignage/xibo-cms/pull/2821

ifarzana:
Agreed, Thanks!

PR #2825, opened 2024-12-04
Elements: Scope style rules to element template only

maurofmferrao:
relates to xibosignageltd/xibo-private#735


modules/src/xibo-player.js
```diff
@@ -704,6 +704,28 @@ XiboPlayer.prototype.init = function() {
   // Create global render array of functions
   window.renders = [];
 
+  // If we have scoped styles
+  // convert the CSS rules to use it
+  $('[data-style-scope]').each((_idx, styleEl) => {
```
dasgarner, line 709: Does this handle a widget cached before this change? I.e. where there is no type/datatype/templateId ?

modules/widget-html-render.twig
```diff
@@ -50,11 +50,17 @@
...
         </style>
         {% if style|length > 0 %}
-        <style type="text/css">
             {% for item in style %}
-            {{ item|raw }}
+                {% if item.content %}
+                    <style type="text/css" data-style-scope="{{item.type}}_{{item.dataType}}__{{item.templateId}}">
+                        {{ item.content|raw }}
+                    </style>
+                {% else %}
```
dasgarner, line 58: This should never be the case because the WidgetHtmlRenderer has been changed at the same time.

rubenberttpingol (approved):

dasgarner (approved):

ifarzana:
TEST: /Layout END

ifarzana:
TEST: /Layout/Editor/layout_editor_mastodon.cy.js END

PR #2834, opened 2024-12-19
Schedule Page: Inline filter is broken

maurofmferrao:
relates to xibosignageltd/xibo-private#883


dasgarner (approved):

ifarzana:
TEST: login.cy.js END

PR #2839, opened 2024-12-23
Web Socket XMR and PWA routes

dasgarner:
This PR adds routes for a PWA to consume resources/data via signed links, which are proxied out to XMDS Soap class. It also includes a player type for chromeOS and handling to extract uploaded `.chrome` files (which are renamed zip files). These are then served at `/pwa`.

This PR also expects the new XMR container and communicates with it via HTTP. It therefore removes the ZMQ dependency from the web container.

Finally `develop` has been merged in to fix merge conflicts related to generating the player bundle.


PeterMis (approved):

ifarzana:
TEST: login.cy.js END

PR #2840, opened 2024-12-23
Library: fix filtering issue with media used as background images on layouts

dasgarner:
fixes xibosignageltd/xibo-private#879

See issue for PR context


ifarzana:
TEST: login.cy,js END

ifarzana:
TEST: login.cy.js END

ifarzana:
TEST: login.cy.js END

PR #2844, opened 2025-01-02
PHP: update twig and other dependencies to latest allowable versions.

dasgarner:


PeterMis (approved):

dasgarner:
@ifarzana this PR updates some dependencies to newer versions and should close out some of the other dependabot PRs too. There are no specific changes to test here, so I think it would be better tested in a release test? if you agree then I will bypass the requirements and merge it. Thanks.

ifarzana:
Please go ahead and merge this, thanks.

PR #2853, opened 2025-01-14
XMR: private API should be configured each time we start the container using the ENV variables

dasgarner:
fixes xibosignageltd/xibo-private#901

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright (C) 2024 Xibo Signage Ltd
+# Copyright (C) 2025 Xibo Signage Ltd
 #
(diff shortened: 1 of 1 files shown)


dasgarner:
@qaxlyka I do not believe it will be possible to QA test this change - running any of the automated test suite would be sufficient to prove the container still runs.

PeterMis (approved):

qaxlyka:
Hi @dasgarner i dont know yet what im doing, i was just looking around to understand how it works. Might have done something I didn't mean during that time 😅

dasgarner:
TEST: login.cy.js END

dasgarner:
@qaxlyka no problem, I ran an automated test which proves the container still works after my changes. That's all we need for this one. Thanks!

PR #2856, opened 2025-01-16
Fixed Reporting: time connected summary report does not work with Today as the range

mgbaybay:
## Changes
- Fixed data calculation for today as date range option in specific reports listed below. Previous behavior uses current datetime as the $toDt when 'today' is selected, hence, will retain this behavior instead of calculating the whole day

Reports affected:
- API Requests
- Display Alerts
- Session History
- Time Disconnected

Relates to: https://github.com/xibosignageltd/xibo-private/issues/893


nadzpogi (approved):
Looks all good to me! 😀

maurofmferrao:
TEST: login.cy.js END

PR #2864, opened 2025-01-21
Increase character limit for Schedule Criteria's type and metric

nadzpogi:
relates to xibosignageltd/xibo-private#900


ifarzana:
TEST: login.cy.js END

PR #2873, opened 2025-01-23
Schedule Page: Fix weekdays trans

maurofmferrao:
relates to xibosignageltd/xibo-private#913


ifarzana:

ifarzana:
TEST: login.cy.js END

PR #2889, opened 2025-01-30
New onboarding page

chaolen:
Applied new design for onboarding page.


views/components.twig
```diff
@@ -0,0 +1,45 @@
...
+                    <a href="{{ otherLink }}" class="optional-link" target="_blank">Learn More</a>
+                {%else%}
+                    <a href="{{firstLink}}" type="button" class="btn-rounded btn-orange" style="font-size: 16px;">Get Started</a>
+                    <a href="{{secondLink}}" type="button" class="btn-rounded btn-outlined" style="font-size: 16px;" target="_blank">Learn More</a>
+                {% endif %}
+            </div>
+        </div>
+    </div>
+{% endmacro %}
+
```
maurofmferrao, line 30: We might need to remove the extra "new lines". Same for the CSS file.

web/theme/default/css/xibo.css
```diff
@@ -1613,4 +1613,266 @@ div.dataTables_wrapper div.dataTables_info {
...
+}
+
+.scroll-up {
+    position: fixed;
+    margin: 30px;
+    bottom: 0;
+    right: 0;
+}
+
+html {
```
maurofmferrao, line 1876: We need to be careful with the scope of the CSS, since it's going to be applied to the whole CMS.
For smooth scroll, I think we can remove this and check later if we can do it with JS. For a class like `.text-left`, we need to wrap with with the scope for the main container, like: `.welcome-page .text-left`

web/theme/default/css/xibo.css
```diff
@@ -1613,4 +1613,266 @@ div.dataTables_wrapper div.dataTables_info {
...
+    font-weight: 500;
+}
+
+.links-list{
+    display: flex;
+    column-gap: 24px;
+    margin-top: auto;
+}
+
+.optional-link{
```
maurofmferrao, line 1852: To make the CSS consistent, can we have a space between the rule and the curly bracket?

`.optional-link {` instead of `.optional-link{`

maurofmferrao (changes requested):
Just some changes for the scope of CSS and code styling

chaolen:
Hi, @maurofmferrao! I made some new changes for this PR.

maurofmferrao (dismissed):
Looks good to me! :)

maurofmferrao (approved):

ifarzana:
TEST: login.cy.js END

PR #2911, opened 2025-02-10
Template: exclude canva regions when no widgets are included in the template save

dasgarner:
fixes xibosignage/xibo#3574


PeterMis (approved):

ifarzana:
TEST: /Layout/Editor/layout_editor_clock.cy.js END

PR #2913, opened 2025-02-12
RSS Ticker with new field categories

ZoranIT:
I have added the categories field - straight forward - in the Article, article-element and RssParser, just to be able to get it and use it in a layout.
It was tested with the feeds we receive from Austrian news agency and another I found online.


ZoranIT:
Oh, sorry. This was my mistake.

dasgarner:
@maurofmferrao

maurofmferrao (dismissed):
Looks good to me!

ZoranIT:
Is there a way to merge these add on into one of the next versions? In my opinion it would be enough to just return the first category in the feed if it is easier.

PR #2914, opened 2025-02-13
Weather Element: The image does not display on the player

maurofmferrao:
relates to xibosignage/xibo#3524

 - Also includes a fix for onTemplateRender, to target the right container and pass the items as data


dasgarner (approved):
I presume passing {} for items is because they are global elements and don't have data?

PeterMis (approved):

ifarzana:
TEST: login.cy.js END

PR #2930, opened 2025-02-27
Library: replace image should handle image based elements and mediaSelectors

dasgarner:
In this PR we extend the element handling further to cover off properties of type mediaSelector. We also cover widgets which may not be the global element, as they can have image elements inside.
 fixes xibosignage/xibo#3614


maurofmferrao (dismissed):
Looks good to me!

ifarzana:
TEST: /Library END

ifarzana:

PR #2931, opened 2025-02-27
DataSet: widgets not showing 0 values due to empty check.

dasgarner:
This PR removes the empty check when parsing out dataset data. It is not clear why the empty check was introduced (in 2023) and there isn't any notes in the commit history or code. My best guess is that it was missing index handling and so I have replaced it with `??`.

 fixes xibosignage/xibo#3564

### DIFF
--- a/lib/Listener/DataSetDataProviderListener.php
+++ b/lib/Listener/DataSetDataProviderListener.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2023 Xibo Signage Ltd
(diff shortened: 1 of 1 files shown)


maurofmferrao (approved):

qaxlyka:
TEST: layout_editor_dataset.cy.js END

qaxlyka:
TEST: /Layout/Widget/layout_editor_dataset.cy.js END

ifarzana:

PR #2938, opened 2025-03-03
Menuboard | Renaming a Category does not update in CMS Layout Editor fix

mgbaybay:
## Changes
- Fixed Menuboard | Renaming a Category does not update in CMS Layout Editor

Relates to: https://github.com/xibosignage/xibo/issues/3596

### DIFF
--- a/lib/Controller/MenuBoardCategory.php
+++ b/lib/Controller/MenuBoardCategory.php
@@ -316,6 +316,7 @@ class MenuBoardCategory extends Base
 
         $menuBoardCategory = $this->menuBoardCategoryFactory->create($id, $name, $mediaId, $code, $description);
         $menuBoardCategory->save();
+        $menuBoard->save();
 
         // Return
         $this->getState()->hydrate([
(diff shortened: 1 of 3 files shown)


lib/Controller/MenuBoardCategory.php
```diff
@@ -316,6 +316,7 @@ public function add(Request $request, Response $response, $id): Response
 
         $menuBoardCategory = $this->menuBoardCategoryFactory->create($id, $name, $mediaId, $code, $description);
         $menuBoardCategory->save();
+        $menuBoard->save();
```
dasgarner, line 319: I think this is a good approach, but I wonder if we need to pass an options array to disable the audit message for this particular save?

mgbaybay: Thank you, Dan :) Requesting review on [this](https://github.com/xibosignage/xibo-cms/pull/2938/commits/4c5d26973201ce491a7e2bb05c9e01a87bd5f2ed), please. Thanks!

dasgarner (approved):

qaxlyka:
TEST: /Library/menuboards.cy.js END

PR #2942, opened 2025-03-06
Actions: Next/Previous Layout shows regions as target

maurofmferrao:
relates to xibosignage/xibo#3621


qaxlyka:
TEST: login.cy.js END

PR #2945, opened 2025-03-07
Optional Javascript input field: Add help text to not use array style syntax

nadzpogi:
relates to xibosignage/xibo#3597


qaxlyka:
TEST: login.cy.js END

PR #2950, opened 2025-03-10
Fixed Advanced > Logs: Date filter is not working

mgbaybay:
## Changes
Contains improvements on the current Advanced Filter for Logs: 
- Updated label from `fromDate` to `Date`
- Used `dateTime` for filter instead of `date` only
- Provided help text for user experience 

*Previously, the backend uses `00:00:00` as the default time, thus, the only update needed is for the frontend to use datetime and include the time when sending data to the backend.

Relates to: https://github.com/xibosignage/xibo/issues/3573


maurofmferrao (approved):
Looks good to me!

PR #2960, opened 2025-03-17
Player Version: fix uploading ChromeOS packages in WL or sub-folder

dasgarner:
When unpacking a ChromeOS package, fully replace the manifest. This requires detecting the base url, so that if we're running in a sub-folder we get the right values.

 fixes xibosignage/xibo#3632


ifarzana:
TEST: /Display END

PR #2971, opened 2025-04-08
Settings > Displays: Enable Stats Collection (set to YES) is not inherited for new layouts

nadzpogi:
relates to xibosignage/xibo#3649


mgbaybay (approved):

qaxlyka:
TEST: /Layout, /Reporting END

PR #2972, opened 2025-04-08
Fixed Media Filters: Clear filters in media grid should set Retired to No

mgbaybay:
## Changes
- Resetting the datatables defaults the selection to the first option, hence changing the option index would solve the issue

Relates to: https://github.com/xibosignageltd/xibo-private/issues/954


qaxlyka:
TEST: /Library END

PR #2974, opened 2025-04-09
Tag: Editing tag name does not reflect in criteria tag in display group

nadzpogi:
relates to xibosignage/xibo#3631


PeterMis (commented):
I think that works well in most cases.

Should we cover a situation where we have tag with value entered in the DisplayGroup dynamic criteria? ie tag|value e.g. myTag|myValue ?

As it stands I don't think it would be handled in the onTagEdit 🤔 ( it would not match  the old tag )

nadzpogi:
Oh right, I missed that one, thanks!

PeterMis (approved):

ifarzana:
TEST: /Display END

PR #2975, opened 2025-04-10
DataSet: remote data sets have a URI box which has helptext for URL - should they be the same?

nadzpogi:
relates to xibosignage/xibo#3650


mgbaybay (approved):

Kent0425:
TEST: datasets.cy.js END

ifarzana:

PR #2981, opened 2025-04-23
Template: Zones not saving when moved

maurofmferrao:
relates to xibosignage/xibo#3656


ifarzana:

PR #2992, opened 2025-05-02
XTR: fix the audit log archive task so that it does not do an unbounded delete

dasgarner:
Fixes xibosignage/xibo#3672


mgbaybay (approved):

ifarzana:
TEST: /Administration END

PR #2994, opened 2025-05-02
Notifications: present notification settings in a clearer way on user and user group. Fix the issue sending email only to the users group settings

dasgarner:
Fixes xibosignage/xibo#3661


mgbaybay (approved):

ifarzana:
TEST: /Display, /Administration END

PR #3014, opened 2025-06-04
Do not display "Private Folder"

MrStefanW:
In the layout editor, for each folder users don't have access to "Private Folder" will be displayed, making accessible folders harder to identify.


MrStefanW:
Opened this as a feature request in the Xibo Community:
https://community.xibo.org.uk/t/do-not-display-private-folder/35498

MrStefanW:
This is no longer needed since version 4.5.0

PR #3018, opened 2025-06-10
XTR: fix install type

dasgarner:
Anon usage stats are being run via XTR which did not have the INSTALL_TYPE environment variable set. During initial testing this came from manually running the CI/Dev container via the command line, which did have this set.
xibosignageltd/xibo-private#961


dasgarner:
@ifarzana if this passes code review and an automated test, I would be inclined to merge it as we can only test it once deployed in a real container. Agreed?

mgbaybay (approved):

ifarzana:
TEST: login.cy.js END

PR #3027, opened 2025-06-18
Dataset Add Row: Show error message when no Value columns exist

maurofmferrao:
relates to xibosignage/xibo#3693


dasgarner (approved):

lykqa:
TEST: Library/datasets.cy.js END

PR #3028, opened 2025-06-20
RSS Ticker: Add new categories field

maurofmferrao:
Relates to https://github.com/xibosignage/xibo-cms/pull/2913


lib/Widget/RssProvider.php
```diff
@@ -160,6 +160,7 @@ public function fetchData(DataProviderInterface $dataProvider): WidgetProviderIn
                 $article->link = $item->getUrl();
                 $article->date = Carbon::instance($item->getDate());
                 $article->publishedDate = Carbon::instance($item->getPublishedDate());
+                $article->categories = $item->getCategories();
```
dasgarner, line 163: What gets returned here? Is it a comma separated list of categories? It is an array?

I am not clear how this gets mapped 1 to 1 with an element?

maurofmferrao: Yeah, it's a comma separated list. Not sure what we wanted to implement for the categories element, but for now it just outputs all the categories with that string. Do you think we should save them separately as an array and use the elements with an index?

That might become confusing since we already have index/slot for each one of the Ticker records... 🤔

dasgarner: The question is whether we think a comma separated string is useful?

ZoranIT: We use the first item of this string in a module template. In the "onTemplateRender" tab we replace the category value:
image
When I was searching for different feeds that use this category I found some that had multiple categories within a feed. That's why I implemented it this way.
All other ideas are also okay, but the first value is necessary in my opinion.

PR #3034, opened 2025-07-02
Fixes for API scope checking and the default scope

dasgarner:
This PR updates the way we check route access against the scopes granted to the access token.

It also grants the configured scopes to applications which do not request scopes (for backwards compatibility). It also ensures that the right scopes are granted based on the request.

### DIFF
--- a/db/migrations/20250702145100_fix_oauth_routes_table_migration.php
+++ b/db/migrations/20250702145100_fix_oauth_routes_table_migration.php
@@ -1,0 +1,64 @@
+<?php
+/*
+ * Copyright (C) 2025 Xibo Signage Ltd
+ *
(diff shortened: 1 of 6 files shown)


dasgarner:
@nadzpogi please can you pull this locally and try it with Canva ensuring that only the scopes needed are granted in the Application record? Thanks!

nadzpogi (approved):
Tested on my local and It worked as expected!

dasgarner:
TEST: login.cy.js END

dasgarner:
@ifarzana this is an API change and both Nadz and I have confirmed it works. Do you want this to pass through QA, or are you happy for it to be merged as-is? Thanks.

ifarzana:
Please go ahead and merge. Thanks.

PR #3035, opened 2025-07-02
CMS Alias: fix an issue with file downloads. Clarified some things.

### LINKED ISSUE #3655: Missing CMS Alias: File downloads via XMDS don't work when CMS runs in a subdirectory
After upgrading from version 3.2.0 to 4.2.1, Players are unable to download files due to incorrect download links when a CMS alias is used.

Details:
In config.env, when we specify: 
`CMS_ALIAS=/ContentManager `
This worked correctly in version 3.2.0.

After upgrading to 4.2.1, required file download links point to:

`https://example.com/xmds.php `

Expected: The correct URL should include /ContentManager/, e.g., 
`https://example.com/ContentManager/xmds.php `

Actual: Missing /ContentManager/, causing 404 errors as requests never reach CMS.

dasgarner:
This PR updates the way we sign links so that it correctly calculates the base URL including any sub-folders. It also attempts to clarify how methods should be used.

fixes xibosignage/xibo#3655

I believe the fixes are good, but to test them properly I need a container built so that I can run a CMS with alias. I have tested without CMS alias.

### DIFF
--- a/lib/Controller/Base.php
+++ b/lib/Controller/Base.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2023 Xibo Signage Ltd
+ * Copyright (C) 2025 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - https://xibosignage.com
  *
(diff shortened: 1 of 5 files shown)


mgbaybay (approved):

ifarzana:
TEST: /Schedule END

PR #3036, opened 2025-07-03
DataSet: prevent columns created with reserved words.

### LINKED ISSUE #3692: Datasets: Column names using disallowed or reserved keywords are currently accepted during dataset creation, but result in errors when used within Layouts
The following names are disallowed for use in dataset column naming due to naming convention restrictions, as defined here:
[Disallowed Names](https://github.com/xibosignage/xibo-cms/blob/4.2.3/lib/Widget/Definition/Sql.php#L30-L45)
Examples: From, Drop

Steps to Reproduce:

1. Create a new dataset.
2. Add the following columns, setting the Data Type to Date:
* From – set to today's date at midnight (e.g., 11-06-2025 00:00)
* Drop – set to today's date at 23:59 (e.g., 11-06-2025 23:59)
3. Also, add a column for an image.
4. Create a new layout and use the dataset to display the image.
5. In Advanced Filter, apply the following condition:
`DATE_ADD(NOW(), INTERVAL 1 HOUR) > From AND DATE_ADD(NOW(), INTERVAL 1 HOUR) < Drop`

Observed Behavior:
> The image does not display.

Expected Behavior:
> The image should be visible within the specified time range.

Root Cause:
Using reserved keywords (e.g., From, Drop) as column names interferes with filtering logic and causes unexpected behavior. To confirm, try using alternative column names that are not in the disallowed list — the filter will then work as intended.

dasgarner:
This PR prevents columns being added or edited with any of the reserved words as the heading.
fixes xibosignage/xibo#3692

### DIFF
--- a/lib/Entity/DataSetColumn.php
+++ b/lib/Entity/DataSetColumn.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2024 Xibo Signage Ltd
+ * Copyright (C) 2025 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - https://xibosignage.com
  *
@@ -22,6 +22,7 @@
 
 namespace Xibo\Entity;
 
+use Illuminate\Support\Str;
 use Respect\Validation\Validator as v;
 use Xibo\Factory\DataSetColumnFactory;
 use Xibo\Factory\DataSetColumnTypeFactory;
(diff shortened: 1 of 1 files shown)


maurofmferrao (approved):
Looks good to me!

lykqa:
TEST: Library/datasets.cy.js END

PR #3037, opened 2025-07-05
XMDS: catch errors with sync group in register display

dasgarner:
In this PR we catch errors in Register where we try to get the Sync Group configuration. If there are errors they will be logged and the display will not be in a sync group until they are resolved.

This PR also updated the Display Edit form to make it clearer what those details are used for.


mgbaybay (approved):

Kent0425:
TEST: /Display END

Kent0425:
Tested! ✅

PR #3041, opened 2025-07-09
Update the help button UI

maurofmferrao:
relates to xibosignageltd/xibo-private#976


dasgarner (dismissed):
The only thing I wasn't sure about was the case change for `isAllowWhiteLabel`. If you're sure that is how it looks in the associated JSON file, then that's fine.

maurofmferrao:
> The only thing I wasn't sure about was the case change for `isAllowWhiteLabel`. If you're sure that is how it looks in the associated JSON file, then that's fine.

The issue with that variable is that in the `help-links.yaml` we use the Camel Case in "Label", and were using "label" in `HelpLink.php`. That means it wasn't working for links we wanted to hide from the white label before.

dasgarner (approved):

lykqa:
TEST: login.cy.js END

PR #3045, opened 2025-07-17
Library: allowMediaTypeChange parameter fix

### LINKED ISSUE #3699: API: library add does not allow media type changes when that option is provided
This used to work in v3 by allowing all file based media as valid extensions when the allowMediaTypeChange was set to 1. Currently in v4 it ignores this option when assessing valid extensions, and only allows extensions for the type of file being uploaded.

dasgarner:
fixes xibosignage/xibo#3699

### DIFF
--- a/lib/Factory/ModuleFactory.php
+++ b/lib/Factory/ModuleFactory.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2024 Xibo Signage Ltd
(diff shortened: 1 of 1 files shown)


mgbaybay (approved):

lykqa:
TEST: Library END

lykqa:
TEST: Library/menuboards.cy.js END

PR #3049, opened 2025-07-21
Ticker: Support special chars in title, author, and summary

maurofmferrao:
relates to xibosignage/xibo#581

### DIFF
--- a/lib/Widget/RssProvider.php
+++ b/lib/Widget/RssProvider.php
@@ -155,8 +155,8 @@ class RssProvider implements WidgetProviderInterface
             foreach ($feedItems as $item) {
                 /* @var Item $item */
                 $article = new Article();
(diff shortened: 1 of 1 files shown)


dasgarner:
Could you explain what we're trying to achieve here?

Lets say someone uses a feed already which has a title of `<h1>Title</h1>`, will that have been displayed previously, and will that change as part of this fix?

maurofmferrao:
The idea was to handle HTML special characters like `&amp;`  so they render properly as `&`. But I didn't try a HTML tag like that, it's basically removing that `<h1>Title</h1>` from the title. 

Do you have a better way of doing that without removing HTML tags?

maurofmferrao:
The fix for this issue could introduce potentially unsafe behaviour. We're closing this PR.

PR #3059, opened 2025-08-06
Implement simplified Schedule form

maurofmferrao:
relates to xibosignageltd/xibo-private#1017


PeterMis (dismissed):
Looks good to me I think.

We probably want a migration for the 3 new settings with default values and expected userSee/userChange parameters, other than that it makes sense to me 👍

PeterMis (dismissed):

PeterMis (dismissed):

lib/Controller/DisplayGroup.php
```diff
@@ -399,6 +399,24 @@ public function grid(Request $request, Response $response)
...
 
+            // Schedule
+            if ($this->getUser()->featureEnabled('schedule.add')
+                && ($this->getUser()->checkEditable($group)
+                    || $this->getConfig()->getSetting('SCHEDULE_WITH_VIEW_PERMISSION') == 1)
+            ) {
+                $group->buttons[] = ['divider' => true];
+
+                $group->buttons[] = array(
+                    'id' => 'display_button_schedule',
```
PeterMis: It does not really matter, but should that be `displaygroup_button_schedule` ?
just to keep the same naming convention

PeterMis (commented):
Looks good, just minor comments below

lib/Controller/Settings.php
```diff
@@ -643,6 +643,21 @@ public function update(Request $request, Response $response)
             $this->getConfig()->changeSetting('WHITELIST_LOAD_BALANCERS', $sanitizedParams->getString('WHITELIST_LOAD_BALANCERS'));
         }
 
+        if ($this->getConfig()->isSettingEditable('SCHEDULE_WITH_GEO')) {
```
PeterMis: Do we still need those?
since we moved them to user features, we can probably remove this from settings?

PeterMis (dismissed):

PeterMis (dismissed):

dasgarner (approved):
Approved, but we need to confirm which scheduled features should be off by default

ifarzana:
TEST: /Schedule END

ifarzana:
TEST: login.cy.js END

PR #3063, opened 2025-08-07
Replace/Expire Media: Unable to delete expired media after replacement

nadzpogi:
relates to xibosignage/xibo#3654


PeterMis (dismissed):
I think that looks correct

In the issue I've seen a mention of : 
> we should make it clear that setting an expiry date would expire and delete all revisions of the media (current and prior)

Is that something that we've considered?

PeterMis (approved):
I think that looks good.
Interesting catch in that LEFT OUTER JOIN 🤔

Kent0425:
TEST: /Library/media.cy.js END

PR #3064, opened 2025-08-09
Images: fix content-type on thumbnails (library), previews (library,layout editor) and screenshots (display)

### LINKED ISSUE #3712: Library: Thumbnails output with content type of text/html
This doesn't appear to be a problem rendering in the grid, but its best practice and isn't good for opening those links in a new tab.

dasgarner:
This PR ensures that the content type is set correctly for images output via the CMS (thumbnails and the like). The key here is setting the content type, but also ensuring we do not `echo` out directly.

@nadzpogi you should be able to pull this locally to test with the Canva thing you are working on.

fixes xibosignage/xibo#3712

### DIFF
--- a/lib/Controller/Display.php
+++ b/lib/Controller/Display.php
@@ -2222,7 +2222,7 @@ class Display extends Base
         $file = 'screenshots/' . $id . '_screenshot.jpg';
 
         // File upload directory.. get this from the settings object
-        $library = $this->getConfig()->getSetting("LIBRARY_LOCATION");
+        $library = $this->getConfig()->getSetting('LIBRARY_LOCATION');
         $fileName = $library . $file;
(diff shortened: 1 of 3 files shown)


nadzpogi (approved):

Kent0425:
TEST: login.cy.js END

Kent0425:
Looks like we have PHP issues on our CI/CD pipeline @dasgarner

dasgarner:
@Kent0425 what do you mean, sorry?

ifarzana:
You can safely merge this, @dasgarner . This PR has no issues with running the tests. 

We’ve encountered a problem where PHPUnit is returning an error for another PR, which is causing the test suite to fail.

PR #3068, opened 2025-08-18
Background images aren't checked for size rules - revisited

nadzpogi:
relates to xibosignage/xibo#3702

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -1846,12 +1846,12 @@ class Layout implements \JsonSerializable
                 $widgetValidator->validate($module, $widget, 'status');
             }
 
-            // Is this module file based? If so, check its released status
-            // Only proceed if the widget has at least one primary media file attached
-            if ($widget->getPrimaryMediaId() != 0) {
-                // Get all primary media IDs for this widget (audio IDs are excluded)
-                $mediaIds = $widget->getPrimaryMedia();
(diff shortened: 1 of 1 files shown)


PeterMis (approved):
Looks good

Kent0425:
TEST: /Library/media.cy.js, /Layout/Editor/layout_editor_empty.cy.js END

Kent0425:
TEST: /Library/media.cy.js

Kent0425:
TEST: login.cy.js END

ifarzana:
TEST: /Library END

ifarzana:
Let us run the XMDS tests locally to see if it shows any error.

PeterMis (changes requested):
Sorry Nadz I had another look at the changes due to the failing xmds test.

I don't think we can just have the `$widget->getPrimaryMedia() != 0` without either as it was regionSpecific check or try/catch on it or some other way.

As it stands this will return 404 `No file to return` on anything that does not have primary media associated with it like for example datasets without images - I think that's why the xmds test fails as it is trying to get data from dataset widget, but that widget and layout are not quite correct after seed database task runs with your changes.

I think the way you have it now, if any widget does not have primary mediaId on it it will come as invalid basically - therefore the layout will also be invalid.

PeterMis (approved):

ifarzana:
TEST: /Library END

PR #3076, opened 2025-08-20
Layout: duration is incorrect in the XLF

mgbaybay:
## Changes

- Fixed region duration calculation to consider the module default length when no duration is set

Relates to https://github.com/xibosignage/xibo/issues/3697

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -1457,15 +1457,10 @@ class Layout implements \JsonSerializable
                         'tempCyclePlaybackAverageDuration',
                         0
                     );
-
-                    // If this widget does not have a set duration, then we use the module default duration
                     if ($tempCyclePlaybackAverageDuration) {
                         $region->duration = $region->duration + $tempCyclePlaybackAverageDuration;
                     } else {
-                        $region->duration = $region->duration + (($widget->setDuration == 1)
-                            ? $widget->calculatedDuration
-                            : $module->defaultDuration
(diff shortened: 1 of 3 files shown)


dasgarner:
I can see why you made this change, however I am not sure if it is right. For example if the widget was a video, use duration would be set to 0, but we'd still want to take the calculated duration (which would be the video actual duration) rather than the module default.

I think the root cause of the issue must be something to do with the `$widget->calculatedDuration` not being set correctly for a widget in the layout that has the issue.

dasgarner (changes requested):
I don't think this is the root cause.

dasgarner (changes requested):
I don't think that can be right either I'm afraid to say. If you have more than one widget in a region, you wouldn't want to set them all to be the region duration.

I think we need to work on this together tomorrow to get to the bottom of it.

dasgarner (commented):
I think the API change is good. 

I'd like @maurofmferrao to look over the viewer change please.

maurofmferrao (approved):
Looks good to me!

lykqa:
TEST: Layout/Editor/layout_editor_options.cy.js END

mgbaybay:
@dasgarner requesting your review to proceed with merging, thank you!

dasgarner (approved):

Kent0425:
TEST: Layout/Editor/layout_editor_options.cy.js END

PR #3077, opened 2025-08-20
Fixed Canvas widget: Layout duration defaults to 1s for global elements

mgbaybay:
## Changes

- Set a default duration of 10s for canvas-only layout with global elements

Relates to https://github.com/xibosignage/xibo/issues/3648

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -1315,6 +1315,9 @@ class Layout implements \JsonSerializable
         // merge regions and drawers into one array and go through it.
         $allRegions = array_merge($this->regions, $this->drawers);
 
+        // Used to identify Layouts which only have Canvas with global elements
+        $isCanvasOnlyRegion = true;
+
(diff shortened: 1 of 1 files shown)


lib/Entity/Layout.php
```diff
@@ -1449,6 +1457,16 @@ public function toXlf()
                     $widgetDuration = Widget::$widgetMinDuration;
                 }
 
+                // Layouts which only have Canvas with global elements
+                if ($region->type == 'canvas'
+                    && $regionLoop == 0
```
dasgarner: This condition may not need to be here?

dasgarner (changes requested):

dasgarner (approved):

lykqa:
TEST: Layout/Editor/layout_editor_clock.cy.js END

lykqa:
TEST: Layout/Widget/layout_editor_clock.cy.js END

PR #3079, opened 2025-08-21
CMS Alias: modify fix for links in HTML requested via a PWA

### LINKED ISSUE #3655: Missing CMS Alias: File downloads via XMDS don't work when CMS runs in a subdirectory

dasgarner:
fixes xibosignage/xibo#3655

### DIFF
--- a/lib/Helper/LinkSigner.php
+++ b/lib/Helper/LinkSigner.php
@@ -49,8 +49,17 @@ class LinkSigner
         $itemId,
         string $storedAs,
         string $fileType = null,
+        bool $isRequestFromPwa = false,
     ): string {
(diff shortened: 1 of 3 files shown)


maurofmferrao (approved):
Looks good to me!

PR #3100, opened 2025-09-09
Cypress Test: Templates flow

lykqa:
https://github.com/xibosignageltd/qa/issues/577

### DIFF
--- a/cypress/e2e/Templates/templates.cy.js
+++ b/cypress/e2e/Templates/templates.cy.js
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2025 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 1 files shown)


cypress/e2e/Templates/templates.cy.js
```diff
@@ -0,0 +1,101 @@
...
+    cy.visit('/template/view');
+    cy.contains('Add Template').click();  
+    cy.get('#dialog_btn_2').should('be.visible').click();
+    cy.contains('Layout Name must be between 1 and 100 characters');
+  });
+
+  it('Creates a new template and exits layout editor', function() {
+    cy.visit('/template/view');  
+    cy.contains('Add Template').click();  
+    cy.get('#name').type('Santee');
```
ifarzana: Using hard-coded template names ("Santee") can lead to conflicts if the template already exists. Consider generating a unique name per test run. Please have a look at /Library/datasets.cy.js and use  cy.get('#name').type('Cypress Test Template ' + testRun + '_1');

cypress/e2e/Templates/templates.cy.js
```diff
@@ -0,0 +1,101 @@
...
+    cy.get('#dialog_btn_2').should('be.visible').click();
+    cy.get('#layout-editor').should('be.visible');
+  });
+
+  it('Opens and edits an existing template', function() {
+    cy.intercept('GET', '**/template*').as('templatesList');
+    cy.visit('/template/view');
+    cy.wait('@templatesList').its('response.statusCode').should('eq', 200);
+
+    cy.contains('Santee', { timeout: 10000 })
```
ifarzana: Timeouts like { timeout: 10000 } are OK, but consider asserting visibility. See
https://github.com/xibosignage/xibo-cms/blob/develop/cypress/e2e/Library/datasets.cy.js#L85

ifarzana (changes requested):
Overall looks good! I only have a few change requests.

cypress/e2e/Templates/templates.cy.js
lykqa: That's a good input. I'll have it re-written. Thanks Israt!

ifarzana (approved):

lykqa:
TEST: Templates/templates.cy.js END

PR #3106, opened 2025-09-16
Android: WebSocket XMR support from v4 R408

dasgarner:
As we update our player apps to support web sockets, we will also need to instruct the CMS to send those messages to XMR unencrypted and marked as "web socket". We can do this by version code and type. ChromeOS was already handled.


mgbaybay (dismissed):

dasgarner:
@ifarzana this PR for 4.4 tells Android R408 to use web socket XMR instead of ZMQ. I don't think there is anything to test here? If you agree I will merge it.

ifarzana:
please go ahead with merge.

PR #3115, opened 2025-09-25
Schedule: Error found when deleting a priority

maurofmferrao:
relates to xibosignage/xibo#3733

### DIFF
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -1727,8 +1727,8 @@ class Schedule extends Base
         $schedule->campaignId = $this->isFullScreenSchedule($schedule->eventTypeId)
             ? $sanitizedParams->getInt('fullScreenCampaignId')
             : $sanitizedParams->getInt('campaignId');
-        $schedule->displayOrder = $sanitizedParams->getInt('displayOrder', ['default' => $schedule->displayOrder]);
-        $schedule->isPriority = $sanitizedParams->getInt('isPriority', ['default' => $schedule->isPriority]);
+        $schedule->displayOrder = $sanitizedParams->getInt('displayOrder', ['default' => 0]);
+        $schedule->isPriority = $sanitizedParams->getInt('isPriority', ['default' => 0]);
         $schedule->dayPartId = $sanitizedParams->getInt('dayPartId', ['default' => $schedule->dayPartId]);
         $schedule->syncTimezone = $sanitizedParams->getCheckbox('syncTimezone');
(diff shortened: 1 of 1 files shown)


dasgarner (commented):
I think this is breaking change for the API which previously would preserve the existing value if the parameter has not been provided for update, now it would change it. We may need to distinguish between the parameter being empty (as provided by the form) vs the parameter being missing (as can be provided by the API)

maurofmferrao:
> I think this is breaking change for the API which previously would preserve the existing value if the parameter has not been provided for update, now it would change it. We may need to distinguish between the parameter being empty (as provided by the form) vs the parameter being missing (as can be provided by the API)

I've made some changes to the code, if param is provided from the form, but empty, set to default (0). If not provided from the form, keep previous existing value if exists, or set to 0. Does that make sense @dasgarner ?

lib/Controller/Schedule.php
```diff
@@ -1727,8 +1727,23 @@ public function edit(Request $request, Response $response, $id)
...
             ? $sanitizedParams->getInt('fullScreenCampaignId')
             : $sanitizedParams->getInt('campaignId');
-        $schedule->displayOrder = $sanitizedParams->getInt('displayOrder', ['default' => 0]);
-        $schedule->isPriority = $sanitizedParams->getInt('isPriority', ['default' => 0]);
+
+        // displayOrder and isPriority: if present but empty (""): set to 0
+        // if missing from form: keep existing value (fallback to 0 if unset)
+        $params = $request->getParams();
+
+        $schedule->displayOrder = array_key_exists('displayOrder', $params)
```
dasgarner: Did you try `$sanitizedParams->hasParam()` ?

maurofmferrao: > Did you try `$sanitizedParams->hasParam()` ?

Didn't know that method existed, that simplifies things a lot, I've made a new commit, thank you!

dasgarner (approved):

Kent0425:

Kent0425:
✅Tested!

lykqa:
I see that this is already merged to the develop since yesterday. Is it possible that im still encountering the issue right now in https://develop.xibo.co.uk/ after the merge?

maurofmferrao:
> I see that this is already merged to the develop since yesterday. Is it possible that im still encountering the issue right now in https://develop.xibo.co.uk/ after the merge?

Issues for `release43` are not being built into the `develop` CMS, for that we're using `cobham`.

PR #3119, opened 2025-09-30
XMDS: cache bandwidth usage (performance)

dasgarner:
fixes xibosignageltd/xibo-private#1040

This PR introduces a cache to the Bandwidth factory so that we only look up the current bandwidth usage from the database every 10 minutes instead of with each request.


mgbaybay (approved):

ifarzana:
TEST: login.cy.js END

PR #3120, opened 2025-10-01
Fixed Layout duration not updating correctly with Global Element/Image

mgbaybay:
## Changes

- Removed previous FE changes on widget duration calculation (the calculation should be from bottom to top)
- Added migration file to set the minimum canvas widget duration to 10s instead of forcing the BE to implement this
- Added a validation when changing the widget duration. The user-provided widget duration should only be used when use duration option is checked

Relates to https://github.com/xibosignage/xibo/issues/3725


maurofmferrao (approved):

lykqa:
TEST: Layout/Editor END

lykqa:
TEST: Layout/Editor/layout_editor_populated.cy.js END

PR #3123, opened 2025-10-01
XMDS: schedule criteria nodes not added to repeating events

dasgarner:
We need unique XML nodes created for each event, we can't reuse them like we were attempting to do
fixes xibosignage/xibo#3713


mgbaybay (approved):

ifarzana:
TEST: /Schedule END

PR #3132, opened 2025-10-07
Fixed layout duration not updating with video and global elements

mgbaybay:
## Changes

- Fixes incorrect duration when a video is added to a layout. The logic now ensures that before setting a minimum region duration for layouts with 'lead' regions (regions with defined durations), we only apply this if the widget’s default duration is not the longest in the layout.

Relates to https://github.com/xibosignage/xibo/issues/3725

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -1413,6 +1413,8 @@ class Layout implements \JsonSerializable
             }
 
             // Work out if we have any "lead regions", those are Widgets with a duration
+            $maxWidgetDurationInLayout = 1;
+
             foreach ($widgets as $widget) {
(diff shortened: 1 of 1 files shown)


dasgarner (commented):
I'd like to understand this fix a little more before it gets merged please! I'm fairly sure the XLF was correct in the report that was raised.

dasgarner (approved):
I think it works!

ifarzana:
TEST: /Layout END

ifarzana:
TEST: /Layout/Widget/layout_editor_rss_ticker.cy.js END

lykqa:
TEST: /Layout/Widget/layout_editor_rss_ticker.cy.js END

PR #3133, opened 2025-10-07
Image / Video / Playlist Scheduling Improvements

mgbaybay:
## Changes

Added:
- `tidyUnusedFullScreenLayout` - deletes unused fullscreen layouts if no associated event is found.
- `getLayoutResolutionId` - returns the closest matching resolution ID of the layout, used when editing a schedule

Modified:
- `getLinkedFullScreenLayout` - added the `properties` params to get the exact layout match to avoid creating duplicate layouts with the same properties

Relates to https://github.com/xibosignageltd/xibo-private/issues/1039


maurofmferrao (dismissed):

lib/Factory/LayoutFactory.php
```diff
@@ -3128,5 +3163,202 @@ private function prepareWidgetAndGetModule(Widget $widget): Module
...
+     * @params $id
+     * @params $resolutionId
+     * @params $backgroundColor
+     * @params $duration
+     * @params $type
+     * @throws InvalidArgumentException
+     * @throws NotFoundException
+     * @throws GeneralException
+     */
+    public function createFullScreenLayout($type, $id, $resolutionId, $backgroundColor, $duration): array
```
dasgarner: It isn't good practice for this to return an array - the expectation is that the factory would return a Layout object.

It seems like the only reason for this is to return a layout and a message, but the message is always the same anyway?

So if this is successful you can imply the message? If not successful raise an exception with the message?

lib/Controller/Schedule.php
```diff
@@ -2661,4 +2725,23 @@ private function isSyncEvent($eventTypeId): int
...
+     */
+    private function doFsLayoutCleanup($schedule, bool $isDelete = false): void
+    {
+        // Is this layout being used in another event?
+        $events = $this->scheduleFactory->getByCampaignId($schedule->campaignId);
+
+        // When deleting an event, we need to check if it's the only event using that layout
+        if (empty($events) || (count($events) === 1 && $isDelete)) {
+            $layout = $this->layoutFactory->getByCampaignId($schedule->campaignId);
+            $event = new LayoutDeleteEvent($layout[0]);
```
dasgarner: The name of this event isn't clear - if you were listening for the event you might assume it fires for all layout deletes, but it is only for deleting unused full screen layouts.

lib/Controller/Schedule.php
```diff
@@ -2661,4 +2725,23 @@ private function isSyncEvent($eventTypeId): int
...
         return ($eventTypeId === \Xibo\Entity\Schedule::$SYNC_EVENT) ? 1 : 0;
     }
+
+    /**
+     * Deletes unused full screen layouts
+     * @param $schedule
+     * @param bool $isDelete
+     * @throws NotFoundException
+     */
+    private function doFsLayoutCleanup($schedule, bool $isDelete = false): void
```
dasgarner: We are not charged by the letter 😄 it took me a while to work out what `fs` meant.

lib/Controller/Schedule.php
```diff
@@ -2057,6 +2105,17 @@ public function edit(Request $request, Response $response, $id)
             }
         }
 
+        // Old layout cleanup
```
dasgarner: Does it make sense to do this inside the transaction, or would it be better as a background task?

lib/Controller/Layout.php
```diff
@@ -3494,161 +3494,28 @@ public function createFullScreenLayout(Request $request, Response $response): Re
     {
         $params = $this->getSanitizer($request->getParams());
         $type = $params->getString('type');
```
dasgarner, line 3496: Should we make sure that the type provided is one that we expect?

dasgarner (changes requested):

lib/Controller/Layout.php
mgbaybay: Thank you, Dan! I've added data validation for the type [here](https://github.com/xibosignage/xibo-cms/pull/3133/commits/752362a5eaa25dbda522b6c9b26d00f429fd92b7#diff-b5cc32097da404b17a192565d806303fc55b0e43787db03d171dc77e1753c5a7R3506-R3510).

lib/Controller/Schedule.php
mgbaybay: Thanks, Dan! You're right, I've added it as a background task under [Regular Maintenance](https://github.com/xibosignage/xibo-cms/pull/3133/commits/752362a5eaa25dbda522b6c9b26d00f429fd92b7#diff-b5cc32097da404b17a192565d806303fc55b0e43787db03d171dc77e1753c5a7R3506-R3510).

lib/XTR/MaintenanceRegularTask.php
```diff
@@ -649,4 +652,34 @@ private function assertXmrKey(): void
...
+     * @throws NotFoundException
+     * @throws GeneralException
+     */
+    private function tidyUnusedFullScreenLayout(): void
+    {
+        $this->runMessage .= '## ' . __('Tidy Unused FullScreen Layout') . PHP_EOL;
+        Profiler::start('RegularMaintenance::tidyUnusedFullScreenLayout', $this->log);
+        $count = 0;
+
+        foreach ($this->layoutFactory->query(null, [
```
dasgarner, line 667: Won't this tidy them all - including in use ones?

lib/Factory/LayoutFactory.php
```diff
@@ -34,12 +34,14 @@
 use Xibo\Entity\User;
 use Xibo\Entity\Widget;
 use Xibo\Helper\DateFormatHelper;
+use Xibo\Helper\Environment;
```
dasgarner: This seems like an odd import?

lib/Controller/Layout.php
```diff
@@ -3494,161 +3494,33 @@ public function createFullScreenLayout(Request $request, Response $response): Re
...
-            } else if ($type === 'playlist') {
-                $resolutionId = $this->resolutionFactory->getClosestMatchingResolution(
-                    1920,
-                    1080
-                )->resolutionId;
-            }
-        }
-
-        $layout = $this->layoutFactory->createFromResolution(
+        $layoutData = $this->layoutFactory->createFullScreenLayout(
```
dasgarner: Sorry to be picky, but we've gone from a very clear variable name - `$draft` to a variable name that doesn't mean anything? What does `$layoutData` mean?

Please can we try to name our variables in a way that makes sense when we read it? 

I think this is `$layout` or `$fullScreenLayout` ?

lib/Listener/LayoutListener.php
```diff
@@ -37,6 +37,7 @@
 use Xibo\Factory\LayoutFactory;
 use Xibo\Factory\PermissionFactory;
 use Xibo\Storage\StorageServiceInterface;
+use Xibo\Support\Exception\GeneralException;
```
dasgarner: I'm not clear what we're using this import for?

dasgarner (commented):
I have a few more comments.

lib/Factory/LayoutFactory.php
mgbaybay: I'm sorry, it was automatically added by the editor I'm using. This [commit](https://github.com/xibosignage/xibo-cms/pull/3133/commits/0664ce5824436298cf5e3a9539cb6c9c8bf251ab) removes all unnecessary imports. Thank you!

lib/XTR/MaintenanceRegularTask.php
mgbaybay: No, we have an existing `filterLayoutStatusId` filter in the query that we use to return either in-use or unused layouts.

dasgarner (dismissed):

dasgarner (approved):

Kent0425:

Kent0425:

Kent0425:
✅Tested!

PR #3139, opened 2025-10-16
Cast autoApplyTransitions and enableStat to int if read as default from DB

movisda:
If default values are read from the database, all values are strings. This can produce wrong quoted json results on API-requests, if the swagger definition has type integer for this variable, e.g.

```
  "enableStat": "1",
  "autoApplyTransitions": "0",
```

This fix produces the correct json response according to the swagger entry:

 ```
  "enableStat": 1,
  "autoApplyTransitions": 0,
```


mgbaybay (approved):

dasgarner:
@mgbaybay thanks for the review. Please can you help create an associated public bug report and internal QA ticket? We also need to re-target to a branch in xibosignage please so that the QA PR containers can be built.

PR #3143, opened 2025-10-23
Parsedown library: Pinned commit to fix PHP 8.4 deprecation warnings

mgbaybay:
## Changes

- Updated Parsedown library to fix deprecated PHP8.4 call

Relates to https://github.com/xibosignage/xibo/issues/3745

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
     "akrabat/ip-address-middleware" : "2.6.*",
     "apereo/phpcas": "dev-master#6ef19a1e636303bda2f05f46b08035d53a8b602d",
     "dragonmantank/cron-expression": "^3.4",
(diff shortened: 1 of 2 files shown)


dasgarner:
Please could you describe why we've had to pin to a commit SHA rather than using a named release?

mgbaybay:
> Please could you describe why we've had to pin to a commit SHA rather than using a named release?

The fix for PHP8.4 isn't in the latest official tagged release yet, only in the master branch. However, pinning to that specific commit ensures we get the deprecation fixes and keep our dependencies consistent until a new stable release is out.

dasgarner (approved):

ifarzana:
TEST: login.cy.js END

PR #3146, opened 2025-10-24
Font Library: Fixed multiple copies of same font in CMS

mgbaybay:
## Changes

- Added logic to validate font uploads to prevent duplicates

Relates to https://github.com/xibosignage/xibo/issues/3746


dasgarner (changes requested):
I can see that this works in the way you've intended (i.e. it allows duplicates by appending a suffix).

However, uploading fonts is quite a specialist use case and you actually do not want to have duplicates, as there is no advantage of having them. Is it useful to have two "Arial" fonts for example? Which one would you choose if you were adding text to a layout?

I think we want this PR to disallow uploading duplicate fonts.

dasgarner (approved):

ifarzana:
TEST: all

ifarzana:

PR #3151, opened 2025-10-29
Logging: XMDS logs showing 1-hour timestamp offset

mgbaybay:
## Changes

- Get the default timezone based on the CMS settings when logging XMDS values

Relates to https://github.com/xibosignage/xibo/issues/3739

### DIFF
--- a/web/xmds.php
+++ b/web/xmds.php
@@ -405,8 +405,12 @@ try {
     }
 
     // logProcessor
+    // Get the default time zone based on the CMS settings and add a fallback value
+    $defaultTimezone = $container->get('configService')->getSetting('defaultTimezone') ?? 'UTC';
     $logProcessor = new \Xibo\Xmds\LogProcessor($container->get('logger'), $uidProcessor->getUid());
-    $container->get('logger')->pushProcessor($logProcessor);
+    $container->get('logger')
+        ->pushProcessor($logProcessor)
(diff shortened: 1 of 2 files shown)


web/xmds.php
```diff
@@ -405,8 +405,12 @@
     }
 
     // logProcessor
+    // Get the default time zone based on the CMS settings and add a fallback value
+    $defaultTimezone = $container->get('configService')->getSetting('defaultTimezone') ?? 'UTC';
```
dasgarner: The `setTimezone` call is right, but I think this is in the wrong place (this will only work for XMDS).

Should we do this everywhere that we call `new Logger(` before then adding that to the DI container?

dasgarner (commented):
I don't really understand what you've done here. Why is it set in both places? Why is the log process block duplicated?

web/xmds.php
```diff
@@ -405,6 +405,7 @@
     }
 
     // logProcessor
+    // Get the default time zone based on the CMS settings and add a fallback value
```
dasgarner: ?

dasgarner (approved):

ifarzana:
TEST: all

ifarzana:
TEST: /Layout/Widget/layout_editor_status_bar.cy.js END

ifarzana:

PR #3158, opened 2025-10-30
Schedule: Fixed CMS does not notify displays after deleting or editing schedules to past dates

mgbaybay:
## Changes

- Updated `inScheduleLookAhead()` to check also events that have been updated to past dates

Relates to https://github.com/xibosignage/xibo/issues/3752

### DIFF
--- a/lib/Entity/Schedule.php
+++ b/lib/Entity/Schedule.php
@@ -938,21 +938,15 @@ class Schedule implements \JsonSerializable
 
         // Notify
         if ($options['notify']) {
-            // Only if the schedule effects the immediate future - i.e. within the RF Look Ahead
-            if ($this->inScheduleLookAhead()) {
-                $this->getLog()->debug(
-                    'Schedule changing is within the schedule look ahead, will notify '
-                    . count($this->displayGroups) . ' display groups'
-                );
-                foreach ($this->displayGroups as $displayGroup) {
-                    /* @var DisplayGroup $displayGroup */
-                    $this
-                        ->getDisplayNotifyService()
-                        ->collectNow()
-                        ->notifyByDisplayGroupId($displayGroup->displayGroupId);
-                }
-            } else {
-                $this->getLog()->debug('Schedule changing is not within the schedule look ahead');
+            // Notify affected displays to clear display cache
+            $this->getLog()->debug('Event update: Will notify ' . count($this->displayGroups) . ' display groups');
+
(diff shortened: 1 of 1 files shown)


lib/Entity/Schedule.php
```diff
@@ -938,21 +938,15 @@ public function save($options = [])
 
         // Notify
         if ($options['notify']) {
-            // Only if the schedule effects the immediate future - i.e. within the RF Look Ahead
-            if ($this->inScheduleLookAhead()) {
```
dasgarner, line 942: This means that any schedule add/edit or delete will cause a notify, which will radically increase load on the server. This can't be the solution unfortunately 😃 

Imagine you're forward planning and in a months time you schedule an event on 500 displays - they will all immediately connect to the CMS as a result of removing this.

lib/Entity/Schedule.php
```diff
@@ -1016,20 +1010,18 @@ public function delete($options = [])
 
         // Notify
         if ($options['notify']) {
-            // Only if the schedule effects the immediate future - i.e. within the RF Look Ahead
-            if ($this->inScheduleLookAhead() && $this->displayNotifyService !== null) {
```
dasgarner, line 1020: Again, can't remove look ahead for the reasons above.

dasgarner (changes requested):

lib/Entity/Schedule.php
```diff
@@ -880,6 +890,8 @@ public function save($options = [])
             $this->validate();
         }
 
+        $originalEvent = $this->getOriginalEventDetails();
```
dasgarner: This is the original event?

lib/Entity/Schedule.php
```diff
@@ -552,6 +553,9 @@ private function inScheduleLookAhead()
         // Dial current date back to the start of the day
         $currentDate->startOfDay();
 
+        // Is the event updated to a past date?
+        $isEventInPast = ($originalEvent !== null && $this->toDt <= $originalEvent['toDt']);
```
dasgarner: I think what you're looking for here is `$this->getOriginalValue('toDt')` ?

`getOriginalValue()` returns you the value of any property as it was when this object was loaded from the database.

dasgarner (changes requested):

dasgarner (approved):

ifarzana:
TEST: all

ifarzana:

PR #3165, opened 2025-11-12
Usage Stat: Capture last user login date

mgbaybay:
## Changes

- Added `dateSinceLastUserLogin` params in anonymous usage stat
- Added `getDateSinceLastUserLogin()` to get the most recent date any user log in happened, converted to UTC timestamp 

Relates to https://github.com/xibosignageltd/xibo-private/issues/1049

### DIFF
--- a/lib/XTR/AnonymousUsageTask.php
+++ b/lib/XTR/AnonymousUsageTask.php
@@ -313,10 +313,10 @@ class AnonymousUsageTask implements TaskInterface
     private function getDateSinceLastUserLogin(): string|null
     {
         $cmsTimezone = $this->getConfig()->getSetting('defaultTimezone');
-        $latestUserLoginDate = $this->runQuery('
-            SELECT MAX(`startTime`) AS startTime FROM `session_history` WHERE `userId` IS NOT NULL AND `userId` <> 0',
+        $latestUserLoginDate = $this->runQuery(
+            'SELECT MAX(`lastAccessed`) AS lastAccessed FROM `user`',
             [],
-            'startTime'
+            'lastAccessed'
         );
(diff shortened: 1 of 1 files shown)


lib/XTR/AnonymousUsageTask.php
```diff
@@ -302,4 +305,22 @@ private function runQuery(string $sql, array $params = [], string $property = 'c
...
+
+    /**
+     * Get the most recent user login timestamp converted to UTC
+     * @return string|null
+     */
+    private function getDateSinceLastUserLogin(): string|null
+    {
+        $cmsTimezone = $this->getConfig()->getSetting('defaultTimezone');
+        $latestUserLoginDate = $this->runQuery('
+            SELECT MAX(`startTime`) AS startTime FROM `session_history` WHERE `userId` IS NOT NULL AND `userId` <> 0',
```
dasgarner: This isn't wrong, but we have a `lastAccessed` field on the user which is updated with this information - it might be more efficient to do that?

mgbaybay: Thank you, @dasgarner! I was looking at the sessions and missed the user info part. I've pushed a [commit ](https://github.com/xibosignage/xibo-cms/pull/3165/commits/ca6b8a8c00df82f36f8233eef9e04a5950fe9a79)that reflects the changes. I've also removed the `WHERE` clause since I believe those would be redundant. Thanks again! :)

dasgarner (approved):

Kent0425:
TEST: /Administration/tasks.cy.js END

Kent0425:
✅Tested!

PR #3167, opened 2025-11-14
Fix logic errors in xibo-cms.js

Iampineda:
This PR fixes three small logic errors in `ui/src/core/xibo-cms.js`:

1. Assignment operator - Line ~856 uses `=` instead of a comparison (`==` or `===`) inside a conditional.
2. Undefined variable - Line ~157 references `e` although the function parameter is `env`.
3. Undeclared variable - Line ~1705 assigns to `formConfirm` without declaring it, which implicitly creates a global variable.

### DIFF
--- a/ui/src/core/xibo-cms.js
+++ b/ui/src/core/xibo-cms.js
@@ -154,7 +154,7 @@ window.XiboInitialise = function(scope, options) {
     .on('mouseenter', function(ev) {
       const formUrl = $(ev.currentTarget).attr('href');
 
-      XiboHoverRender(formUrl, e.pageX, e.pageY);
+      XiboHoverRender(formUrl, env.pageX, env.pageY);
 
       return false;
     }).on('mouseleave', function() {
@@ -853,7 +853,7 @@ window.XiboInitialise = function(scope, options) {
 
                       bodyElements.forEach((element) => {
                         const elementSplit = element.split('=');
-                        if (elementSplit.length = 2) {
+                        if (elementSplit.length == 2) {
                           newParsedElements[elementSplit[0]] = elementSplit[1];
                         }
                       });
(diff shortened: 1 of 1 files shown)


ui/src/core/xibo-cms.js
```diff
@@ -154,7 +154,7 @@ window.XiboInitialise = function(scope, options) {
     .on('mouseenter', function(ev) {
       const formUrl = $(ev.currentTarget).attr('href');
 
-      XiboHoverRender(formUrl, e.pageX, e.pageY);
+      XiboHoverRender(formUrl, env.pageX, env.pageY);
```
maurofmferrao: Instead of `env`, it should be `ev` here.

ui/src/core/xibo-cms.js
```diff
@@ -1702,7 +1702,7 @@ window.XiboMultiSelectFormRender = function(button) {
         formOpenCallback = $button.data().formCallback;
 
         // If form needs confirmation
-        formConfirm = $button.data().formConfirm;
+        let formConfirm = $button.data().formConfirm;
```
maurofmferrao: `formConfirm` indeed needs to be initialised, but we need to do it on line 1690 next to the `let formOpenCallback = null;` initialisation. That's because this line will only init for the block scope inside the `each` loop.

maurofmferrao (changes requested):
Thank you for your contribution @Iampineda!

I've just requested some changes that need to be done before we can merge.

Iampineda:
Updated as requested: switched to using ev.pageX/ev.pageY and moved the formConfirm initialization next to formOpenCallback.

dasgarner:
Thank you for sending us an email indicating that you agree to the CLA. For a small fix like this one, we do not require you to follow the CLA process.

If you are planning on making further contributions, you will need to submit the CLA as per the instructions [here](https://github.com/xibosignage/xibo?tab=contributing-ov-file#contributor-licence-agreement-cla). Please engage with us before making larger changes as we're already actively working on some big changes in the codebase.

ifarzana:
TEST: all

PR #3168, opened 2025-11-14
Update the instance suspended setting to have 3 options.

dasgarner:
This PR expands the INSTANCE_SUSPENDED setting so that it has 3 options: yes, no, partial.

When it is set to partial, the XMDS and XTR should be disabled, and a banner shown in the CMS. When set to yes, the CMS UI should also be disabled.

fixes to xibosignageltd/xibo-private#1048

### DIFF
--- a/views/authed.twig
+++ b/views/authed.twig
@@ -95,7 +95,7 @@
                         {% block actionMenu %}{% endblock %}
 
                         {% if settings.INSTANCE_SUSPENDED == "partial" %}
-                            <div class="alert alert-warning">{{ "CMS suspended. Displays will showed cached content. Please contact your administrator."|trans }}</div>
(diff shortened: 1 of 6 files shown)


dasgarner:
@mgbaybay just checking in on this review

views/authed.twig
```diff
@@ -93,6 +93,11 @@
                 <div class="row">
                     <div class="col-sm-12">
                         {% block actionMenu %}{% endblock %}
+
+                        {% if settings.INSTANCE_SUSPENDED == "partial" %}
```
mgbaybay, line 97: @dasgarner should we also add the "yes" option in the condition?

dasgarner: It's a good point, but the whole user interface is locked out at this point, so it shouldn't matter.

mgbaybay (dismissed):

mgbaybay (approved):

ifarzana:
TEST: login.cy.js END

PR #3174, opened 2025-11-17
Media: when saving a file ensure the original file name is not too long

### LINKED ISSUE #3771: Widget sync task: fails for RSS widget with long image URLs
CMS version: 4.3.1

The widget sync task fails when adding an RSS widget that contains images with very long filenames in their URLs. The following error is observed:

```
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'originalFileName' at row 1
```

Steps to reproduce:

- Add an RSS widget using a feed that contains images with very long filenames in their URLs.
- Trigger the widget sync task.
- Observe the error in the logs.

Expected behaviour:
The widget sync task should complete successfully, even when images in the RSS feed have long filenames.

Actual behaviour:
The task fails because the media filename from the RSS feed exceeds the database column length for originalFileName.

dasgarner:
fixes xibosignage/xibo#3771

### DIFF
--- a/lib/Entity/Media.php
+++ b/lib/Entity/Media.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2024 Xibo Signage Ltd
+ * Copyright (C) 2025 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - https://xibosignage.com
  *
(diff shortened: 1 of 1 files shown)


mgbaybay (approved):

lykqa:
TEST: Library/media.cy.js END

PR #3177, opened 2025-11-19
Interactivity: Fixed action validation on layout import

mgbaybay:
## Changes

- Updated `manageActions` to skip action validations during layout import. This is to allow users to manage the imported layouts in a different CMS that might have a different environment or settings.

Relates to https://github.com/xibosignage/xibo/issues/3765


dasgarner (approved):

lykqa:
TEST: Layout/Editor/layout-action-menu.cy.js END

PR #3188, opened 2025-11-26
Schedule Criteria: Weather criteria shows all condition options instead of only applicable ones

nadzpogi:
relates to xibosignage/xibo#3754


mgbaybay (approved):

lykqa:
TEST: Schedule/schedule.cy.js END

PR #3190, opened 2025-11-26
RSS: Characters are not read properly

maurofmferrao:
relates to xibosignage/xibo#3779


dasgarner (commented):
Are we sure that we want to accept any HTML characters from the title/summary? This would mean that HTML would be allowed completely, which seems risky to me?

Would it be better to decode HTML entities, but not actual HTML?

PR #3192, opened 2025-11-27
Add Schedule API: Fixed mediaId not documented and CMS auto-assigns the wrong media

mgbaybay:
## Changes

- Added validation on Add Schedule API for media event types

Relates to https://github.com/xibosignage/xibo/issues/3777

### DIFF
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -1195,8 +1195,7 @@ class Schedule extends Base
 
             if (!$id) {
                 throw new InvalidArgumentException(
-                    __(ucfirst($type) . 'Id is required when scheduling ' . $type . ' events.'),
-                    $type . 'Id'
+                    sprintf('%sId is required when scheduling %s events.', ucfirst($type), $type)
                 );
             }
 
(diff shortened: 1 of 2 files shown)


lib/Controller/Schedule.php
```diff
@@ -1186,6 +1193,13 @@ public function add(Request $request, Response $response)
             $type = $schedule->eventTypeId === \Xibo\Entity\Schedule::$MEDIA_EVENT ? 'media' : 'playlist';
             $id = ($type === 'media') ? $sanitizedParams->getInt('mediaId') : $sanitizedParams->getInt('playlistId');
 
+            if (!$id) {
+                throw new InvalidArgumentException(
+                    __(ucfirst($type) . 'Id is required when scheduling ' . $type . ' events.'),
```
dasgarner: When we translate messages that have substitutions in them, we should use variables and `sprintf`. This is because different languages may need those variables to be in a different position in the string.

dasgarner (changes requested):

lib/Controller/Schedule.php
mgbaybay: Thank you, Dan! I've pushed a [commit](https://github.com/xibosignage/xibo-cms/pull/3192/commits/1bf752f2ad2c8c3765aa1bc5ee988ea016f1523c) to fix this

dasgarner (approved):

Kent0425:

Kent0425:
✅Tested!

PR #3199, opened 2025-12-03
Dataset: Disable isRealTime and dataConnectorSource option via API

mgbaybay:
## Changes

- Disable isRealTime and dataConnectorSource option via API

Relates to https://github.com/xibosignageltd/xibo-private/issues/1086

### DIFF
--- a/lib/Controller/DataSet.php
+++ b/lib/Controller/DataSet.php
@@ -681,9 +681,13 @@ class DataSet extends Base
         $dataSetColumn->dataSetColumnTypeId = 1;
         $dataSetColumn->dataTypeId = 1;
 
-        // Add column, real time and data connector settings when we are not routing through the API
+        // Add Column
+        // only when we are not routing through the API
         if (!$this->isApi($request)) {
             $dataSet->assignColumn($dataSetColumn);
+        }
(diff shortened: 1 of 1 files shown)


lib/Controller/DataSet.php
```diff
@@ -683,10 +681,11 @@ public function add(Request $request, Response $response)
         $dataSetColumn->dataSetColumnTypeId = 1;
         $dataSetColumn->dataTypeId = 1;
 
-        // Add Column
-        // only when we are not routing through the API
+        // Add column, real time and data connector settings when we are not routing through the API
         if (!$this->isApi($request)) {
```
dasgarner, line 686: Apologies I wasn't clear. These properties still need to be available through the API.

What I meant is that they should be disabled when the user has the feature flag for it disabled.

dasgarner (changes requested):

dasgarner (approved):

Kent0425:
TEST: /Display/datasets.cy.js END

Kent0425:
TEST: login.cy.js END

Kent0425:
✅Tested!

PR #3200, opened 2025-12-08
Dataset: Disable access to dataset dataconnector for users without feature access

mgbaybay:
## Changes

- Disable access to dataset data connector for users without feature access

Relates to https://github.com/xibosignageltd/xibo-private/issues/1093

### DIFF
--- a/lib/Controller/DataSet.php
+++ b/lib/Controller/DataSet.php
@@ -1074,9 +1074,12 @@ class DataSet extends Base
         $dataSet = $this->dataSetFactory->getById($id);
         $sanitizedParams = $this->getSanitizer($request->getParams());
 
-        if (!$this->getUser()->checkEditable($dataSet)
-            || !$this->getUser()->featureEnabled('dataset.data')) {
-            throw new AccessDeniedException(__('Feature is disabled'));
+        if (!$this->getUser()->checkEditable($dataSet)) {
+            throw new AccessDeniedException();
+        }
+
+        if (!$this->getUser()->featureEnabled('dataset.data')) {
+            throw new AccessDeniedException(__('Feature not enabled'));
         }
 
         if ($dataSet->isRealTime === 1) {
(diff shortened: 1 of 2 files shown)


lib/Middleware/FeatureAuth.php
```diff
@@ -76,7 +76,7 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             }
         }
 
-        throw new AccessDeniedException(__('Feature not enabled'), __('This feature has not been enabled for your user.'));
+        throw new AccessDeniedException(__('Feature is disabled'), __('This feature has been disabled for your user.'));
```
dasgarner: Although I see why you might want to change this, changing translations should only be done if really necessary as it invalidate all of our currently translated instances of this string.

Unless there is something I am not aware of, I don't see this as being a necessary change?

lib/Controller/DataSet.php
```diff
@@ -1074,8 +1074,9 @@ public function updateDataConnector(Request $request, Response $response, $id)
         $dataSet = $this->dataSetFactory->getById($id);
         $sanitizedParams = $this->getSanitizer($request->getParams());
 
-        if (!$this->getUser()->checkEditable($dataSet)) {
-            throw new AccessDeniedException();
+        if (!$this->getUser()->checkEditable($dataSet)
+            || !$this->getUser()->featureEnabled('dataset.data')) {
+            throw new AccessDeniedException(__('Feature is disabled'));
```
dasgarner: I think this message is confusing? Not having access due to missing edit sharing rights is different from the feature not being enabled?

dasgarner (changes requested):

lib/Middleware/FeatureAuth.php
mgbaybay: Thank you, Dan! Sorry I did not consider the possible impact, I've reverted the changes.

dasgarner (approved):

ifarzana:
TEST: all

ifarzana:

PR #3201, opened 2025-12-08
DataSet Data Connector: Sandbox isolation for testing environment

dasgarner:
This PR adds iframe sandbox isolation to the DataSet data connecter testing page.

Now that the testing page is in a different origin, we need to use postMessage to exchange information.

It also removes the ability to use the CMS as a proxy route for making an HTTP request. This means that CORS will need to be configured for testing HTTP data sources because it will request using browser instead.

fixes xibosignageltd/xibo-private#1101

### DIFF
--- a/lib/Controller/DataSet.php
+++ b/lib/Controller/DataSet.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2024 Xibo Signage Ltd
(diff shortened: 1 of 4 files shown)


maurofmferrao (approved):
Looks good to me!

ifarzana:
TEST: /Library END

PR #3206, opened 2025-12-10
Deleted Recurring Event still shows on schedule calendar fix

maurofmferrao:
relates to xibosignageltd/xibo-private#1088

### DIFF
--- a/lib/Controller/Schedule.php
+++ b/lib/Controller/Schedule.php
@@ -2554,6 +2554,11 @@ class Schedule extends Base
                         !empty($repeatsUntil) ? ' until ' . $repeatsUntil : ''
                     ))
                 );
+
+                $event->setUnmatchedProperty(
+                    'scheduleExclusions',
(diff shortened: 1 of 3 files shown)


lib/Controller/Schedule.php
```diff
@@ -2554,6 +2554,11 @@ public function grid(Request $request, Response $response)
                         !empty($repeatsUntil) ? ' until ' . $repeatsUntil : ''
                     ))
                 );
+
+                $event->setUnmatchedProperty(
+                    'scheduleExclusions',
+                    $this->scheduleExclusionFactory->query(null, ['eventId' => $event->eventId])
```
dasgarner, line 2560: This will be too intensive to return without paging. We will need a different strategy to do it.

dasgarner (changes requested):

dasgarner (approved):

ifarzana:
TEST: /Schedule END

PR #3209, opened 2025-12-18
Preview: isolating user supplied JavaScript

dasgarner:
- Start separating out the preview routes from the main web routes.
 - Rework library public thumbnails to use new preview route
 - Add sandbox to all preview iframes.
 - Implement JWT token as well as S3 style
 - Add layoutId into JWT and check auth


maurofmferrao (dismissed):
Looks good!

PR #3210, opened 2025-12-19
4.4 schedule page performance changes

maurofmferrao:
relates to xibosignageltd/xibo-private#1116

### DIFF
--- a/ui/src/core/xibo-calendar.js
+++ b/ui/src/core/xibo-calendar.js
@@ -21,9 +21,13 @@
 
 // Global calendar object
 window.calendar = undefined;
+window.calendarEnabled = false;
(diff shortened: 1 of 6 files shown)


dasgarner (commented):
These changes look good, will we also remove the no paging option from the API?

maurofmferrao:
> These changes look good, will we also remove the no paging option from the API?

~~We can remove it, or we can keep it so it can be used on another page/grid if needed, what do you think?~~

Disable paging is actually being used by a fix for displaying Actions: https://github.com/xibosignage/xibo/issues/3768

dasgarner (approved):

lykqa:
TEST: Schedule/schedule.cy.js END

PR #3212, opened 2025-12-29
User: Fixed cannot delete oauth user entry

mgbaybay:
## Changes

- Added deletion of user oauth entry in `oauth_lkclientuser` table

Relates to https://github.com/xibosignage/xibo/issues/3785


dasgarner (approved):

lykqa:
TEST: User_Account/user_account.cy.js END

PR #3223, opened 2026-01-08
Preview: connector preview is authed by its own token and therefore can be public in preview routes

dasgarner:
relates to xibosignageltd/xibo-private#1114


maurofmferrao (approved):
Looks good!

dasgarner:
TEST: dashboard.cy.js END

PR #3225, opened 2026-01-12
Media: Data Table changes

mgbaybay:
## Changes

- Implemented a JSON entrypoint checker via `isJson`
- Added response format for JSON via `renderJsonResponse`
- In the existing `gridRenderSort`, added a flag to get the `sortBy` and `sortDir` in the parameters
- Added `buildSortQuery` to map sortable columns and sanitize the parameters
- Removed buttons
- Updated swagger docs

Relates to https://github.com/xibosignageltd/xibo-private/issues/1100


lib/Controller/Library.php
```diff
@@ -668,6 +671,11 @@ public function grid(Request $request, Response $response)
 
             $media->buttons = [];
 
+            // Skip the buttons for now and do it once routing is updated
+            if ($this->isJson($request)) {
```
dasgarner: Speak to @maurofmferrao, but we will want to add properties for current user permissions, e.g. canView, canEdit, canDelete, canModifyPermissions

$media->buttons can be removed

dasgarner (approved):

PR #3229, opened 2026-01-16
Fetature/neujmin media components

chaolen:
Created new components: https://github.com/xibosignageltd/xibo-private/issues/1104

image

Also added Input Filters (some select option are just sample texts)

image

### DIFF
--- a/frontend/src/components/ui/Button.tsx
+++ b/frontend/src/components/ui/Button.tsx
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2026 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+.*/
+
+import { type LucideIcon } from 'lucide-react';
+
+type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'link';
+  icon?: LucideIcon;
+  removeTextOnMobile?: boolean;
+  ariaLabel?: string;
+};
+
+function Button({
+  variant = 'primary',
+  icon: Icon,
+  children,
+  className,
+  removeTextOnMobile = false,
+  ariaLabel,
+  ...props
+}: ButtonProps) {
+  const showText = Boolean(children);
+
+  return (
+    <button
+      type="button"
+      className={`button ${variant} ${className ?? ''}`}
+      aria-label={!showText ? ariaLabel : undefined}
+      {...props}
+    >
+      {Icon && <Icon size={21} />}
+
+      {showText && (
+        <span className={removeTextOnMobile ? 'hidden sm:inline' : undefined}>{children}</span>
+      )}
+    </button>
+  );
+}
+
+export default Button;
--- a/frontend/src/components/ui/Button.tsx
+++ b/frontend/src/components/ui/Button.tsx
@@ -38,11 +38,21 @@ function Button({
   ...props
 }: ButtonProps) {
   const showText = Boolean(children);
+  const buttonVariant: Record<NonNullable<ButtonProps['variant']>, string> = {
+    primary: 'text-white bg-xibo-blue-600 hover:bg-xibo-blue-700 focus:ring-4',
+    secondary:
+      'text-xibo-blue-600 border border-xibo-blue-600 bg-white hover:border-xibo-blue-800 hover:text-xibo-blue-800',
+    tertiary: 'text-xibo-blue-600 bg-gray-50 hover:bg-gray-100 hover:text-xibo-blue-800',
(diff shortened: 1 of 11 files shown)


chaolen:
Hi @maurofmferrao, changes to the Button Component is now good for checking.

frontend/src/components/ui/Button.tsx
```diff
@@ -0,0 +1,68 @@
...
+  variant = 'primary',
+  icon: Icon,
+  children,
+  className,
+  removeTextOnMobile = false,
+  ariaLabel,
+  ...props
+}: ButtonProps) {
+  const showText = Boolean(children);
+  const buttonVariant: Record<NonNullable<ButtonProps['variant']>, string> = {
```
maurofmferrao: Maybe we should move constants like `buttonVariant` and `baseClasses` outside the component function? That might be a good performance win.

frontend/src/components/ui/Button.tsx
```diff
@@ -0,0 +1,68 @@
...
+    link: 'text-xibo-blue-600 underline hover:text-xibo-blue-800 bg-transparent',
+  };
+
+  const baseClasses =
+    'p-2 rounded-lg gap-2 text-[14px] text-center leading-[0.07px] tracking-[150%] flex items-center h-[45px] box-border cursor-pointer font-semibold focus:ring-blue-500/25';
+
+  return (
+    <button
+      type="button"
+      className={`${baseClasses} ${buttonVariant[variant]} ${className}`}
```
maurofmferrao: If we're going to merge Tailwind classes, we probably need a library like `tailwind-merge` (https://www.npmjs.com/package/tailwind-merge?activeTab=readme) to avoid conflicts between classes?

frontend/src/components/ui/Button.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const buttonVariant: Record<NonNullable<ButtonProps['variant']>, string> = {
+    primary: 'text-white bg-xibo-blue-600 hover:bg-xibo-blue-700 focus:ring-4',
+    secondary:
+      'text-xibo-blue-600 border border-xibo-blue-600 bg-white hover:border-xibo-blue-800 hover:text-xibo-blue-800',
+    tertiary: 'text-xibo-blue-600 bg-gray-50 hover:bg-gray-100 hover:text-xibo-blue-800',
+    link: 'text-xibo-blue-600 underline hover:text-xibo-blue-800 bg-transparent',
+  };
+
+  const baseClasses =
+    'p-2 rounded-lg gap-2 text-[14px] text-center leading-[0.07px] tracking-[150%] flex items-center h-[45px] box-border cursor-pointer font-semibold focus:ring-blue-500/25';
```
maurofmferrao: Is the value `leading-[0.07px]` correct? A line height of almost 0 will crush the text into a line 🤔 (https://tailwindcss.com/docs/line-height) 

From Win's designs for button, the CSS in Figma shows that value for `letter-spacing`, I think using `text-sm` and `font-semibold` classes are enough for the `App/text-sm-semibold` ( can also replace `text-[14px]` ).

```
/* App/text-sm-semibold */
font-family: Inter;
font-size: 14px;
font-style: normal;
font-weight: 600;
line-height: 150%; /* 21px */
letter-spacing: 0.07px;
```

frontend/src/components/ui/SelectFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+          >
+            {option.label}
+          </li>
+        ))}
+      </ul>
+    </div>
+  );
+};
+
+export default SelectFilter;
```
maurofmferrao: Should we use the `export default function ComponentName` pattern on all components like we're doing on the TopNav, for consistency?

frontend/src/components/ui/SelectFilter.tsx
```diff
@@ -0,0 +1,111 @@
...
+  useEffect(() => {
+    const handleClickOutside = (e: MouseEvent) => {
+      if (ref.current && !ref.current.contains(e.target as Node)) {
+        setOpen(false);
+      }
+    };
+
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => document.removeEventListener('mousedown', handleClickOutside);
+  }, []);
```
maurofmferrao, line 62: Should we add the variable `open` to the dependency array here and a check on top of the function checking if the filter is opened?

E.g.

```
useEffect(() => {
    if (!open) {
      return;
    }

    const handleClickOutside = (e: MouseEvent) => {
      if (ref.current && !ref.current.contains(e.target as Node)) {
        setOpen(false);
      }
    };

    document.addEventListener('mousedown', handleClickOutside);
    return () => document.removeEventListener('mousedown', handleClickOutside);
  }, [open]);
  ```

File: frontend/src/components/ui/media/MediaTopNav.tsx [hunk 6]
```diff
@@ -0,0 +1,71 @@
...
+        {navigation.map((tab) => (
+          <button
+            key={tab}
+            type="button"
+            className={`py-2 text-[14px] px-1 inline-flex text-gray-500 items-center gap-x-2 border-b-2 whitespace-nowrap focus:outline-none transition-colors font-semibold ${
+              tab === activeTab
+                ? 'text-xibo-blue-600 border-xibo-blue-600'
+                : 'border-transparent  hover:text-gray-600 hover:border-gray-300'
+            }`}
+            aria-current={tab === 'Media' ? 'page' : undefined}
```
maurofmferrao: We're using `Media` string here, are we sure this won't fail if it's being used in a different language?

maurofmferrao (changes requested):
Some changes. mostly for code consistency ( I've been doing the same to the code I wrote previously xD). Everything else looks good 😃

frontend/src/components/ui/SelectFilter.tsx
chaolen: Hmmm.. I think the current version is good enough. I don’t think we need to add `open` because the effect’s only responsibility is to close the component on outside click, and it doesn’t depend on open.

chaolen:
Changes applied now, @maurofmferrao

maurofmferrao (approved):
Looks good to me!

PR #3230, opened 2026-01-19
Media: Add user permission and feature access

mgbaybay:
## Changes

- Added user permissions per media item
- Added user features

Relates to https://github.com/xibosignageltd/xibo-private/issues/1155

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -692,47 +692,11 @@ class Library extends Base
 
             // User permissions
             $userPermissions = [
-                'edit'          => false,
-                'copy'          => false,
-                'enableStat'    => false,
-                'delete'        => false,
-                'share'         => false,
-                'folderView'    => false,
-                'usageReport'   => false,
-                'schedule'      => false,
-                'download'      => true, // No feature permissions here, anyone can get a file based on sharing.
+                'view'      => $user->checkViewable($media),
+                'edit'      => $user->checkEditable($media),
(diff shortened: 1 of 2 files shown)


lib/Controller/Library.php
```diff
@@ -689,6 +689,52 @@ public function grid(Request $request, Response $response)
                 $media->setUnmatchedProperty('hasFullScreenLayout', (!empty($fullScreenCampaignId)));
                 $media->setUnmatchedProperty('fullScreenCampaignId', $fullScreenCampaignId);
             }
+
+            // User permissions
+            $userPermissions = [
```
dasgarner, line 694: Sorry, but I think this still mixes concerns - because if we needed to add a new row menu item in the front end we would also have to come and update this array in the back end?

I think it would be better to output what the back end knows and understands, i.e. the user add, edit, delete and modifySharing permissions (being the 4 things we know in the back end).

And then in the /user/me route we should output the users access to feature flags.

Then the front end decided whether or not the user can see the folder view or schedule, etc. @maurofmferrao  does that make sense to you?

dasgarner (changes requested):

lib/Controller/Library.php
maurofmferrao: Yeah, I think it makes sense to send the information we have on the back-end ( the 4 flags only), and handle the rest on the front-end for each feature/option, even if they share the same flag.

maurofmferrao (approved):
Looks good to me!

PR #3232, opened 2026-01-19
Table Design: First version

maurofmferrao:
relates to xibosignageltd/xibo-private#1102

- App structured refactored
- usePreline, useClickOutside and useCloseOnScroll hooks added
- Inter font installed
- React compiler added


chaolen (changes requested):
Table looks good. One thing I notice is column visibility or the DataTableOptions checkboxes don’t toggle even though the table updates. The check icon also does not update onChange.

frontend/src/components/ui/table/DataTableOptions.tsx
```diff
@@ -98,8 +108,18 @@ export function DataTableViewOptions<TData>({ table }: DataTableViewOptionsProps
               })}
             </div>
```
chaolen, line 113: I can't highlight the specific line 😅 But it's this line:

`onChange={(e) => column.toggleVisibility(!!e.target.checked)}`

It appears to prevent the checkbox from toggling correctly, causing the UI to appear stuck.

maurofmferrao:
> Table looks good. One thing I notice is column visibility or the DataTableOptions checkboxes don’t toggle even though the table updates. The check icon also does not update onChange.

I've added a new commit with new components and the table mostly done, with only the Media Preview, No Results message, Multi Col sort and that bug you've found.

I think we should merge this PR as soon as possible since there are a lot of structural changes on the Media page, so we should all work on top of it, and I'll then create a new PR to finish the table, if you agree.

chaolen (approved):
LGTM 🚀

frontend/src/components/ui/table/DataTableOptions.tsx
maurofmferrao: Yeah, but we need to have the checkbox status rely on the table visibility for that item, for some reason DataTableOptions isn't updating the render when we change the table value, and it should, I will fix it in a later PR. 😃

PR #3241, opened 2026-01-22
Media: Added email and folder name on details modal

mgbaybay:
## Changes

- Added folder name `folderName` and user email `userEmail` for the details modal

Relates to https://github.com/xibosignageltd/xibo-private/issues/1147

### DIFF
--- a/lib/Factory/MediaFactory.php
+++ b/lib/Factory/MediaFactory.php
@@ -552,6 +552,8 @@ class MediaFactory extends BaseFactory
                `media`.width,
                `media`.height,
                `user`.UserName AS owner,
+               `user`.email AS userEmail,
+               `folder`.folderName AS location,
             ';
         $select .= '     (SELECT GROUP_CONCAT(DISTINCT `group`.group)
                               FROM `permission`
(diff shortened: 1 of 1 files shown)


lib/Factory/MediaFactory.php
```diff
@@ -552,6 +552,8 @@ public function query($sortOrder = null, $filterBy = [])
                `media`.width,
                `media`.height,
                `user`.UserName AS owner,
+               `user`.email AS userEmail,
+               `folder`.folderName AS location,
```
dasgarner: Do you think this terminology might be confusing? Location in the CMS typically refers to geography currently (ie.. Location in the weather widget, geolocation, etc).

Wouldn't it be clearer to just call it folderName?

dasgarner (changes requested):

lib/Factory/MediaFactory.php
mgbaybay: @dasgarner Thank you, Dan! I agree that it would be more intuitive to use folderName. I've pushed a [commit](https://github.com/xibosignage/xibo-cms/pull/3241/commits/f71d810de628c61a2a557ec5c91e1806b42772ba) to reflect this change.

dasgarner (approved):

PR #3242, opened 2026-01-22
Table final design

maurofmferrao:
relates to xibosignageltd/xibo-private#1102

 - Media Preview
 - Download Media
 - Col selector fix
 - Table: No data message
 - Sort with more than 1 column


chaolen (approved):
LGTM! 👌

PR #3243, opened 2026-01-22
CSRF: output a cookie

dasgarner:
(SPA CSRF pattern), readable by JavaScript. Axios automatically sends this cookie due to naming convention.
fixes to xibosignageltd/xibo-private#1150


maurofmferrao (approved):
Looks good to me!

PR #3244, opened 2026-01-23
Interactive action: Key Press trigger only supports a single action

maurofmferrao:
relates to xibosignage/xibo#3807


dasgarner (approved):

lykqa:
TEST: Layout/IA/toggle_mode_on_off.cy.js END

PR #3246, opened 2026-01-27
Media: Add fulltext search

mgbaybay:
## Changes

- Added `buildSearchQuery` for fulltext search
- Refactored Media `grid`
- Fixed media sorting to handle table prefix
- Added migration file for media indexing

Relates to https://github.com/xibosignageltd/xibo-private/issues/1146

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2025 Xibo Signage Ltd
+ * Copyright (C) 2026 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - https://xibosignage.com
  *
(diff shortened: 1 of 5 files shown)


lib/Controller/Library.php
```diff
@@ -2892,4 +2747,180 @@ private function updateMediaRevision(Media $media, $folderId)
...
+    /**
+     * Get the media released description
+     * @param $released
+     * @return string
+     */
+    private function getMediaReleasedDescription($released): string
+    {
+        return match ($released) {
+            1 => '',
+            2 => 'The uploaded image is too large and cannot be processed, please use another image.',
```
dasgarner: Have we removed the translations here on purpose?

lib/Factory/MediaFactory.php
```diff
@@ -671,16 +671,10 @@ public function query($sortOrder = null, $filterBy = [])
...
-            $this->nameFilter(
-                'media',
-                'name',
-                $terms,
-                $body,
-                $params,
-                ($sanitizedFilter->getCheckbox('useRegexForName') == 1),
-                $logicalOperator
+            // Fulltext search
+            $body .= $this->buildSearchQuery(
```
dasgarner, line 675: It looks like we've removed the advanced filtering on `name`  (logical operator, and/or etc) in place of the new search. I'm not sure this is right?

I thought the new search would be an addition, not a replacement? I.e. a new `search` parameter.

lib/Controller/Library.php
```diff
@@ -2892,4 +2747,180 @@ private function updateMediaRevision(Media $media, $folderId)
...
+
+    /**
+     * Get the media enable stat description
+     * @param $enableStat
+     * @return string
+     */
+    private function getMediaEnableStatDescription($enableStat): string
+    {
+        return match ($enableStat) {
+            'On' => 'This Media has enable stat collection set to ON',
```
dasgarner: Translations?

dasgarner (changes requested):

dasgarner (approved):

PR #3253, opened 2026-01-29
Implement filters components

maurofmferrao:
relates to xibosignageltd/xibo-private#1144

- Reset button added to filters
- Filters style fix
- user and userGroup api connection created
- user and userGroup api fix for json
- Edit modal scroll fix
- Upload dock title and scroll fix
- tag type moved to target file

### DIFF
--- a/frontend/src/pages/Library/Media/hooks/useMediaFilterOptions.ts
+++ b/frontend/src/pages/Library/Media/hooks/useMediaFilterOptions.ts
@@ -87,7 +87,5 @@ export function useMediaFilterOptions() {
     };
   }, []);
 
-  console.log(filterOptions);
-
   return { filterOptions, isLoading };
 }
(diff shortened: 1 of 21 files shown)


frontend/src/pages/Library/Media/hooks/useMediaFilterOptions.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    loadDynamicOptions();
+
+    return () => {
+      ignore = true;
+    };
+  }, []);
+
+  console.log(filterOptions);
```
chaolen: Looks like this console.log was left in.

chaolen (commented):
Can we add a loader when filter values change and the data is refetched? This would make it clearer to users that filtering is in progress, especially if the request takes a moment.

chaolen (commented):
Looks good to me! 👌

PR #3260, opened 2026-02-05
User: Fixed duplicate media name when assigning media items

mgbaybay:
## Changes

- Append numeric suffix if the target user already has a content with the same name

Relates to https://github.com/xibosignage/xibo/issues/3804

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -892,7 +892,8 @@ class Layout implements \JsonSerializable
             'appendCountOnDuplicate' => false,
             'setModifiedDt' => true,
             'auditMessage' => 'Saved',
-            'type' => null
+            'type' => null,
+            'isLayoutReassigned' => false
         ], $options);
 
         if ($options['validate']) {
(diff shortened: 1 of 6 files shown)


dasgarner (changes requested):
The logic here looks great, but why `_imported` ? Perhaps it would make more sense for them to be `_transferred` ?

mgbaybay:
> The logic here looks great, but why `_imported` ? Perhaps it would make more sense for them to be `_transferred` ?

Thanks, Dan! I've based it on the suggestions in the reported ticket, but I agree that `_transferred` makes more sense so I've updated it :)

dasgarner (dismissed):

ifarzana:
TEST: /Library END

mgbaybay:
@dasgarner requesting your review again on this since upon checking, it seems that we need to apply similar logic to other duplicate contents such as media, layouts and playlist. 

Moreover, I've updated the logic so we can handle the naming gracefully (i.e. appending the correct numeric suffix) instead of just appending `_transferred` to the name.

dasgarner (approved):

ifarzana:
TEST: /Library END

PR #3265, opened 2026-02-10
Added test for: Sidebar

Kent0425:
- https://github.com/xibosignageltd/xibo-private/issues/1154

Added Sidebar tests which includes:
- Check if main menus are present when sidebar is opened
- Check submenus under main menus when sidebar is opened
- Check if sidebar is closed if hamburger icon is clicked
- Check if no main menus are present when sidebar is closed


ifarzana (approved):
Looks good.

PR #3269, opened 2026-02-10
Nav bar permissions

maurofmferrao:
relates to xibosignageltd/xibo-private#1156

### DIFF
--- a/frontend/src/components/layout/SideBar.tsx
+++ b/frontend/src/components/layout/SideBar.tsx
@@ -19,7 +19,7 @@
  * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router-dom';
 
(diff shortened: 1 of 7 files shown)


frontend/src/components/layout/SideBar.tsx
```diff
@@ -43,22 +45,39 @@ export default function SidebarMenu({
...
   const { t } = useTranslation();
+  const { user } = useUserContext();
   const location = useLocation();
   const [openMenu, setOpenMenu] = useState<string | null>(null);
 
+  console.log(user);
+
+  const visibleRoutes = useMemo(() => {
+    return user ? filterRoutesByUser(APP_ROUTES, user) : [];
+  }, [user]);
```
chaolen, line 54: Looks like a `console.log` was left here 😊
Also just checking, is the `useMemo` intentional? I remember we were avoiding it with React 19.

chaolen (changes requested):
Just a minor request for changes.

frontend/src/components/layout/SideBar.tsx
maurofmferrao: Oh, I missed that console.log, it's just some debugging that I forgot to clear. I'm going to remove it 😄 

As for the useMemo, in this case I had to use since I couldn't find a way to use the visibleRoutes in the useEffect after without triggering a warning from ESLint.

chaolen (approved):
LGTM! 🎉

PR #3270, opened 2026-02-11
Layouts: Fixed preview URL for CMS alias or subdirectory

mgbaybay:
## Changes

- When using the preview URL, explicitly add the CMS alias or subdirectory so that it correctly renders the widget elements
- Updated the preview URL to handle CMS alias or subdirectory

Relates to https://github.com/xibosignage/xibo/issues/3815

### DIFF
--- a/lib/Controller/Campaign.php
+++ b/lib/Controller/Campaign.php
@@ -1289,7 +1289,8 @@ class Campaign extends Base
         $layouts = $this->layoutFactory->getByCampaignId($id);
         $duration = 0 ;
         $extendedLayouts = [];
-        $baseUrl = (new HttpsDetect())->getBaseUrl($request);
+        // Include the CMS alias or subdirectory to get the correct preview url
+        $baseUrl = (new HttpsDetect())->getBaseUrl($request) . rtrim(dirname($_SERVER['SCRIPT_NAME']), '/');
 
         foreach ($layouts as $layout) {
             $duration += $layout->duration;
--- a/lib/Controller/Campaign.php
+++ b/lib/Controller/Campaign.php
@@ -1289,8 +1289,7 @@ class Campaign extends Base
         $layouts = $this->layoutFactory->getByCampaignId($id);
         $duration = 0 ;
         $extendedLayouts = [];
-        // Include the CMS alias or subdirectory to get the correct preview url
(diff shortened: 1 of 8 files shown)


lib/Controller/Campaign.php
```diff
@@ -1289,7 +1289,8 @@ public function preview(Request $request, Response $response, $id)
         $layouts = $this->layoutFactory->getByCampaignId($id);
         $duration = 0 ;
         $extendedLayouts = [];
-        $baseUrl = (new HttpsDetect())->getBaseUrl($request);
+        // Include the CMS alias or subdirectory to get the correct preview url
+        $baseUrl = (new HttpsDetect())->getBaseUrl($request) . rtrim(dirname($_SERVER['SCRIPT_NAME']), '/');
```
dasgarner: I don't think handling this on a case by case basis is the right thing to do. `getBaseUrl` says that it should:

// Get the base URL for the instance
//  this should give us the CMS URL including alias and file

Which appears to not be working?

The question is why doesn't work work? Does it only not work when in the preview entrypoint? Can we make it work always? etc.

dasgarner (changes requested):

web/.htaccess
```diff
@@ -24,8 +24,7 @@ RewriteRule ^ install/index.php [QSA,L]
 
 # preview
 RewriteCond %{REQUEST_FILENAME} !-f
-RewriteCond %{REQUEST_URI} ^/preview/.*$
-RewriteRule ^ preview/index.php [QSA,L]
+RewriteRule ^preview/(.*)$ preview/index.php [QSA,L]
```
dasgarner: Can you explain the change here? None of our other entrypoints are like this, so it seems odd?

web/.htaccess
```diff
@@ -23,9 +23,9 @@ RewriteCond %{REQUEST_URI} ^/install/.*$
 RewriteRule ^ install/index.php [QSA,L]
 
 # preview
+# preview is routed through the web, thus, matching must include the alias
```
dasgarner: I'm still not clear why this is different to any of the other declarations here - all of them are accessible via an alias, but only this one has changed.

dasgarner (dismissed):

dasgarner (approved):

ifarzana:

PR #3271, opened 2026-02-11
Implement copy functionality on Media Page

chaolen:
Relates to: https://github.com/xibosignageltd/xibo-private/issues/1164

Changes:
- added a modal to copy existing media with an option to change media name
- refactored code for modal states

### DIFF
--- a/frontend/src/services/mediaApi.ts
+++ b/frontend/src/services/mediaApi.ts
@@ -254,6 +254,98 @@ export async function downloadMedia(mediaId: number | string, fileName: string):
   window.URL.revokeObjectURL(url);
 }
 
+export interface CloneMediaRequest {
+  mediaId: number | string;
+  name: string;
+  fileName: string;
+  duration: number;
+  folderId?: number | string;
+  tags?: string[];
+  orientation?: 'portrait' | 'landscape';
+  expires?: string;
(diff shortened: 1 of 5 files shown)


frontend/src/services/mediaApi.ts
```diff
@@ -254,6 +254,98 @@ export async function downloadMedia(mediaId: number | string, fileName: string):
...
+  orientation?: 'portrait' | 'landscape';
+  expires?: string;
+  mediaNoExpiryDate?: number;
+  overrideName?: string;
+  signal?: AbortSignal;
+}
+
+export type CloneMediaResponse = Media;
+
+function incrementCopySuffix(value: string): string {
```
maurofmferrao: Should we move these string methods (incrementCopySuffix, incrementFileName and incrementDisplayName) to a separate utility file like in /src/utils/stringUtils.ts so we can re-use them in more places? ( I also see that it could be helpful to avoid repeating the same code that we have in the CopyMediaModal.tsx )

maurofmferrao (changes requested):
Just some changes to the structure of the code, everything else looks good! 👍🏻

maurofmferrao (approved):
Looks good to me! ✔️

PR #3278, opened 2026-02-13
Add date picker on Last Modified filter

chaolen:
First phase of the Last modified functionality:
- added custom range and date picker under Last modified filter
- edited query parameters to filter requests (will be fixed on BE)


maurofmferrao (approved):
Looks good to me!

PR #3282, opened 2026-02-16
Fix sqlerror with empty CampaignID

movisda:
This PR fixes issue [3819](https://github.com/xibosignage/xibo/issues/3819).


movisda:
It seems, that the core developers are working on a more complete fix, this PR is obsolete.

PR #3286, opened 2026-02-17
Layouts: Fixed show filter and sql error on campaignId

mgbaybay:
## Changes

The fix is divided into two parts: 
1. Handle empty campaignId lists properly to address the SQL query issue for the show filter and missing campaignId ([#3819](https://github.com/xibosignage/xibo/issues/3819), [#3814](https://github.com/xibosignage/xibo/issues/3814))
2. When running XTR Maintenance, the `tidyUnusedFullScreenLayout` process removes fullscreen layouts marked as unused, including some that were scheduled in the past ([#3818](https://github.com/xibosignage/xibo/issues/3818)). However, it should only remove fullscreen layouts that belong to deleted fullscreen events.

Relates to https://github.com/xibosignage/xibo/issues/3819, https://github.com/xibosignage/xibo/issues/3814, https://github.com/xibosignage/xibo/issues/3818

### DIFF
--- a/lib/Factory/LayoutFactory.php
+++ b/lib/Factory/LayoutFactory.php
@@ -2575,21 +2575,24 @@ class LayoutFactory extends BaseFactory
                 $campaignIds[] = $row['CampaignID'];
             }
 
-            if (!empty($campaignIds)) {
-                if ($parsedFilter->getInt('filterLayoutStatusId') == 2) {
-                    // Only show used layouts
-                    $body .= ' AND ('
-                        . '      campaign.CampaignID IN ( ' . implode(',', array_filter($campaignIds)) . ' ) 
-                             OR layout.layoutID IN (SELECT DISTINCT defaultlayoutid FROM display) 
-                             OR layout.layoutID IN (SELECT DISTINCT layoutId FROM lklayoutdisplaygroup)'
-                        . ' ) ';
-                } else {
-                    // Only show unused layouts
-                    $body .= ' AND campaign.CampaignID NOT IN ( ' . implode(',', array_filter($campaignIds)) . ' )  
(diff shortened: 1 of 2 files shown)


lib/Factory/LayoutFactory.php
```diff
@@ -2559,34 +2559,42 @@ public function query($sortOrder = null, $filterBy = [])
...
-                    . '      campaign.CampaignID IN ( ' . implode(',', array_filter($campaignIds)) . ' ) 
+            // Get the current or future scheduled layouts
+            $now = Carbon::now()->format('U');
+            $sql = 'SELECT DISTINCT schedule.CampaignID FROM schedule';
+            $campaignIds = [];
+
+            // Note: For the maintenance XTR, we will only delete unused/unscheduled fullscreen campaign layouts
+            // (i.e. when the user schedules a media and later on deletes the schedule)
+            if ($parsedFilter->getInt('isFullScreenCampaign', ['default' => -1]) != 1) {
+                $sql .= ' WHERE ( ( schedule.fromDt < ' . $now
```
dasgarner: I don't understand this bit - previously this filter related to the campaign type media/playlist, but now it relates to date/time?

dasgarner: As a side note, I don't think we can assume `isFullScreenCampaign` will only be used inside this other condition.

dasgarner: I think the call from maintenance should just exclude repeating events?

mgbaybay: Thank you for the review, @dasgarner! I have pushed new commits and also updated the description to give more context on the fix.

dasgarner (approved):
Looks good!

lykqa:
TEST: Library/playlists.cy.js END

PR #3294, opened 2026-02-23
Update Date Picker on Modal and Filter

chaolen:
relates to:
https://github.com/xibosignageltd/xibo-private/issues/1178
https://github.com/xibosignageltd/xibo-private/issues/1207

Changes:
- Updated Date Picker with two versions
- Expiry Date Select - use React `createPortal` for the dropdown component, applied position changes for dropdown and changed into single date selection.
- Select Filter (Last Modified) - applied "range version" of updated Date Picker

### DIFF
--- a/frontend/src/components/ui/SelectFilter.tsx
+++ b/frontend/src/components/ui/SelectFilter.tsx
@@ -138,7 +138,7 @@ export default function SelectFilter({
         className={twMerge(
           'absolute z-50 mt-1 min-w-full transition-all duration-200 ease-linear flex bg-xibo-white overflow-hidden rounded-lg border border-gray-200 shadow-lg',
           open
-            ? 'max-h-[400px] opacity-100 top-[68px] right-0'
+            ? 'max-h-[700px] opacity-100 top-[68px] right-0'
             : 'max-h-0 opacity-0 top-[60px] pointer-events-none',
         )}
       >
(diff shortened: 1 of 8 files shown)


frontend/src/components/ui/forms/ExpiryDateSelect.tsx
```diff
@@ -46,84 +46,153 @@ export default function ExpiryDateSelect({
 }: ExpiryDateSelectProps) {
   const { t } = useTranslation();
   const [openDatePicker, setOpenDatePicker] = useState(false);
+  const dropdownRef = useRef<HTMLDivElement | null>(null);
+  const containerRef = useRef<HTMLDivElement | null>(null);
+
+  const updatePosition = useCallback(() => {
```
maurofmferrao, line 52: With React 19 using the Compiler, we don't need to use `useCallback` anymore ( same with `useMemo` ) in most cases, like here 🙂

frontend/src/components/ui/SelectFilter.tsx
```diff
@@ -138,7 +138,7 @@ export default function SelectFilter({
         className={twMerge(
           'absolute z-50 mt-1 min-w-full transition-all duration-200 ease-linear flex bg-xibo-white overflow-hidden rounded-lg border border-gray-200 shadow-lg',
           open
-            ? 'max-h-[400px] opacity-100 top-[68px] right-0'
+            ? 'max-h-[700px] opacity-100 top-[68px] right-0'
```
maurofmferrao, line 141: Just a quick question, are the max values ( 700px here and 380px below ) coming from Win's designs? It's just that it feels wrong to me seeing those defined max values in px 🤔

maurofmferrao (changes requested):
Just some minor tweaks/comments 👍🏻

maurofmferrao (approved):
Looks good to me! 👍🏻

maurofmferrao (approved):
Looks good to me!

PR #3295, opened 2026-02-24
User: Add homePageUrl in MyDetails

mgbaybay:
## Changes

- Add `homePageUrl` in `user/me` (mydetails)

Relates to https://github.com/xibosignageltd/xibo-private/issues/1202

### DIFF
--- a/lib/Controller/User.php
+++ b/lib/Controller/User.php
@@ -235,7 +235,10 @@ class User extends Base
         ];
         $settings['accountId'] = defined('ACCOUNT_ID') ? constant('ACCOUNT_ID') : null;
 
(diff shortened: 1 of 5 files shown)


dasgarner (changes requested):
The homepageId should be passed into `urlFor` in order to get the URL of the route that has been declared for that homepageId. If the route changes, the URL returned will get automatically updated.

Hardcoding it in the definition is fine, but means we'd have to update it in two places instead of just in the route definition.

dasgarner:
Also your commit ID tracker points to a different issue.

dasgarner (approved):

PR #3313, opened 2026-03-04
Fix for dynamic playlist filters when creating playlist

maurofmferrao:
relates to xibosignageltd/xibo-private#1226


chaolen (approved):
Looks good!

PR #3316, opened 2026-03-04
Page permissions

maurofmferrao:
relates to xibosignageltd/xibo-private#1225

- Tab nav permissions also applied
- Unauth route added


chaolen (approved):
LGTM 💯

PR #3318, opened 2026-03-05
Notification: Fixed data sanitation and sandbox isolation

mgbaybay:
## Changes

- Add sandbox isolation and data sanitation for custom notifications

Relates to https://github.com/xibosignageltd/xibo-private/issues/1240


lib/routes-web.php
```diff
@@ -110,6 +110,9 @@
 $app->get('/drawer/notification/show/{id}', ['\Xibo\Controller\Notification','show'])->setName('notification.show');
 $app->get('/drawer/notification/interrupt/{id}', ['\Xibo\Controller\Notification','interrupt'])->setName('notification.interrupt');
```
dasgarner, line 111: We also have a "notification interrupt" route which shows notifications and would need protecting.

lib/routes-web.php
```diff
@@ -110,6 +110,9 @@
...
 $app->get('/notification/export/{id}', ['\Xibo\Controller\Notification','exportAttachment'])->setName('notification.exportattachment');
+$app->get('/notification/show/preview/{id}', ['\Xibo\Controller\Notification','showPreview'])
```
dasgarner: This being in the `web` entrypoint means it will load the whole web stack - do we want that?

views/notification-form-show.twig
```diff
@@ -37,9 +37,14 @@
         <div class="col-md-12">
             {% set readDt %}<span class="notification-date">{{ notification.readDt }}</span>{% endset %}
             <p>{% trans "Published" %} <span class="notification-date">{{ notification.releaseDt }}</span>, {% trans %}you read this {{ readDt }}.{% endtrans %}</p>
-            {{ notification.body|raw }}
+                <iframe
```
dasgarner, line 40: I can see why you've done it this way and I think it will work.

Do you think it might be overkill for showing a notification though? The HTML generated by the editor will be simple text, no file references, etc. Could we just inject that HTML into this sandboxed iframe, and forget about all of the token stuff, etc?

mgbaybay: Thank you, Dan! I have updated the PR to remove the token and used the `srcdoc` functionality instead. I've added the data sanitation prior to sending the data in backend, as well as adding `html_attr` attribute when rendering the notification body inside the iframe.

dasgarner (approved):

ifarzana:
TEST: all

ifarzana:
TEST: /Layout/Widget/layout_editor_dataset.cy.js END

PR #3332, opened 2026-03-11
fix: PdfProvider page count fallback when FPDI fails (wrong duration)

### LINKED ISSUE #3834: PdfProvider: FPDI fails on encrypted PDFs, calculatedDuration is wrong
## Bug description

`PdfProvider::fetchDuration()` uses mPdf/FPDI (`setSourceFile()`) to count PDF pages. FPDI cannot open encrypted or password-protected PDFs — it throws an exception that the outer catch handles by silently falling back to `pageCount=1`.

When `durationIsPerItem=1`, this produces `calculatedDuration = duration × 1` instead of `duration × actualPageCount`. The XLF then contains the raw per-page value as the total widget duration, causing pages to cycle at sub-second speeds.

## Previous reports

This was reported in #944 and #2629. Both were closed with a player-side workaround (`xiboIC.setWidgetDuration()` in the CMS HTML template). However:

1. The workaround only works for players that render the CMS HTML template — players with native PDF rendering never see the `xiboIC` call
2. The `durationIsPerItem` flag is not included in the XLF (`includeInXlf` defaults to `false` for HTML-rendered modules), so players cannot implement their own correction
3. The root cause in `PdfProvider` remains: `calculatedDuration` is wrong in the database and XLF

## Steps to reproduce

1. Upload an encrypted/password-protected PDF (e.g. owner-password, PDF 1.6+)
2. Add it to a layout with "Duration is per page" checked and a per-page duration (e.g. 5s)
3. Build/publish the layout
4. Check the XLF — `duration` will equal the per-page value (5) instead of the total (5 × pageCount)

## Proposed fix

Wrap the FPDI call in an inner try/catch and fall back to counting page objects in the raw PDF bytes. PDF encryption only covers content streams, not the object structure — `/Type /Page` entries and `/Count` values are always readable.

See xibosignage/xibo-cms PR for implementation.

## Workaround

Use **total duration** instead of **duration per page** for encrypted PDF widgets until this is fixed.

linuxnow:
## Summary

Fixes xibosignage/xibo#3834 — `PdfProvider::fetchDuration()` silently falls back to `pageCount=1` when FPDI fails on encrypted PDFs, producing incorrect `calculatedDuration` in the XLF.

Previously reported as xibosignage/xibo#944 and xibosignage/xibo#2629, closed with a player-side workaround (`xiboIC.setWidgetDuration()` in the HTML template). That workaround does not help players with native PDF rendering, and the root cause in PdfProvider remains.

### `lib/Widget/PdfProvider.php`

Wrap FPDI call in inner try/catch, add `countPdfPages()` fallback with 3 strategies:

1. Regex `/Type /Page` leaf nodes in raw bytes — works for PDFs with unencrypted object structure
2. `/Count N` from page tree root — works for PDFs with unencrypted cross-reference tables
3. **`qpdf --show-npages`** — handles all encryption types including fully encrypted object streams where methods 1-2 return 0

Note: PDFs with owner-password encryption and encrypted object streams (e.g. PDF 1.6+) encrypt the entire object structure, not just content streams. In this case raw byte parsing cannot find `/Type /Page` markers. `qpdf` is the only reliable method for these files.

### `Dockerfile`

Add `qpdf` package to the container image for method 3 fallback.

## Workaround

Use **total duration** instead of **duration per page** for encrypted PDF widgets until this is fixed.

## Test plan

- [ ] Upload unencrypted PDF with `durationIsPerItem=1` — FPDI counts pages (existing behavior unchanged)
- [ ] Upload encrypted PDF (owner-password, encrypted object streams) with `durationIsPerItem=1` — fallback counts pages via qpdf
- [ ] Verify `calculatedDuration = duration × pageCount` in both cases
- [ ] Verify XLF contains correct total duration after layout rebuild

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,6 +97,7 @@ RUN LC_ALL=C.UTF-8 DEBIAN_FRONTEND=noninteractive apt update && apt upgrade -y &
     openssl \
     cron \
     default-mysql-client \
+    qpdf \
     && dpkg-reconfigure --frontend noninteractive tzdata \
     && rm -rf /var/lib/apt/lists/*
 
--- a/lib/Widget/PdfProvider.php
+++ b/lib/Widget/PdfProvider.php
@@ -55,12 +55,23 @@ class PdfProvider implements WidgetProviderInterface
                     $this->getLog()->debug('fetchDuration: loading PDF file to get the number of pages, file: '
                         . $sourceFile);
 
-                    $mPdf = new Mpdf([
-                        'tempDir' => $widget->getLibraryTempPath(),
-                    ]);
-                    $pageCount = $mPdf->setSourceFile($sourceFile);
+                    // Try mPdf/FPDI first (works for unencrypted PDFs)
+                    try {
+                        $mPdf = new Mpdf([
+                            'tempDir' => $widget->getLibraryTempPath(),
+                        ]);
+                        $pageCount = $mPdf->setSourceFile($sourceFile);
+                    } catch (\Exception $fpdiException) {
+                        // FPDI fails on encrypted/password-protected PDFs.
+                        // Fall back to counting page objects in the raw PDF.
+                        $this->getLog()->info('fetchDuration: FPDI failed (' . $fpdiException->getMessage()
+                            . '), trying raw page count for: ' . $sourceFile);
+                        $pageCount = $this->countPdfPages($sourceFile);
+                    }
 
-                    $widget->setOptionValue('pageCount', 'attrib', $pageCount);
+                    if ($pageCount >= 1) {
+                        $widget->setOptionValue('pageCount', 'attrib', $pageCount);
+                    }
                 } catch (\Exception $e) {
                     $this->getLog()->error('fetchDuration: unable to get PDF page count, e: ' . $e->getMessage());
                 }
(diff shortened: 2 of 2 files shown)


linuxnow:
Rebased on current develop. Cleanly applies — single commit, single file. See issue xibosignage/xibo#3834 for context on encrypted PDF handling.

dasgarner:
Thanks for this PR - `qpdf ` doesn't exist across any of our infrastructure as far as I am aware, so this will need reviewing and approvals from the ops team.

@alexharrington

linuxnow:

linuxnow:
> Thanks for this PR - `qpdf ` doesn't exist across any of our infrastructure as far as I am aware, so this will need reviewing and approvals from the ops team.
> 
> @alexharrington

Thank you for your answer Dan, I will keep on testing it and keep the PR mergeable. I will be glad to adopt any other solution you might find.

dasgarner:
I've discussed this with the ops team and to use a common phrase - we do not think the juice is worth the squeeze. Our preference would be to say that encrypted PDFs are not supported.

The approach you've taken is the only way I think its possible to do it, however it means we have:
 - a new dependency to build into the stack
 - a new security hole - we're relying on `qpdf` not exposing any security vulnerabilities
 - a new resource usage avenue outside of PHP (exec shell) - this effectively means someone can upload a 2GB PDF and load it all into memory

linuxnow:
Thanks for the review. I understand the concern about qpdf, but I think the core issue got lost in the PR title — let me clarify.

The real problem is that **FPDI silently returns pageCount=1 for many valid, unencrypted PDFs** where it fails to parse the structure. This produces wrong `calculatedDuration` in the XLF (e.g. a 32-page PDF gets 10s instead of 320s). This has been reported multiple times (xibosignage/xibo#944, xibosignage/xibo#2629, xibosignage/xibo#3834).

The fix has 3 fallback strategies:

1. **Regex `/Type /Page`** — pure PHP, no dependency, works for most PDFs where FPDI fails
2. **Regex `/Count N`** — pure PHP, no dependency, alternative parsing
3. **`qpdf --show-npages`** — only for fully encrypted object streams (rare edge case)

Methods 1 and 2 are zero-dependency, zero-risk pure PHP. They fix the vast majority of cases. If the qpdf fallback (method 3) is a concern, I'm happy to remove it and only keep the PHP regex methods. That addresses the dependency, security and memory concerns while still fixing the core bug.

Would a version with only methods 1-2 (no qpdf, no new dependency) be acceptable?

dasgarner:
Thank you for clarifying.

> Methods 1 and 2 are zero-dependency, zero-risk pure PHP.

`$content = @file_get_contents($filePath);` loads the entire file into memory, which is significantly less efficient than mPDF which only parses the cross-reference structure.

Is there a way to do your fallback methods without parsing the entire file? I that (and dropping qpdf) would be a good solutiuon.

It would be even better if we could do a quick check for Encryption somehow so that we don't open the file twice.

PR #3333, opened 2026-03-12
feat: add pwa clientType to isPwa() and isWebSocketXmrSupported()

### LINKED ISSUE #3830: Support WebSocket XMR for linux clientType
### Summary

Displays registering with `clientType: 'linux'` cannot receive real-time XMR messages because `isWebSocketXmrSupported()` in `Display.php` doesn't include `linux`. Since the CMS already accepts `linux` as a valid client type at registration, this is an inconsistency — the display registers successfully but XMR silently fails.

### Problem

When the CMS checks `isWebSocketXmrSupported()`, it only returns `true` for `chromeOS`, `windows >= 407`, and `android >= 408`. Displays with `clientType: 'linux'` are excluded, so the CMS sends them RSA-encrypted ZeroMQ messages instead of plain WebSocket XMR.

Any browser-based or web-technology player running on Linux cannot decrypt RSA-encrypted ZeroMQ — it needs WebSocket XMR, just like `chromeOS` players do.

### Proposed Change

In `lib/Entity/Display.php`:

```php
public function isWebSocketXmrSupported(): bool
{
    return $this->clientType === 'linux'
        || $this->clientType === 'chromeOS'
        || ($this->clientType === 'windows' && $this->clientCode >= 407)
        || ($this->clientType === 'android' && $this->clientCode >= 408);
}
```

One-line addition, no impact on existing client types.

### Use Cases

- **Raspberry Pi players** — the community forum has many users running custom players on Pi hardware (Chromium kiosk, Python-based, etc.) that register as `linux`. They currently cannot receive real-time schedule updates.
- **Docker-based test players** — anyone testing the CMS with a headless browser in Docker registers as `linux` and can't verify XMR behavior.
- **Linux thin clients** — enterprise deployments using thin client hardware (IGEL, ThinLinX, etc.) running browser-based kiosk players.
- **Third-party Linux players** — any community or commercial player built with web technologies (Electron, CEF, WebView) on Linux.

linuxnow:
## Summary

Add a dedicated `pwa` clientType for web-based players, as suggested in the review.

This keeps clean separation:
- PWA players register as `pwa` — gets PWA URL handling and WebSocket XMR
- Native Linux players stay as `linux` — completely unaffected
- chromeOS unchanged

## Changes

- `isPwa()`: add `|| $this->clientType === 'pwa'`
- `isWebSocketXmrSupported()`: add `|| $this->clientType === 'pwa'`

## Why

Web-based players need `isPwa()` to return true for correct URL handling in `decorateForPlayer()`. Using a dedicated `pwa` clientType avoids any impact on the official native Linux player.

Fixes xibosignage/xibo#3830

### DIFF
--- a/lib/Entity/Display.php
+++ b/lib/Entity/Display.php
@@ -695,7 +695,8 @@ class Display implements \JsonSerializable
      */
     public function isPwa(): bool
     {
-        return $this->clientType === 'chromeOS';
+        return $this->clientType === 'chromeOS'
+            || $this->clientType === 'pwa';
     }
 
     /**
@@ -704,6 +705,7 @@ class Display implements \JsonSerializable
     public function isWebSocketXmrSupported(): bool
     {
         return $this->clientType === 'chromeOS'
+            || $this->clientType === 'pwa'
             || ($this->clientType === 'linux' && $this->clientCode >= 400)
             || ($this->clientType === 'windows' && $this->clientCode >= 407)
             || ($this->clientType === 'android' && $this->clientCode >= 408);
--- a/tests/Unit/DisplayEntityTest.php
+++ b/tests/Unit/DisplayEntityTest.php
@@ -1,0 +1,106 @@
+<?php
+/*
+ * Copyright (C) 2026 Xibo Players
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ */
+
+namespace Xibo\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Xibo\Entity\Display;
+
+/**
+ * Unit tests for Display::isPwa() and Display::isWebSocketXmrSupported()
+ */
+class DisplayEntityTest extends TestCase
+{
+    private function makeDisplay(string $clientType, int $clientCode = 0): Display
+    {
+        // Use reflection to create Display without constructor (avoids DI dependencies)
+        $ref = new \ReflectionClass(Display::class);
+        $display = $ref->newInstanceWithoutConstructor();
+        $display->clientType = $clientType;
+        $display->clientCode = $clientCode;
+        return $display;
+    }
+
+    // ── isPwa() ──────────────────────────────────────────────────
+
+    public function testIsPwaReturnsTrueForChromeOS(): void
+    {
+        $this->assertTrue($this->makeDisplay('chromeOS')->isPwa());
+    }
+
+    public function testIsPwaReturnsTrueForPwa(): void
+    {
+        $this->assertTrue($this->makeDisplay('pwa')->isPwa());
+    }
+
+    public function testIsPwaReturnsFalseForLinux(): void
+    {
+        $this->assertFalse($this->makeDisplay('linux', 400)->isPwa());
+    }
+
+    public function testIsPwaReturnsFalseForWindows(): void
+    {
+        $this->assertFalse($this->makeDisplay('windows', 407)->isPwa());
+    }
+
+    public function testIsPwaReturnsFalseForAndroid(): void
+    {
+        $this->assertFalse($this->makeDisplay('android', 408)->isPwa());
+    }
+
+    // ── isWebSocketXmrSupported() ────────────────────────────────
+
+    public function testXmrSupportedForChromeOS(): void
+    {
+        $this->assertTrue($this->makeDisplay('chromeOS')->isWebSocketXmrSupported());
+    }
+
+    public function testXmrSupportedForPwa(): void
+    {
+        $this->assertTrue($this->makeDisplay('pwa')->isWebSocketXmrSupported());
+    }
+
+    public function testXmrSupportedForLinuxAbove400(): void
+    {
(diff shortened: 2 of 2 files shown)


linuxnow:
Rebased on current develop. This PR now contains only the `isPwa()` change — the `isWebSocketXmrSupported()` part was already merged in #3367. Single commit, single file, minimal diff.

lib/Entity/Display.php
```diff
@@ -695,7 +695,8 @@ public function getLanguages()
      */
     public function isPwa(): bool
     {
-        return $this->clientType === 'chromeOS';
+        return $this->clientType === 'chromeOS'
+            || ($this->clientType === 'linux' && $this->clientCode >= 400);
```
dasgarner: Xibo's official Linux Player is a native player, not a PWA, and this PR would break that.

We likely need to address this with a new client type called `pwa` which has its own Display settings profile, etc.

How are you getting your PWA served from the right domain/context to go alongside this change?

dasgarner (changes requested):

lib/Entity/Display.php
linuxnow: Thanks for the review. You're right — reusing `clientType=linux` is not ideal since it would affect the official Linux player.

We're happy to switch to a dedicated `clientType=pwa` instead. This would be cleaner:

- `isPwa()` returns true for `clientType === 'pwa'` (no version gate needed)
- `isWebSocketXmrSupported()` adds `clientType === 'pwa'` (same as chromeOS)
- The official Linux player stays completely unaffected

If you agree with this approach, we can update the PR to use `clientType=pwa`.

Would a separate Display settings profile for `pwa` be needed, or would it inherit from `chromeOS` since the behavior is identical?

linuxnow:

lib/Entity/Display.php
dasgarner: This approach is much clearer.

> Would a separate Display settings profile for pwa be needed, or would it inherit from chromeOS since the behaviour is identical?

A separate one would be excellent for completeness.

linuxnow:
Thanks Dan. Here's what we'd add for the separate Display settings profile. Does this approach look right before we push the final version?

**1. `Display.php`** — add `pwa` to both methods:
```php
public function isPwa(): bool
{
    return $this->clientType === 'chromeOS'
        || $this->clientType === 'pwa'
        || ($this->clientType === 'linux' && $this->clientCode >= 400);
}

public function isWebSocketXmrSupported(): bool
{
    return $this->clientType === 'chromeOS'
        || $this->clientType === 'pwa'
        || ($this->clientType === 'linux' && $this->clientCode >= 400)
        || ($this->clientType === 'windows' && $this->clientCode >= 407)
        || ($this->clientType === 'android' && $this->clientCode >= 408);
}
```

**2. `DisplayProfileFactory.php`** — PWA profile config (mirrors chromeOS since it's the same web-based player, plus `forceHttps`):
```php
'pwa' => [
    ['name' => 'collectInterval', 'default' => 300],
    ['name' => 'xmrNetworkAddress', 'default' => ''],
    ['name' => 'xmrWebSocketAddress', 'default' => ''],
    ['name' => 'statsEnabled', 'default' => (int)$this->config->getSetting('DISPLAY_PROFILE_STATS_DEFAULT', 0), 'type' => 'checkbox'],
    ['name' => 'aggregationLevel', 'default' => $this->config->getSetting('DISPLAY_PROFILE_AGGREGATION_LEVEL_DEFAULT'), 'type' => 'string'],
    ['name' => 'playerVersionId', 'default' => null],
    ['name' => 'dayPartId', 'default' => null],
    ['name' => 'logLevel', 'default' => 'error'],
    ['name' => 'elevateLogsUntil', 'default' => 0, 'type' => 'int'],
    ['name' => 'sendCurrentLayoutAsStatusUpdate', 'default' => 0, 'type' => 'checkbox'],
    ['name' => 'screenShotRequestInterval', 'default' => 0, 'type' => 'int'],
    ['name' => 'screenShotSize', 'default' => (int)$this->config->getSetting('DISPLAY_PROFILE_SCREENSHOT_SIZE_DEFAULT', 200)],
    ['name' => 'forceHttps', 'default' => 1, 'type' => 'checkbox'],
],
```

**3. Migration** — insert default PWA profile (same pattern as chromeOS migration):
```php
class AddDefaultPwaDisplayProfileMigration extends AbstractMigration
{
    public function change(): void
    {
        if (!$this->fetchRow('SELECT * FROM displayprofile WHERE type = \'pwa\' AND isDefault = 1')) {
            $user = $this->fetchRow('SELECT userId FROM `user` WHERE userTypeId = 1');
            $this->table('displayprofile')->insert([
                'name' => 'PWA',
                'type' => 'pwa',
                'config' => '[]',
                'userId' => $user['userId'],
                'isDefault' => 1
            ])->save();
        }
    }
}
```

**4. Unit test** for the new clientType.

We haven't included `DisplayProfileConfigFields.php` or a Twig template — happy to add those if you want a custom settings UI, or the PWA profile can share the chromeOS form initially.

Let us know if anything needs adjusting.

dasgarner:
Thanks - having a separate client type is definitely the right approach. 

Please give me some time to consider this before you put more effort into it. There are a couple of caveats and things that need to be considered:

 - we can only introduce migrations in major/minor releases, the next of which is 5.0 (`neujmin` branch)
 - having a web player display settings profile will confuse users - Xibo does not have an official web player
 - we used to have a way to inject custom client types (a hook called `registerCustomDisplayProfile`) which might be a better way, but will need some work

PR #3340, opened 2026-03-13
fix: return empty response from GetWeather when display has no coordinates

### LINKED ISSUE #3835: XMDS GetWeather throws SoapFault (HTTP 500) when display has no coordinates
## Summary

The XMDS `GetWeather` method in `Soap7.php` throws a `SoapFault` when a display has no latitude/longitude configured. PHP's SoapServer translates all SoapFaults to HTTP 500, which makes clients interpret this as a server error rather than a missing-data condition.

## Current behavior

When `$display->latitude` or `$display->longitude` is empty/null:

```php
// lib/Xmds/Soap7.php:317-320
throw new \SoapFault(
    'Receiver',
    'Display coordinates is not configured'
);
```

This returns HTTP 500 to the client. For players that call GetWeather on every collection cycle, this produces continuous 500 errors in logs — noisy and misleading.

## Expected behavior

Return an empty or default weather response (e.g., `""` or `"{}"`), allowing the player to handle the missing data gracefully without error-level logging. The absence of coordinates is a normal configuration state (e.g., new displays, displays without GPS), not a server error.

## Suggested fix

```php
if ($latitude && $longitude) {
    $event = new XmdsWeatherRequestEvent($latitude, $longitude);
    $this->getDispatcher()->dispatch($event, XmdsWeatherRequestEvent::$NAME);
    return $event->getWeatherData();
} else {
    // No coordinates — return empty response instead of faulting
    return '';
}
```

## Environment

- Xibo CMS 4.x (current `develop` branch)
- `lib/Xmds/Soap7.php:283-325`
- Affects any XMDS v5+ client calling GetWeather

linuxnow:
## Summary

- Return empty string from `GetWeather` when display has no latitude/longitude instead of throwing a `SoapFault`
- Missing coordinates is a normal configuration state (new displays, no GPS), not a server error

## Problem

`Soap7::GetWeather()` throws `SoapFault('Receiver', 'Display coordinates is not configured')` when a display has no coordinates. PHP's `SoapServer` maps all `SoapFault` exceptions to HTTP 500. Players call `GetWeather` on every collection cycle, producing continuous 500 errors in logs.

## Fix

Return an empty string instead of faulting. Players already handle empty weather responses gracefully.

## Test plan

- [ ] Display with coordinates: weather data returned as before
- [ ] Display without coordinates: empty string returned, HTTP 200
- [ ] Player logs: no more 500 errors for GetWeather on displays without GPS

Fixes xibosignage/xibo#3835

### DIFF
--- a/lib/Xmds/Soap7.php
+++ b/lib/Xmds/Soap7.php
@@ -313,14 +313,11 @@ class Soap7 extends Soap6
             // Dispatch an event to initialize weather data.
             $event = new XmdsWeatherRequestEvent($latitude, $longitude);
             $this->getDispatcher()->dispatch($event, XmdsWeatherRequestEvent::$NAME);
-        } else {
-            throw new \SoapFault(
-                'Receiver',
-                'Display coordinates is not configured'
-            );
+            return $event->getWeatherData();
         }
 
-        // return weather data
-        return $event->getWeatherData();
+        // No coordinates — return error in response instead of SoapFault.
+        // Players can log this; server avoids HTTP 500 on every collection.
+        return '<weather><error>Display has no coordinates configured</error></weather>';
     }
 }


linuxnow:
Any chance this could be reviewed? The fix is minimal — returns an empty XML response from GetWeather instead of throwing a SoapFault (HTTP 500) when the display has no coordinates set. Currently any new display without coordinates configured causes a server error on every XMDS check.

dasgarner:
I agree with you that missing coordinates is not a server error and in a non-soap world would be in the 400 range. However I am not sure if returning an empty response, thereby hiding the error, is the right solution.

In a troubleshooting scenario where a user has a weather based schedule that isn't working, then it is helpful to see logging showing that the Display is making the request, but that it doesn't have coordinates. Accepting this PR would make that less clear.

> Players call GetWeather on every collection cycle, producing continuous 500 errors in logs.

This is especially true when you consider that Displays should only make requests if there is a weather based schedule currently active. I.e. the user really is expecting there to be coordinates, but there isn't.

Does that help explain why having the error is important?

linuxnow:
That makes sense — the error is useful as a diagnostic signal when coordinates are expected but missing.

Would a compromise work? Instead of a SoapFault (500), return the weather response with an error field that the player can log, but without crashing the SOAP call:

```xml
<weather>
  <error>Display has no coordinates configured</error>
</weather>
```

This way:
- The CMS returns a proper response (not HTTP 500)
- The player sees the error and can log it
- Server logs stay clean (no stack traces from expected conditions)
- Users can still diagnose "no coordinates" from the player logs

Or if you prefer keeping the current behavior, I can close this PR — we handle it in our player by catching the SOAP fault gracefully.

dasgarner:
I'd prefer to keep the current behaviour please - while your suggestion is good, it would have to be based on player version for backwards compatibility and it doesn't really get us much benefit.

I think its likely we will change the way XMDS works in future, replacing SOAP with a JSON API. We do this we can have a better HTTP code issued.

linuxnow:
Understood, thanks for considering it. We'll keep our local patch to avoid the unnecessary error on our players and look forward to the JSON API replacement.

PR #3345, opened 2026-03-18
Fix wrong label sizes for input fields

maurofmferrao:
relates to xibosignageltd/xibo-private#1284

 - Allow breaks on helptexts


chaolen (approved):
LGTM 💯

PR #3365, opened 2026-03-25
Media: backend improvement and cleanup

mgbaybay:
## Changes

**Added:**
- `LibraryDescription` helper file
- JSON response for `grid` 

**Improved:**
- Refactored Library controller

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1219


views/authed-topbar.twig
```diff
@@ -91,7 +91,7 @@
             {% endif %}
 
             {% if currentUser.featureEnabled("library.view") %}
-                <a class="{{ groupElementClass }}" href="{{ url_for("library.view") }}">{% trans "Media" %}</a>
+                <a class="{{ groupElementClass }}" href="/prototype/library/media">{% trans "Media" %}</a>
```
dasgarner: I know this is only temporary, but it won't work if the CMS runs inside a sub-folder (need to use the baseURL)

dasgarner (changes requested):

dasgarner (changes requested):
Sorry Mae, what I meant in my prior review is that you need to put `{{ theme.rootUri() }}` in front of the hardcoded URL. That will prepend the appropriate sub-folder if one is present.

No need for redirecting via the API.

dasgarner (approved):

PR #3366, opened 2026-03-25
Feat: add new Templates page

chaolen:
note:
Reused the `copyLayout` and `deleteLayout` wrapper from `layoutsApi.ts` since the endpoint and URL-encoded payload structure are identical with templates.

-----

### Backend Changes

**Added:**
- `keyword` search and `sortBy`, `sortDir` sorting

**Improved:**
- Refactored Templates controller

**Removed:**  

Relates to:
FE: https://github.com/xibosignageltd/xibo-private/issues/1254
BE: https://github.com/xibosignageltd/xibo-private/issues/1259


frontend/src/pages/Design/Templates/components/DeleteTemplateModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+    >
+      <div className="flex flex-col p-5 gap-3">
+        <div>
+          <div className="flex justify-center mb-4">
+            <div className="bg-red-100 w-15.5 h-15.5 text-red-800 border-red-50 border-[7px] rounded-full p-3">
+              <Trash2Icon size={26} />
+            </div>
+          </div>
+          <h2 className="text-center text-lg font-semibold mb-2 text-red-800">
+            {itemCount === 1 ? t('Delete Layout?') : t('Delete Layouts?')}
```
maurofmferrao: Shouldn't this be referring to Template instead of Layout?

frontend/src/pages/Design/Templates/components/DeleteTemplateModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+        </div>
+        <p className="text-center text-gray-500">
+          {itemCount === 1 ? (
+            <>
+              {t('Are you sure you want to delete ')}"<strong>{templateName}</strong>?"
+            </>
+          ) : (
+            <>
+              {t('Are you sure you want to delete ')}
+              <strong>{itemCount}</strong> {t('layouts')}?
```
maurofmferrao: Layout reference instead of Template

frontend/src/pages/Design/Templates/hooks/useTemplateActions.ts
```diff
@@ -0,0 +1,162 @@
...
+    try {
+      setIsCloning(true);
+
+      await copyLayout(selectedLayout.layoutId, {
+        name: newName,
+        description,
+        copyMediaFiles: copyMediaFiles ? 1 : 0,
+      });
+
+      notify.success(t('Layout copied successfully'));
```
maurofmferrao: Layout reference instead of Template

frontend/src/pages/Design/Templates/hooks/useTemplateActions.ts
```diff
@@ -0,0 +1,162 @@
...
+      handleRefresh();
+      closeModal();
+      console.log('copy layout', selectedLayout, newName, description, copyMediaFiles);
+    } catch (error) {
+      console.error('Copy layout failed', error);
+      notify.error(t('Failed to copy layout'));
```
maurofmferrao: Layout reference instead of Template

frontend/src/pages/Design/Templates/Templates.tsx
```diff
@@ -0,0 +1,470 @@
...
+          <div className="bg-red-50 border border-red-200 text-red-800 p-4" role="alert">
+            {error}
+          </div>
+        )}
+
+        <div className="min-h-0 flex flex-col">
+          {!isHydrated ? (
+            <div className="flex-1 flex items-center justify-center bg-gray-50 animate-pulse rounded-lg border border-gray-200">
+              <span className="text-gray-400 font-medium">
+                {t('Loading your layout preferences...')}
```
maurofmferrao: Layout reference instead of Template

maurofmferrao (changes requested):
Apart from some UI text referring to Layout instead of Template, all looks good! ( not sure if I caught all the cases for the Layout references )

maurofmferrao (approved):
Looks good to me!

chaolen:
> Apart from some UI text referring to Layout instead of Template, all looks good! ( not sure if I caught all the cases for the Layout references )

Good catch! I’ve updated the UI text to consistently use Template instead of Layout where visible to the user. Also, just to avoid confusion too, the payload uses `layout` for the name and `layoutId` for the ID per the data schema of template.

maurofmferrao:
@chaolen and @mgbaybay , I just tested this PR to see if the page is working, and the search doesn't seem to be working, for the Search bar and also the Filters. Can you please check that before we merge?

chaolen:

mgbaybay:
> @chaolen and @mgbaybay , I just tested this PR to see if the page is working, and the search doesn't seem to be working, for the Search bar and also the Filters. Can you please check that before we merge?

Thank you for checking, @maurofmferrao! There were some adjustments on the filter and the search, could you please verify if it works on your end? You might need to run the migration beforehand for the indexing to take effect.

maurofmferrao:
The search bar is now working for me, but I see 2 things that are still not working:

- The Tag filter input is a text input, in v4 it was a Tag input
image


- Since it's a text input, you can add something like "one, two," and it will return all results since the 3rd tag on the array is empty string and matches all
image


Also, can we have this branch already with the fixes that we applied to all the other pages in the https://github.com/xibosignage/xibo-cms/pull/3376 PR?

Those being:
- All modals have a `isOpen` optional property, set to `true` by default
- Wrap all modals in a condition instead of relying on the `isOpen` so they can unmount when closed
```
{isModalOpen('share') && ( ... )}
```
- Use `isAxiosError` in each modal to grab and process the back-end error
- For bulk edit, replace the following code:

```
if (rowSelection[id]) {
    if (!next[id]) {
          next[id] = item;
          hasChanges = true;
     }
}
```
        
with
 
```      
if (rowSelection[id] && next[id] !== item) {
    next[id] = item;
     hasChanges = true;
}     
```

maurofmferrao (approved):
Looks good to me!

PR #3371, opened 2026-03-26
Timer Picker new input and input style adjustments

maurofmferrao:
- Time Picker Input created
- Adjust Input heights
- Enable disabling options on Select Dropdown


chaolen (approved):
Looks good!

PR #3383, opened 2026-03-31
Implemented searchById routes where applicable

PeterMis:
Added searchById route to Library, Resolutions, Template and DayPart Controllers.

Relates to : 
- https://github.com/xibosignageltd/xibo-private/issues/1308

Misc : 
- Removed displayPage function in Template controller ( twig and route were already removed )
- Fixed urls in top/side navigation twig files
- Minor code cleanup

### DIFF
--- a/lib/Controller/DayPart.php
+++ b/lib/Controller/DayPart.php
@@ -22,6 +22,7 @@
 namespace Xibo\Controller;
 
 use OpenApi\Attributes as OA;
+use Psr\Http\Message\ResponseInterface;
 use Slim\Http\Response as Response;
 use Slim\Http\ServerRequest as Request;
 use Xibo\Factory\DayPartFactory;
@@ -29,6 +30,7 @@ use Xibo\Factory\ScheduleFactory;
 use Xibo\Service\DisplayNotifyServiceInterface;
 use Xibo\Support\Exception\AccessDeniedException;
 use Xibo\Support\Exception\InvalidArgumentException;
+use Xibo\Support\Exception\NotFoundException;
 
 /**
  * Class DayPart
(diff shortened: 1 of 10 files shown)


lib/Controller/DayPart.php
```diff
@@ -174,6 +176,60 @@ public function grid(Request $request, Response $response)
...
+            : [];
+
+        if (!in_array('exceptions', $embed)) {
+            $dayPart->excludeProperty('exceptions');
+        }
+        $dayPart->setUnmatchedProperty('userPermissions', $this->getUser()->getPermission($dayPart));
+
+        return $response
+            ->withStatus(200)
+            ->withHeader('X-Total-Count', $this->dayPartFactory->countLast())
```
dasgarner: Is `getById` still doing two queries to get the count, and if so, is that a great opportunity to optimise?

dasgarner (approved):
It looks good, although I did have a question.

lib/Controller/DayPart.php
PeterMis: Ah, that's a good point I've missed that, it does not run the second query, as we do not pass paging parameters, therefore that header would be 0 as it is - perhaps the question is do we actually need this header in this request? 🤔

dasgarner: I would say that we don't need it. If you remove it we should ensure the OA block matches.

dasgarner (approved):

lib/Controller/DayPart.php
PeterMis: That's removed now  - OA did not mention that header
That said I think some grid functions are missing OA about it, but that can be sorted out in a new issue/PR

PR #3386, opened 2026-04-01
Slider input and lazy loading support for select dropdown

maurofmferrao:
- Add prop filters for daypart
- Added player softare API


chaolen (approved):
Looks good!

PR #3394, opened 2026-04-02
Refactor: simplify data flow across main CMS pages

maurofmferrao:
- Remove redundant local copies of server data and replace chains ofsync watchers with direct data use and single event handlers

- Remove useMemo usage


chaolen (approved):
Looks good to me!

PR #3398, opened 2026-04-06
Datasets: initial backend improvement and cleanup

mgbaybay:
## Changes

**Added:**
- `keyword` search indexing
- `folderName` for Datasets page

**New routes:**

Datasets:
- `searchById` 

Dataset columns:
- `getDataSetColumnTypes` 
- `getDataTypes` 
- `searchById` 

Dataset RSS:
- `searchById`

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1256


lib/Entity/DataSet.php
```diff
@@ -1132,7 +1138,7 @@ private function add()
             'isRemote' => $this->isRemote,
             'isRealTime' => $this->isRealTime,
             'dataConnectorSource' => $this->dataConnectorSource,
-            'lastDataEdit' => 0,
+            'lastDataEdit' => Carbon::now()->format('U'),
```
PeterMis: Is there a reason why we want to set the date here on add?
Have we considered if the previous default ( 0 ) is in use anywhere? ( like say dataSetRss feed function )

PeterMis (commented):
I think it looks good, just one comment below.

There are a few things that I'd normally point out as well, however I believe those were addressed on develop and will be resolved when we merge that

Few FE issues with the filtering of dataSet column/data, but my understanding is that fixes are in progress, will re-review once that's updated

lib/Entity/DataSet.php
mgbaybay: Thank you for the review, @PeterMis! I've reverted this change and updated the query instead :D This was initially added to accommodate the `Last Modified` filter in the Datasets page. It uses `lastDataEdit` as the reference, so when we filter newly created datasets, they previously get excluded since their value is `0` (outside the date filter).

maurofmferrao (approved):
Looks good to me!

PeterMis (approved):
Looks good 👍

PR #3407, opened 2026-04-10
Campaigns Page

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1302

Description:
- New Campaign Page
- Edit button for Ad Campaign Type is temporarily hidden
- Already linked to Layouts Page -> Jump To Campaign
- Edit Modal is not yet final and will need approval from Win.

_Note:_
Ad Campaign and Preview Campaign is not yet implemented.

------
### Backend Changes:

**Added:**
- `keyword` search indexing 
- `userPermissions`, `folderName` in JSON response
- `searchByUserId` route

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1304


db/migrations/20260408171106_add_index_to_campaign_table_migration.php
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Add search index in Campaign Table
+ * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+ */
+class AddIndexToCampaignTableMigration extends AbstractMigration
+{
+    public function change(): void
+    {
+        $this->execute('ALTER TABLE campaign ADD FULLTEXT idx_campaign_search(campaign, `type`) WITH PARSER ngram');
```
PeterMis: This might be correct, but just for sanity check, do we think the type makes sense in full text index?
Wouldn't it be cleaner to have the type filter as it was in v4 and search to be for name only? ( well and id )

lib/Controller/Campaign.php
```diff
@@ -227,240 +261,93 @@ public function displayCampaignBuilder(Request $request, Response $response, $id
...
-                        ['name' => 'custom-handler', 'value' => 'XiboMultiSelectPermissionsFormOpen'],
-                        ['name' => 'custom-handler-url', 'value' => $this->urlFor($request,'user.permissions.multi.form', ['entity' => 'Campaign'])],
-                        ['name' => 'content-id-name', 'value' => 'campaignId']
-                    ]
-                ];
-            }
+            $this->decorateCampaignProperties($campaign, $embed);
         }
 
+        // TODO: Convert this to JSON response object
```
PeterMis: I presume this is in use on some other pages/forms, would it make sense to convert it for isJson/isApi ( + add the appropriate header to OA ) ?
Keep the old way for old web requests with todo to clean that up when possible

lib/Factory/CampaignFactory.php
```diff
@@ -264,15 +273,14 @@ public function query($sortOrder = null, $filterBy = [])
...
+        if ($sanitizedFilter->getInt('hasLayouts') !== null) {
             $body .= " AND (
                 SELECT COUNT(*)
                 FROM lkcampaignlayout
                 WHERE lkcampaignlayout.campaignId = `campaign`.campaignId
-                )";
+            )";
 
-            $body .= ($sanitizedFilter->getInt('hasLayouts', ['default' => 0]) == 1) ? " = 0 " : " > 0";
+            $body .= ($sanitizedFilter->getInt('hasLayouts') == 0) ? " = 0 " : " > 0 ";
```
PeterMis, line 283: Just a sanity check, I think this is correct, I presume the filter in FE was adjusted?
Previously it was rather odd ( sending 0 for no filter, 1 for no layouts and 2 for layouts...)
I'm assuming now we just send null / 0 / 1 so it's makes more sense, is that correct?

PeterMis (commented):
Looks good to me ( I've only looked at the backend changes )
Just a few small comments, likely most important one being on the migration

frontend/src/pages/Design/Campaigns/Campaigns.tsx
```diff
@@ -0,0 +1,469 @@
...
+ */
+
+import { useQueryClient } from '@tanstack/react-query';
+import type { RowSelectionState } from '@tanstack/react-table';
+import { Search, Filter, FilterX, Plus } from 'lucide-react';
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router-dom';
+
+import type { ModalType } from '../Templates/TemplatesConfig';
```
maurofmferrao: Shouldn't we be importing this from CampaignConfig?

frontend/src/pages/Design/Campaigns/Campaigns.tsx
```diff
@@ -0,0 +1,469 @@
...
+    return Object.keys(rowSelection)
+      .map((id) => selectionCache[id])
+      .filter((item): item is Campaign => !!item);
+  };
+
+  const bulkActions = getBulkActions({
+    t,
+    onDelete: () => {
+      const allItems = getAllSelectedItems();
+      setItemsToDelete(allItems);
```
maurofmferrao, line 278: We're missing `setDeleteError(null);` here ( check other Page components )

frontend/src/pages/Design/Campaigns/CampaignConfig.tsx
```diff
@@ -0,0 +1,424 @@
+import type { ColumnDef } from '@tanstack/react-table';
```
maurofmferrao, line 22: We're missing the copyright headers in some of the new files.

frontend/src/pages/Design/Campaigns/Campaigns.tsx
```diff
@@ -0,0 +1,469 @@
...
+
+      if (targetFolder) {
+        handleFolderChange({ id: targetFolder.id, text: targetFolder.text });
+      } else {
+        handleRefresh();
+      }
+    },
+  });
+
+  useEffect(() => {
```
maurofmferrao: Here and line 125: Other pages are using `useQuery` for this. So we need to replace `useState` in line 125 and this `useEffect` with something like:

```
const { data: folderPerms } = useQuery({
  queryKey: ['folderPermissions', selectedFolderId],
  queryFn: () => fetchContextButtons(selectedFolderId as number),
  enabled: selectedFolderId !== null,
  staleTime: 1000 * 60 * 5,
});
const canAddToFolder = folderPerms?.create || false;
```

frontend/src/pages/Design/Campaigns/Campaigns.tsx
```diff
@@ -0,0 +1,469 @@
...
+              onClick={() => setOpenFilter((prev) => !prev)}
+              removeTextOnMobile
+            >
+              {t('Filters')}
+            </Button>
+          </div>
+        </div>
+
+        <FilterInputs
+          onChange={(name, value) => {
```
maurofmferrao, line 366: The null-check and `as CampaignFilterInput` cast aren't needed. Other pages just do `setFilterInputs((prev) => ({ ...prev, [name]: value }))`

frontend/src/services/campaignApi.ts
```diff
@@ -62,3 +54,163 @@ export async function fetchCampaigns(
...
+
+  signal?: AbortSignal;
+}
+
+export interface FetchCampaignTableResponse {
+  rows: Campaign[];
+  totalCount: number;
+}
+
+export async function fetchCampaigns(
```
maurofmferrao, line 53: What's the difference between this fetchCampaigns and fetchCampaignsList? could we use the same for both cases?

frontend/src/pages/Design/Campaigns/hooks/useCampaignFilterOptions.ts
```diff
@@ -0,0 +1,14 @@
+import type { TFunction } from 'i18next';
+import { useEffect, useState } from 'react';
+
+import { getCampaignFilterKeys } from '../CampaignConfig';
+
+export function useCampaignFilterOptions(t: TFunction) {
```
maurofmferrao, line 26: We don't need `useEffect` and `useState` on this file, since the filter options are static, we can just write the method as:

```
export function useCampaignFilterOptions(t: TFunction) {
  return { filterOptions: getCampaignFilterKeys(t), isLoading: false };
}
```

maurofmferrao (changes requested):
Just some changes to the structure and some copyright comments missing, other than that, it's looking great! 👍🏻

lib/Factory/CampaignFactory.php
mgbaybay: Thanks, @PeterMis! Yes, you are correct - the frontend sends null/0/1, so the backend was adjusted to handle the values.

PeterMis (approved):
Backend changes look good to me 🚀 

That said we should wait with merging for after frontend changes are approved as well please

maurofmferrao (approved):
Looks good to me!

PR #3411, opened 2026-04-13
POP and Preview

dasgarner:
## POP
Attempt to look up widget details from widgethistory.
relates to xibosignage/xibo#3838

I am not sure if this covers all cases, because the widgethistory table doesn't look to be maintained in all cases. I think fixing that issue might be out of scope for a bug fix like this.

## Preview
Fixed layout download URL


PeterMis (dismissed):

dasgarner:
Whoops, I accidentally added a fix for https://github.com/xibosignage/xibo/issues/3845

PeterMis (approved):

ifarzana:
TEST: /Reporting END

PR #3414, opened 2026-04-14
Playlist : backend improvements and cleanup.

PeterMis:
- Relates to https://github.com/xibosignageltd/xibo-private/issues/1327


lib/Controller/Playlist.php
```diff
@@ -274,143 +262,79 @@ enum: [
...
-            }
+            $this->decoratePlaylistProperties($sanitizedParams, $playlist);
+        }
 
-            if ($sanitizedParams->getCheckbox('fullScreenScheduleCheck')) {
-                $fullScreenCampaignId = $this->hasFullScreenLayout($playlist);
-                $playlist->setUnmatchedProperty('hasFullScreenLayout', (!empty($fullScreenCampaignId)));
-                $playlist->setUnmatchedProperty('fullScreenCampaignId', $fullScreenCampaignId);
-            }
+        if ($this->isJson($request) || $this->isJson($request)) {
```
mgbaybay: I think the other condition is for API?

PeterMis: ah, you're quite right, good catch :)

mgbaybay (approved):

PR #3415, opened 2026-04-14
Refactor layout handling and add layout and template actions

ifarzana:
- Remove name from the Layout TypeScript type 
- the API never returns this field; 
- the display name is always layout.                                                                                                                                                   
- Fix a copy-paste bug in EnableStatsLayoutModal where the modal title read "Retire Layout" and the confirm button read "Retire" (copied from RetireLayoutModal). Now reads "Enable Stats   
  Collection" / "Save".   

- https://github.com/xibosignageltd/xibo-private/issues/1356
- https://github.com/xibosignageltd/xibo-private/issues/1357


chaolen (approved):
LGTM💯

Kent0425 (approved):

PR #3419, opened 2026-04-15
User: revert regression and use sanitiser default to correctly ignore missing libraryQuota via the API.

dasgarner:
fixes xibosignageltd/xibo-private#1282

For posterity this PR reverts the change made previously to make libraryQuota not required for the API, which fixes this issue. It then reimplements the fix for the API issue, so that the API does not change behaviour (not providing a value keeps the same value)

This makes this field work differently to others, so that if you clear its value and leave it empty, it will not change the existing value - makes sense for the API, weird for the form.

A better original fix would be to make the parameter required in the API and on the form. Doing so now would be a breaking change for the original issue OP, so cannot be done in a patch.

### DIFF
--- a/lib/Controller/User.php
+++ b/lib/Controller/User.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2025 Xibo Signage Ltd
+ * Copyright (C) 2026 Xibo Signage Ltd
  *
  * Xibo - Digital Signage - https://xibosignage.com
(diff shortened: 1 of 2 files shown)


PeterMis (approved):

ifarzana:
TEST: /Administration END

ifarzana:
TEST: /Administration/folder.cy.js END

PR #3439, opened 2026-04-23
Modules: backend improvement and cleanup

mgbaybay:
## Tasks
- [x] https://github.com/xibosignageltd/xibo-private/issues/1479
- [x] https://github.com/xibosignageltd/xibo-private/issues/1332

----
### Backend Changes

**Added:**
- `keyword` search
- `searchById` route implementation

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1479, https://github.com/xibosignageltd/xibo-private/issues/1332


lib/Controller/Module.php
```diff
@@ -59,92 +59,152 @@ public function __construct(
...
+        content: new OA\JsonContent(ref: '#/components/schemas/Module')
+    )]
+    /**
+     * @param Request $request
+     * @param Response $response
+     * @param int $id
+     * @return Response|ResponseInterface
+     * @throws NotFoundException
+     */
+    public function searchById(Request $request, Response $response, int $id): Response|ResponseInterface
```
PeterMis: should the id here be a string instead?

lib/Controller/Module.php
```diff
@@ -171,17 +231,18 @@ public function grid(Request $request, Response $response)
...
+     * @return ResponseInterface|Response
      * @throws AccessDeniedException
      * @throws GeneralException
      * @throws NotFoundException
-     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     * @throws ControllerNotImplemented
      */
     // phpcs:enable
-    public function getProperties(Request $request, Response $response, $id)
+    public function getProperties(Request $request, Response $response, $id): Response|ResponseInterface
```
PeterMis: $id here is likely expected to be a string?

lib/Controller/Module.php
```diff
@@ -275,40 +307,17 @@ public function settings(Request $request, Response $response, $id)
...
      * @param $id
-     * @return \Psr\Http\Message\ResponseInterface|Response
+     * @return ResponseInterface|Response
      * @throws GeneralException
      * @throws NotFoundException
-     * @throws \Xibo\Support\Exception\ControllerNotImplemented
+     * @throws ControllerNotImplemented
      */
-    public function clearCache(Request $request, Response $response, $id)
+    public function clearCache(Request $request, Response $response, $id): Response|ResponseInterface
```
PeterMis: can we typehint string for $id here as well?

PeterMis (changes requested):
Looks good overall, just couple sugestions:
- We should use PHP8 Constructor Promotion in the controller/factory to make it cleaner where possible
- Return simple JSON responses instead of getState/render where possible ( i.e. if the endpoint is api / react only we can do that, if it's still in use by twigs then we can leave it for now )
- Couple comments in regards to the $id in some functions below - especially the searchById needs to be changed please

frontend/src/pages/Administration/Modules/components/ClearCacheModuleModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <div className="bg-red-100 w-15.5 h-15.5 text-red-800 border-red-50 border-[7px] rounded-full p-3">
+              <Trash2Icon size={26} />
+            </div>
+          </div>
+          <h2 className="text-center text-lg font-semibold mb-2 text-red-800">
+            {t('Clear Cache?')}
+          </h2>
+        </div>
+        <p className="text-center text-gray-500">
+          {t('Clear the cache for ')}
```
maurofmferrao: Splitting a sentence across multiple t() calls makes translation impossible,  translators see two disconnected fragments and can't reorder them. I'd collapse it into a single call with an interpolation:

```
<p className="text-center text-gray-500">
  {t('Clear the cache for {{name}}. This process is not reversible.', { name: module.name })}
</p>
```

maurofmferrao (changes requested):
Just a minor comment, everything else looks good to me!

maurofmferrao (dismissed):

PeterMis (dismissed):

mgbaybay:
TEST: all

mgbaybay:
@PeterMis, @maurofmferrao Requesting review again since some tests failed earlier. Thank you!

maurofmferrao (approved):

PeterMis (approved):

mgbaybay:
TEST: all

PR #3440, opened 2026-04-24
DataSet: filter clause changes

dasgarner:
Changes:
 - tidy up and remove unused functions
 - make is-empty/is-not-empty filtering clearer
 - throw on invalid filter, saves executing (which would likely fail anyway)
 - misc extra filtering and sanitisation


mgbaybay (approved):

ifarzana:

PR #3451, opened 2026-04-28
Module Templates: page implementation

mgbaybay:
## Tasks
- [x] https://github.com/xibosignageltd/xibo-private/issues/1512
- [x] https://github.com/xibosignageltd/xibo-private/issues/1385

----
### Backend Changes

**Added:**
- `keyword` search
- `searchById` route implementation

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1512, https://github.com/xibosignageltd/xibo-private/issues/1385


frontend/src/pages/Developer/ModuleTemplates/components/tabs/GeneralTab.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const dataTypeOptions: SelectOption[] = dataTypes.map((dt) => ({
+    label: dt.name,
+    value: dt.id,
+  }));
+
+  return (
+    <div className="flex flex-col gap-6">
+      {/* Options */}
+      <div className="bg-white rounded-lg p-5 flex flex-col gap-4">
+        <p className="font-bold text-l">{t('Template Identity')}</p>
```
maurofmferrao: `text-l` isn't a Tailwind class, is it supposed to be `text-lg`?

frontend/src/pages/Developer/ModuleTemplates/components/tabs/PropertiesTab.tsx
```diff
@@ -0,0 +1,938 @@
...
+  'tickerTagStyle',
+  'worldClock',
+].map((t) => ({ label: t, value: t }));
+
+// ── Shared section label ──────────────────────────────────────────────────────
+
+function SectionLabel({ label, action }: { label: string; action?: React.ReactNode }) {
+  return (
+    <div className="flex items-center justify-between">
+      <span className="text-s font-semibold text-gray-500 tracking-wide gap-1.5 inline-flex">
```
maurofmferrao: Same as the comment for `text-l`, is `text-s` suppose to be `text-xs`?

frontend/src/pages/Developer/ModuleTemplates/hooks/useModuleTemplateEditForm.ts
```diff
@@ -0,0 +1,80 @@
...
+    hbs: template.stencil?.hbs ?? '',
+    style: template.stencil?.style ?? '',
+    head: template.stencil?.head ?? '',
+    onTemplateRender: template.onTemplateRender ?? '',
+    onTemplateVisible: template.onTemplateVisible ?? '',
+    properties: template.properties ?? [],
+  };
+}
+
+export const useModuleTemplateEditForm = (template: ModuleTemplate | null) => {
```
maurofmferrao, line 47: After a successful save the baseline (`originalValues`) is never updated, so `hasUnsavedChanges` stays true and Cancel (line 71) reverts to the *first-loaded* values rather than what was just saved. Expose a `markSaved()` that sets `originalValues` to the current `formValues` and call it from `handleSave` after the request resolves. The `template && !formValues` guard in the effect (line 51) also means the form won't reinitialise if the route `id` changes without a remount.

frontend/src/pages/Developer/ModuleTemplates/ModuleTemplateEdit.tsx
```diff
@@ -0,0 +1,250 @@
...
+                >
+                  {t(tab.labelKey)}
+                </button>
+              ))}
+            </div>
+          </div>
+
+          {/* Right content panel */}
+          <div className="flex-1 flex flex-col overflow-hidden min-h-0">
+            {isError && (
```
maurofmferrao: This error block is unreachable, the earlier `if (isError || !template) return …` (which renders the same message at line 111) bails out before we get here. Consider removing it.

lib/routes.php
```diff
@@ -841,3 +841,27 @@
 $app->group('', function (RouteCollectorProxy $group) {
     $group->post('/schedule/sync/add', ['\Xibo\Controller\Schedule', 'syncAdd'])->setName('schedule.add.sync');
 })->addMiddleware(new FeatureAuth($app->getContainer(), ['schedule.sync']));
+
+//
+// Developer
+//
+$app->group('', function (\Slim\Routing\RouteCollectorProxy $group) {
```
PeterMis: Should those routes be accessible via API?

If the intention is for it to be web ui only ( as I think it was previously? ) then perhaps we should move those routes to routes-react?

frontend/src/components/ui/forms/SelectDropdown.tsx
```diff
@@ -222,7 +222,7 @@ export default function SelectDropdown({
       {label && (
         <label
           id={`${id}-label`}
-          className="flex items-center justify-between text-sm font-semibold text-gray-500 leading-5"
```
maurofmferrao, line 225: We should avoid making changes to global components, if it's something we want to change for a specific page, we need to extend the className for the component so we can keep the original styling.

frontend/src/components/ui/forms/TextInput.tsx
```diff
@@ -68,7 +68,7 @@ export default function TextInput({
   const generatedId = useId();
 
   return (
-    <div className={twMerge('flex flex-col gap-1 w-full', wrapperClassName)}>
+    <div className={twMerge('flex flex-col gap-2 w-full', wrapperClassName)}>
```
maurofmferrao: Same as above

maurofmferrao (changes requested):
Just some minor changes

PeterMis (dismissed):
Backend looks good to me!

Just a note in regards to npm for consideration ( and perhaps for Mauro to comment on )
We've added a new dependencies in this PR in package.json, would we expect the updated package-lock.json file to also be included in this PR as well?

frontend/src/pages/Developer/ModuleTemplates/ModuleTemplatesConfig.tsx
```diff
@@ -0,0 +1,209 @@
...
+  t,
+  onEdit,
+  onCopy,
+  onDelete,
+  onShare,
+}: ModuleTemplateActionsProps): ((template: ModuleTemplate) => ActionItem[]) => {
+  return (template: ModuleTemplate) => {
+    const actions: ActionItem[] = [
+      {
+        label: t('Edit'),
```
maurofmferrao: Edit is shown for all rows, but searchById throws AccessDenied when checkEditable fails, non-owners get a 403 after clicking Edit. Gate this like Copy/Delete.

frontend/src/pages/Developer/ModuleTemplates/components/tabs/CodeTab.tsx
```diff
@@ -0,0 +1,73 @@
...
+  value: string;
+  onChange: (value: string) => void;
+  language: 'twig' | 'handlebars' | 'css' | 'html' | 'javascript';
+}
+
+export default function CodeTab({ value, onChange, language }: CodeTabProps) {
+  const { t } = useTranslation();
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = async () => {
```
maurofmferrao, line 36: Can we wrap the clipboard write in try/catch? In a non-secure context navigator.clipboard.writeText rejects, producing an unhandled rejection and leaving the button stuck.

maurofmferrao (changes requested):
Some other minor changes, and as Peter said, we also need to update `package-lock.json`, you can do it by running `npm install --package-lock-only`.

PeterMis (approved):

maurofmferrao (approved):
LGTM!

mgbaybay:
TEST: all

PeterMis (approved):

mgbaybay:
TEST: all

maurofmferrao (approved):

PR #3460, opened 2026-04-30
Logs: backend improvement and cleanup

mgbaybay:
## Tasks
- [x] Frontend: Log page
- [x] https://github.com/xibosignageltd/xibo-private/issues/1388

----
### Backend Changes

**Added:**
- `searchById` route

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1388


PeterMis (dismissed):
Looks good to me

One thing that isn't clear from just the code changes - are the `/json/log` and  `/json/user/pref` etc correctly excluded when excludeLog=1 is in the filter?
( most of those will be sql queries logged, so more in test CMS mode )

Not critical, hence I've approved the PR, but if we could have a look that would be great

PeterMis (approved):

mgbaybay:
TEST: all

PR #3469, opened 2026-05-05
Frontend: Agenda view in Calendar

maurofmferrao:
relates to xibosignageltd/xibo-private#1384


frontend/src/pages/Schedule/Schedule/components/AgendaModal.tsx
```diff
@@ -0,0 +1,1380 @@
...
+          ) : (
+            <ChevronsUpDown className="size-4" />
+          )}
+        </div>
+      </div>
+    </th>
+  );
+}
+
+function EventTypeTable({
```
chaolen, line 902: I'm not sure if this table is intentional, or if we should extend our `DataTable` with a couple of new props for more flexibility. We could add `onRowClick` and `getRowClassName`(row) props to `DataTable` to support clickable-row functionality.

frontend/src/pages/Schedule/Schedule/components/AgendaModal.tsx
```diff
@@ -0,0 +1,1380 @@
...
+  onGeoLatChange,
+  onGeoLngChange,
+  onGetLocation,
+  onClearLocation,
+  showMap,
+  onToggleMap,
+  defaultLat,
+  defaultLng,
+}: {
+  showTimeline: boolean;
```
chaolen: Could we move these props into a named type above the function? Just for consistency with what we do in our codebase. Same goes for the other functions that use inline props.

chaolen (changes requested):
Everything looks good! Just a few comments for code adjustments. Feel free to disregard if any are intentional.

frontend/src/pages/Schedule/Schedule/components/AgendaModal.tsx
maurofmferrao: The custom table is intentional, Datatable is built for server-side pagination and sorting, while this one operates entirely on the client side. If in the future we need more client-side tables, maybe we can create a new component. Let's leave it like this for now. 👍🏻

chaolen (dismissed):
LGTM!

chaolen (approved):

PR #3494, opened 2026-05-11
QA: (Part 2) Display edit form

ifarzana:
- https://github.com/xibosignageltd/xibo-private/issues/1343

Part 2

- Integration tests for the edit flow, modal opens and data refreshes correctly after saving.
- Tests for row actions - each action opens the correct modal or triggers the expected side effect.
- Tests for column visibility and filter functionality.
- Rendering tests to check for the presence of key UI elements and the correct display of data.


maurofmferrao (approved):

chaolen (approved):

ifarzana:
TEST: all

PR #3498, opened 2026-05-11
Display Send Command: duplicates and not populating field

maurofmferrao:
relates to xibosignageltd/xibo-private#1409

 - Fix linter errors for tests
 - Eliminate unnecessary re-fetch on close


chaolen (approved):
LGTM!

PR #3508, opened 2026-05-14
Migrations: fix failed upgrade and add new indexes for performance

dasgarner:
fix: old migrations for saved report movement and latest news URL. - xibosignage/xibo##3862
feat: new indexes for layout actions to improve XMDS performance - xibosignage/xibo##3859

bonus: also adjust old latest news feed during install (mimics upgrade)


mgbaybay (approved):

dasgarner:
The test for this is a clean installation, which can be verified by running the test suite.

dasgarner:
TEST: all

PR #3519, opened 2026-05-16
Lazy-load HLS.js and PDF.js vendor libraries per-module

dasgarner:
## Summary

Refactor vendor library loading to lazy-load HLS.js and PDF.js only on modules that require them, rather than bundling them globally in the player bundle. This reduces the initial bundle size and improves load performance for displays that don't use HLS or PDF widgets.

## Changes

- **New vendor entry points**: Created `vendor-hls.js` and `vendor-pdfjs.js` as separate webpack entry points that expose `window.Hls` and `window.pdfjsLib` respectively
- **Updated webpack config**: Added `vendor-hls` and `vendor-pdfjs` entries to the module build configuration
- **Removed global includes**: Removed HLS.js and PDF.js requires from `player_bundle.js` with a comment explaining they now load per-module via `<assets>` declarations
- **Module asset declarations**: 
  - Added `hlsLib` asset to `hls.xml` pointing to `/modules/vendor-hls.min.js`
  - Added `pdfjsLib` asset to `pdf.xml` pointing to `/modules/vendor-pdfjs.min.js`
- **Build artifacts**: Updated `.gitignore` to exclude minified vendor files (`modules/vendor-*.min.js*`)

## Implementation Details

The vendor libraries are now conditionally loaded only when their respective modules are instantiated. Each module declares its vendor dependency via the `<assets>` section with `isAutoInclude="true"`, ensuring the library is available when needed without bloating the core player bundle.

https://claude.ai/code/session_01JEzQCz9p3KvpFPWnj1sQQb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ results.xml
 containers/
 web/dist
 modules/bundle*.js*
+modules/vendor-*.min.js*
 modules/assets/pdfjs
 
 # The OpenOOH specification
--- a/modules/hls.xml
+++ b/modules/hls.xml
@@ -122,4 +122,7 @@ video {
 }
         ]]></style>
     </stencil>
+    <assets>
+        <asset id="hlsLib" type="path" mimeType="text/javascript" isAutoInclude="true" path="/modules/vendor-hls.min.js"></asset>
+    </assets>
 </module>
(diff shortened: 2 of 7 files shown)


dasgarner:
@maurofmferrao @rubenberttpingol this is something that came to mind last night - PDF/HLS libraries are big and we're loading them for every widget - but we rarely use them. 

Particularly on low powered SoCs it can take ages to parse the bundle.min.js - I think this drops its size by nearly 50%.

What do you think?

Also Claude can't sign the CLA, that will be interesting. 😀

PR #3522, opened 2026-05-19
Frontend: Remove dead catch blocks, debug logs, and unsafe error casts

maurofmferrao:
- Promise.allSettled never rejects, so catch blocks in confirmDelete/confirmLogout were dead code. Removed across all 15 action hooks

- Replaced unsafe `as { response?... }` error casts with `isAxiosError()` in 26 locations across modals and components.

- Removed two stray console.log statements from useLayoutActions and useTemplateActions.

- Updated 12 test files to mock Axios errors with isAxiosError: true so they  match the type guard.


chaolen (approved):
LGTM!

PR #3436, opened 2026-04-22
Fonts: Page Implementation

mgbaybay:
## Tasks
- [x] Frontend: Fonts page
- [x] https://github.com/xibosignageltd/xibo-private/issues/1335

----
### Backend Changes

**Added:**
- `sortBy` and `sortDir` sorting logic
- `keyword` search
- `userPermissions` in JSON response
- `searchById` route implementation

**Removed:**  
- Buttons and grid in return data
- Unused twig files and routes

Relates to https://github.com/xibosignageltd/xibo-private/issues/1335

### Frontend Changes

- Created new Fonts Page

relates to: https://github.com/xibosignageltd/xibo-private/issues/1432

PR #3524, opened 2026-05-19
Frontend: Fonts page

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1432

Note: added minor changes on `FileUploader.tsx`

### DIFF
--- a/frontend/src/pages/Administration/Fonts/components/DeleteFontModal.tsx
+++ b/frontend/src/pages/Administration/Fonts/components/DeleteFontModal.tsx
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2026 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Trash2Icon } from 'lucide-react';
+import { useTranslation } from 'react-i18next';
+
+import Modal from '@/components/ui/modals/Modal';
+
+interface DeleteFontModalProps {
+  isOpen?: boolean;
+  onClose: () => void;
+  onDelete: () => void;
+  itemCount: number;
+  fontName?: string;
+  error?: string | null;
+  isLoading?: boolean;
+}
+
+export default function DeleteFontModal({
+  isOpen = true,
+  onClose,
+  onDelete,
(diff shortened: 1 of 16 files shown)


frontend/src/pages/Administration/Fonts/components/FontDetailsModal.tsx
```diff
@@ -0,0 +1,156 @@
...
+export default function FontDetailsModal({
+  isOpen = true,
+  onClose,
+  fontId,
+  font,
+}: FontDetailsModalProps) {
+  const { t } = useTranslation();
+  const [fontFaceFamily, setFontFaceFamily] = useState<string | null>(null);
+  const [isFontLoading, setIsFontLoading] = useState(false);
+
```
maurofmferrao, line 49: We can have a FontFace memory leak here, we need to cleanup `document.fonts`. 

We can start by adding `const fontFaceRef = useRef<FontFace | null>(null);` here, and at line 71 add `fontFaceRef.current = loadedFace;` and finally on the useEffect cleanup at line 85 we can add the following block:

```
  if (fontFaceRef.current) {
    document.fonts.delete(fontFaceRef.current);
    fontFaceRef.current = null;
  }
```

frontend/src/pages/Administration/Fonts/components/DeleteFontModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+            </div>
+          </div>
+          <h2 className="text-center text-lg font-semibold mb-2 text-red-800">
+            {itemCount === 1 ? t('Delete Font?') : t('Delete Fonts?')}
+          </h2>
+        </div>
+        <p className="text-center text-gray-500">
+          {itemCount === 1 ? (
+            <>
+              {t('Are you sure you want to delete ')}"<strong>{fontName}</strong>?"
```
maurofmferrao, line 81: To prevent a broken translated string here, we might need to add something like `{t('Are you sure you want to delete "{{name}}"?', { name: fontName })}` and `{t('Are you sure you want to delete {{count}} fonts?', { count: itemCount })}`, but we will lose the functionality of the strong, which I think it preferable.

frontend/src/pages/Administration/Fonts/components/FontModals.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import DeleteFontModal from './DeleteFontModal';
+import FontDetailsModal from './FontDetailsModal';
+import UploadFontModal from './UploadFontModal';
+
+import type { Font } from '@/types/font';
+
+interface FontModalsProps {
+  actions: {
+    activeModal: string | null;
```
maurofmferrao, line 30: Use `activeModal: ModalType` instead.

frontend/src/schema/font.ts
```diff
@@ -0,0 +1,45 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import type { TFunction } from 'i18next';
+import { z } from 'zod';
+
+const VALID_FONT_EXTENSIONS = ['.otf', '.ttf', '.eot', '.svg', '.woff'];
+
+export const getFontUploadSchema = (t: TFunction) =>
```
maurofmferrao, line 27: This is never imported or used, should it be removed?

maurofmferrao (changes requested):
Just some minor changes. Also, can we cherry pick this into the PR and branches that already have the back-end changes for fonts?

Branch: feature/neujmin_fonts_page
PR: https://github.com/xibosignage/xibo-cms/pull/3436

PR #3539, opened 2026-05-22
Test Suite : ESLint should only run on develop.

PeterMis:
As the ESLint is configured to run on the `/frontend` folder, therefore it likely should only run on develop.

- Israt review request more so like a CC for awareness ;)


dasgarner (approved):

PeterMis:
TEST: all

PR #3546, opened 2026-05-27
Fix clipped tag dropdown in filter form

J-Rodriguez10:
Fixes part of the tag filtering UI clipping issue reported in the Xibo Community forum:
https://community.xibo.org.uk/t/tag-input-and-dropdown-ui-elements-become-visually-clipped-in-cms-filtering-interface/37952 

This change updates the filter form overflow behavior so tag dropdown suggestions remain visible instead of being clipped by the filter container.

Tested locally using the Docker CMS setup by comparing the behavior before and after the CSS change.

Before: The dropdown menu was clipped, with no way for the user to see or access it:
image

After: The dropdown menu is easy to see now:
image

### DIFF
--- a/web/theme/default/css/xibo.css
+++ b/web/theme/default/css/xibo.css
@@ -992,7 +992,7 @@ ul li.active.tab-design:hover a {
 }
 
 .XiboFilter .form-inline {
-    overflow: auto;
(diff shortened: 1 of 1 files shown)


J-Rodriguez10:

PR #3549, opened 2026-05-28
Frontend: Welcome Page

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1469


frontend/src/pages/Welcome/Welcome.tsx
```diff
@@ -0,0 +1,205 @@
...
+  };
+
+  // Filter service cards by user features
+  const visibleServiceCards = SERVICE_CARDS.filter((card) =>
+    hasRequiredFeatures(card.featureFlags),
+  );
+
+  // Filter resource cards (some are Xibo-branded only)
+  // TODO: Replace with actual theme check when available
+  const isXiboThemed = true;
```
PeterMis: Would be good to make a note of that somewhere, so we don't forget, as it's fairly important :)

I think the Connectors PR introduced the `app_name` to the return from `/me` so, once that's merged we could re-use that here probably

lib/routes.php
```diff
@@ -547,6 +547,10 @@
     ->addMiddleware(new \Xibo\Middleware\FeatureAuth($app->getContainer(), ['users.add']))
     ->setName('user.add');
 
+// Welcome wizard state
```
PeterMis, line 552: Could we add TODO here, to move those 2 routes to routes-json once we have it please?

I think it's also introduced in the Connectors PR - as both of those routes should be only available in the web ui they should eventually end up in routes-json

PeterMis (dismissed):
Looks good to me, just 2 not really code related comments below

frontend/src/pages/Welcome/WelcomeConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+  {
+    serviceId: 'displays',
+    titleKey: 'Displays',
+    descriptionKey:
+      "Displays are managed from the CMS and control when and how content is shown. Get started by installing one of our Player App's to start scheduling content to show on Displays.",
+    imageUrl: '/theme/default/img/onboarding/displays.png',
+    featureFlags: ['displays.view'],
+    links: [
+      { labelKey: 'Get Started', to: '/displays/displays', variant: 'primary' },
+      { labelKey: 'Learn More', to: '', isExternal: true, variant: 'link' },
```
maurofmferrao: We're missing the "Learn More" links in the SERVICE_CARDS, is that to be added later?

frontend/src/pages/Welcome/Welcome.tsx
```diff
@@ -0,0 +1,205 @@
...
+      return (
+        <Link key={link.to} to={link.to}>
+          <Button variant={variant}>{t(link.labelKey)}</Button>
+        </Link>
+      );
+    });
+  };
+
+  return (
+    <div ref={containerRef} className="relative h-full overflow-y-auto">
```
maurofmferrao: Do we need `containerRef` here?

frontend/src/pages/Welcome/Welcome.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const { user } = useUserContext();
+  const [videoModalData, setVideoModalData] = useState<VideoLink[] | null>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const isSuperAdmin = user?.userTypeId === UserType.SuperAdmin;
+
+  // Mark welcome wizard as seen on first visit
+  useEffect(() => {
+    if (user?.newUserWizard === 0) {
+      markWelcomeSeen();
```
maurofmferrao: If the request fails silently, `newUserWizard` stays `0` and the user lands here on every login. 

We could use this to avoid the redirect loop on failure:
```
markWelcomeSeen().catch(() => {});
```

frontend/src/pages/Welcome/WelcomeConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+  xiboOnly?: boolean;
+  links: {
+    labelKey: string;
+    url: string;
+    superAdminOnly?: boolean;
+  }[];
+}
+
+// Onboarding quick-nav cards — scroll to corresponding service card section
+export const ONBOARDING_CARDS: OnboardingCard[] = [
```
maurofmferrao: This is not being imported anywhere, did we remove it from the `Welcome.tsx` design?

frontend/src/services/welcomeApi.ts
```diff
@@ -0,0 +1,30 @@
...
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import http from '@/lib/api';
+
+export async function markWelcomeSeen(): Promise<void> {
+  await http.put('/user/welcome');
+}
+
+export async function markWelcomeUnseen(): Promise<void> {
```
maurofmferrao: `markWelcomeUnseen` is not being used anywhere else, should we remove it?

maurofmferrao (changes requested):
Just some questions and some suggestions to add to Peter's review.

maurofmferrao (dismissed):
LGTM!

maurofmferrao (approved):
LGTM!

lib/routes.php
```diff
@@ -549,6 +549,10 @@
...
+$app->put('/user/welcome', ['\Xibo\Controller\User','userWelcomeSetSeen'])->setName('welcome.wizard.seen');
```
PeterMis, line 553: We could probably move the seen/unseen routes to routes-react instead

PeterMis (approved):
Looks good to me - just one comment on the routes, but it's not critical, I just think since those were in web routes they probably should be in react routes now

lib/routes.php
chaolen: Will take note of this. Thanks, Peter!

chaolen:
TEST: all

PR #3552, opened 2026-05-29
Frontend: Import Layout

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1416

### DIFF
--- a/frontend/src/pages/Design/Layouts/components/ImportLayoutModal.tsx
+++ b/frontend/src/pages/Design/Layouts/components/ImportLayoutModal.tsx
@@ -1,0 +1,525 @@
+/*
+ * Copyright (C) 2026 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


frontend/src/pages/Design/Layouts/components/ImportLayoutModal.tsx
```diff
@@ -0,0 +1,525 @@
...
+
+  // Sync name/tags changes to the backend after import completes
+  useEffect(() => {
+    if (isSyncing) {
+      return;
+    }
+
+    const itemToSync = files.find(
+      (f) =>
+        f.status === 'completed' && f.hasUnsavedChanges && f.layoutId && (f.retryCount || 0) < 3,
```
maurofmferrao: The retry guard `(f.retryCount || 0) < 3` will never fire: on failure, `hasUnsavedChanges` is set to false, so `itemToSync` is never found again. Either remove retryCount entirely and let failures be permanent, or keep `hasUnsavedChanges: true` on error so the three-attempt logic actually runs.

As written, retryCount and the < 3 check are dead code.

maurofmferrao (changes requested):
Just a quick change, everything else looks good!

frontend/src/pages/Design/Layouts/components/ImportLayoutModal.tsx
chaolen: Good catch — the `retryCount ` was carried over from `useUploadQueue.ts` which uses the same pattern for syncing edits after upload. Removed it here since it was dead code. Failures are now permanent and the error is shown to the user.

Side note: `useUploadQueue.ts ` has the same bug. `isDirty: false` on error prevents the retry logic from ever running.

maurofmferrao: Can we also fix it for `useUploadQueue.ts` in this PR?

maurofmferrao (approved):

chaolen:
TEST: all

PR #3558, opened 2026-06-03
Frontend: Reports pages

maurofmferrao:
relates to xibosignageltd/xibo-private#1475


frontend/src/pages/Reporting/AllReports/ExportStatisticsModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+          '/stats/getExportStatsCount',
+          {
+            params,
+            headers: { 'X-Requested-With': 'XMLHttpRequest' },
+            signal: controller.signal,
+            withCredentials: true,
+          },
+        );
+        if (!cancelled) {
+          setRecordCount(response.data.data.total);
```
chaolen: the backend returns the count directly via `->withJson($resultSet)`, so `response.data.data.total` will be undefined. It should just be response.data (or the backend needs to wrap it in { total: count }).

frontend/src/pages/Reporting/AllReports/ExportStatisticsModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+      setCountError(null);
+      try {
+        const params: Record<string, string> = {
+          fromDt: toPhpDatetime(fromDt),
+          toDt: toPhpDatetime(toDt),
+        };
+        if (displayId) {
+          params.displayId = displayId;
+        }
+        const response = await axios.get<{ data: { total: number } }>(
```
chaolen: This uses raw `axios` with manual `withCredentials` and `X-Requested-With` headers, but `reportApi.ts` already uses http from @/lib/api which handles those defaults. I'd suggest moving this to a fetchExportStatsCount() function in reportApi.ts to stay consistent with the rest of the service layer.

The function in reportApi.ts would look something like:

```
export const fetchExportStatsCount = async (
    params: { fromDt: string; toDt: string; displayId?: string },
    signal?: AbortSignal,
  ): Promise<number> => {
    const response = await http.get<number>('/stats/export/count', {
      params,
      signal,
    });
    return response.data;
  };
```

One more thing, `/stats/getExportStatsCount` no longer exists. The route was moved to `/stats/export/count` (defined in routes-react.php).

frontend/src/pages/Reporting/ReportSchedules/components/ToggleActiveModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const confirmLabel = isPausing
+    ? isLoading
+      ? t('Pausing…')
+      : t('Yes, Pause')
+    : isLoading
+      ? t('Resuming…')
+      : t('Yes, Resume');
+  const message = isPausing
+    ? t('Are you sure you want to pause ')
+    : t('Are you sure you want to resume ');
```
chaolen: This should use `<Trans>` with interpolation like the Delete and Reset modals do — string concatenation with t() breaks translations for languages that need to reorder the sentence.

chaolen (changes requested):
Looks good overall. Just few change requests:

chaolen (dismissed):
LGTM!

chaolen (dismissed):

chaolen (approved):

maurofmferrao:
TEST: all

PR #3564, opened 2026-06-05
Frontend: Screenshot Thumbnail and Modal for Displays

maurofmferrao:
relates to xibosignageltd/xibo-private#1522

### DIFF
--- a/frontend/src/pages/Displays/Displays/DisplaysConfig.tsx
+++ b/frontend/src/pages/Displays/Displays/DisplaysConfig.tsx
@@ -798,11 +798,7 @@ export const getDisplayColumns = (props: DisplayActionsProps): ColumnDef<Display
       enableSorting: false,
       cell: ({ row }) => (
         <MediaCell
-          thumb={
-            row.original.screenShotRequested
-              ? `/display/screenshot/${row.original.displayId}`
-              : undefined
-          }
+          thumb={row.original.thumbnail || undefined}
           alt={row.original.display}
           mediaType="image"
           onPreview={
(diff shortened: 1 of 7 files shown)


frontend/src/pages/Displays/Displays/DisplaysConfig.tsx
```diff
@@ -797,9 +798,16 @@ export const getDisplayColumns = (props: DisplayActionsProps): ColumnDef<Display
       enableSorting: false,
       cell: ({ row }) => (
         <MediaCell
-          thumb={`/display/screenshot/${row.original.displayId}`}
+          thumb={
+            row.original.screenShotRequested
+              ? `/display/screenshot/${row.original.displayId}`
+              : undefined
+          }
```
chaolen: The `screenShotRequested `conditional appears to be inverted. `screenShotRequested = 1` means the screenshot is pending (requested but not yet delivered by the player). Once the player uploads the screenshot, it resets to 0. So this condition hides the thumbnail when a screenshot file is actually available, and only shows it during the pending state.

frontend/src/services/displaysApi.ts
```diff
@@ -291,6 +291,26 @@ export async function requestScreenShot(displayId: number | string): Promise<voi
...
+  const response = await http.get(`/display/screenshot/${displayId}`, {
+    responseType: 'blob',
+    signal,
+  });
+  return response.data;
+}
+
+export type DisplayStatusWindow = Record<string, string | number> | string | unknown[];
+
+export async function fetchDisplayStatusWindow(
```
chaolen, line 307: `fetchDisplayStatusWindow `doesn't accept an `AbortSignal` parameter, unlike `fetchDisplayScreenshotBlob` which does. Can we add AbortSignal here to be consistent? Without it, if a user rapidly switches between displays, stale requests can't be cancelled.

chaolen (changes requested):
Just minor tweaks, but everything else looks good.

chaolen (approved):
LGTM!

maurofmferrao:
TEST: all

PR #3565, opened 2026-06-08
Feat: new Folders Page

chaolen:
- [x] Frontend Changes:

relates to: https://github.com/xibosignageltd/xibo-private/issues/1473
- New Folder Page
- Update `FolderBreadCrumbs `-  Replaced N serial `fetchFolderById `calls with single `fetchFolderTree `+ in-memory tree walk.

- [x] Backend
relates to: https://github.com/xibosignageltd/xibo-private/issues/1526


mgbaybay (dismissed):

mgbaybay (dismissed):

frontend/src/utils/date.ts
```diff
@@ -24,6 +24,17 @@ export type ExpiryValue =
   | { type: 'preset'; value: string }
   | { type: 'datePicked'; date: Date };
 
+export function formatDate(date: Date, timeZone?: string) {
```
maurofmferrao, line 32: Both `formatDate` and `formatDateTime` rely on locale formats to produce the strings: `'en-US'` + `.replace(/\//g, '-')` for `MM-DD-YYYY`, and `'sv-SE'` + `.replace(',', '')` for `YYYY-MM-DD HH:MM:SS`.

The fix can be something like `formatToParts()` : it returns named parts (`year`, `month`, `day`…)
regardless of locale, so you assemble the string yourself:

```
  function getParts(date: Date, options: Intl.DateTimeFormatOptions) {
    const parts = new Intl.DateTimeFormat('en', options).formatToParts(date);
    const get = (type: string) => parts.find((p) => p.type === type)?.value ?? '00';
    return get;
  }

  export function formatDate(date: Date, timeZone?: string) {
    const get = getParts(date, { year: 'numeric', month: '2-digit', day: '2-digit', ...(timeZone ? { timeZone } : {}) });
    return `${get('year')}-${get('month')}-${get('day')}`;
  }

  export function formatDateTime(date: Date, timeZone?: string) {
    const get = getParts(date, { year: 'numeric', month: '2-digit', day: '2-digit', hour: '2-digit', minute: '2-digit', second: '2-digit', hour12: false, ...(timeZone ? { timeZone } : {}) });
    return `${get('year')}-${get('month')}-${get('day')} ${get('hour')}:${get('minute')}:${get('second')}`;
  }
```

Also a good opportunity to align both functions on ISO order (`YYYY-MM-DD`),
since `formatDate` currently outputs month-first and `formatDateTime` year-first.

frontend/src/pages/Administration/Folders/components/FolderInfoPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  if (error || !folder) {
+    return (
+      <div className="flex items-center justify-center h-full text-sm text-red-500">
+        {error || t('Folder not found')}
+      </div>
+    );
+  }
+
+  const childrenRaw = folder.children as unknown;
```
maurofmferrao: The `as unknown` cast is hiding a real type mismatch: `/folders/{id}` returns `children` as a raw comma-separated string from the DB, but `Folder.children` is typed as `Folder[] | null`. I'd make the type reflect what the API actually sends: `Folder[] | string | null`,  and handle the normalisation once in the service layer or a shared helper, rather than here inline with a cast.

frontend/src/pages/Administration/Folders/components/FolderInfoPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+            variant="tertiary"
+            leftIcon={RefreshCw}
+            onClick={() => setLocalRefresh((n) => n + 1)}
+          >
+            {t('Refresh')}
+          </Button>
+        </div>
+        <DataTable
+          columns={usageColumns}
+          data={usage}
```
maurofmferrao: `data={usage}` is passed unsorted, when the user clicks a column header, `sorting` state updates but the row order never changes.

Derive a sorted copy before passing it in:
```
  const sortedUsage = sorting.length > 0
    ? [...usage].sort((a, b) => {
        const key = sorting[0].id as keyof FolderUsageEntry;
        return sorting[0].desc
          ? String(b[key]).localeCompare(String(a[key]))
          : String(a[key]).localeCompare(String(b[key]));
      })
    : usage;
```

Then pass `data={sortedUsage}` to the table.

maurofmferrao (changes requested):
Just some changes. Everything else looks good!

frontend/src/utils/date.ts
chaolen: I kept `formatDate` as `MM-DD-YYYY` intentionally — the FolderInfoPanel uses that format for display. The alignment suggestion makes sense for consistency, but the US date format is what we want base on the UI design.

maurofmferrao (approved):
LGTM!

mgbaybay (dismissed):

chaolen:
TEST: all

chaolen:
TEST: all

maurofmferrao (dismissed):

PeterMis:
TEST: all

PeterMis:
I don't see failing tests on develop itself, but we seem to have couple that fail on this branch, as such we should investigate if it's due to something we've introduced here - I will try to have a look

maurofmferrao (approved):
LGTM!

PeterMis:
TEST: all

PR #3576, opened 2026-06-11
Notifications: page implementation

mgbaybay:
## Tasks
- [x] https://github.com/xibosignageltd/xibo-private/issues/1510
- [x] https://github.com/xibosignageltd/xibo-private/issues/1511

----
### Backend Changes
**Improved:**
- Backend logic and refactoring

**Removed:**  

Relates to https://github.com/xibosignageltd/xibo-private/issues/1510, https://github.com/xibosignageltd/xibo-private/issues/1511


lib/Controller/Notification.php
```diff
@@ -25,114 +25,82 @@
...
         $notification->setRead(Carbon::now()->format('U'));
         $notification->save();
 
+        if ($this->isJson($request) || $this->isApi($request)) {
+            return $response
+                ->withStatus(200)
+                ->withJson($notification);
+        }
+
         $this->getState()->template = 'notification-interrupt';
```
PeterMis: I think we can probably drop the support for twig way of doing this?

lib/Controller/Notification.php
```diff
@@ -143,12 +111,16 @@ public function show(Request $request, Response $response, $id)
...
-            $this->getState()->setData(['notification' => $notification]);
-
-            return $this->render($request, $response);
+        } else if ($this->isJson($request) || $this->isApi($request)) {
+            return $response
+                ->withStatus(200)
+                ->withJson($notification);
         }
+
+        $this->getState()->template = 'notification-form-show';
```
PeterMis: Same as interrupt, do we still need to handle this?
We could also remove the associated twig files

lib/Controller/Notification.php
```diff
@@ -223,249 +220,144 @@ public function grid(Request $request, Response $response): Response|\Psr\Http\M
...
-        foreach ($this->userGroupFactory->query(['`group`'], ['isUserSpecific' => -1]) as $userGroup) {
-            /* @var UserGroup $userGroup */
-
-            if ($userGroup->isUserSpecific == 0) {
-                $userGroups[] = $userGroup;
-            } else {
-                $users[] = $userGroup;
-            }
-        }
+        $notification = $this->notificationFactory->getById($id);
```
PeterMis: Does this bypass the ACL check?
should we check if it's viewable by the user?

lib/Controller/Notification.php
```diff
@@ -223,249 +220,144 @@ public function grid(Request $request, Response $response): Response|\Psr\Http\M
...
-            if ($displayGroup->isDisplaySpecific == 1) {
-                $displays[] = $displayGroup;
-            } else {
-                $groups[] = $displayGroup;
+        if ($notificationId !== null) {
+            $notification = $this->userNotificationFactory->getByNotificationId($notificationId);
+            $notification->setRead($readDt);
+            $notification->save();
+        } else {
+            foreach ($this->userNotificationFactory->query(
```
PeterMis: Could we make this more efficient?
Perhaps one simple SQL update statement to set read=1, readDt="now()" for the specific user?

Minor thing, but if there would be a lot of notifications, going through them 1 by 1 is not very efficient

lib/Entity/Notification.php
```diff
@@ -110,6 +110,12 @@ class Notification implements \JsonSerializable
     #[OA\Property(description: 'Additional email addresses to which a saved report will be sent')]
     public $nonusers;
 
+    /**
+     * @var int
+     */
+    #[OA\Property(description: 'Flag whether the current user has read this notification')]
+    public $read = 0;
```
PeterMis: Are we sure this belongs here? Wouldn't it be always 0? Which could be confusing
I think this is more so related to the UserNotification entity

lib/routes.php
```diff
@@ -68,16 +68,18 @@
  * Notification
  */
 $app->get('/notification', ['\Xibo\Controller\Notification','grid'])->setName('notification.search');
+$app->get('/notification/export/{id}', ['\Xibo\Controller\Notification', 'exportAttachment'])->setName('notification.exportattachment');
```
PeterMis: This was web only route I believe?
Should we move it to routes-react instead?

Unless we think this should be available via api now?

lib/routes-react.php
```diff
@@ -108,6 +108,14 @@
         ->setName('playlistdashboard.show');
 })->add(new \Xibo\Middleware\FeatureAuth($app->getContainer(), ['dashboard.playlist']));
 
+//
+// notification
+//
+$app->get('/notification/mynotifications', ['\Xibo\Controller\Notification', 'myNotifications'])->setName('notification.mynotifications');
+$app->get('/notification/interrupt', ['\Xibo\Controller\Notification', 'getInterrupt'])->setName('api.notification.interrupt');
```
PeterMis: is there a reason why the naming convention here is a bit different?
I mean the `api.` perhaps even more so since this isn't actually available via API :)

lib/Middleware/Actions.php
```diff
@@ -109,10 +109,14 @@ public function process(Request $request, RequestHandler $handler): Response
 
                 // User notifications
                 $notifications = array_merge($notifications, $factory->getMine());
+
+                /*
+                 * TODO: Remove this once the SPA migration is complete since the interrupt notification
```
PeterMis: I think we just get rid of it here, no need to leave this commented out I don't think?

frontend/src/components/layout/NotificationDropdown.tsx
```diff
@@ -0,0 +1,325 @@
...
+        };
+      },
+    );
+
+    setIsMarkingRead(true);
+    try {
+      await markAllNotificationsRead();
+      queryClient.invalidateQueries({ queryKey: notificationQueryKeys.all });
+    } catch {
+      console.error('Error updating the notifications');
```
maurofmferrao: Should we show an error toast notification here?

maurofmferrao (dismissed):
LGTM with a minor suggestion that can be ignored

lib/Controller/Notification.php
```diff
@@ -25,114 +25,82 @@
...
-    }
-
-    /**
-     * @param Request $request
-     * @param Response $response
-     * @return \Psr\Http\Message\ResponseInterface|Response
-     * @throws \Xibo\Support\Exception\ControllerNotImplemented
-     * @throws \Xibo\Support\Exception\GeneralException
-     */
-    public function displayPage(Request $request, Response $response)
```
PeterMis, line 93: While removing this function and associated route is correct
We should ensure that nothing that is still on legacy twigs attempts to call `notification.view` route ( or any other remoed route ) to avoid errors

Hint: the authed-notification-drawer.twig calls it, for now we could just change the URL to the new notification centre perhaps? With a view to remove this twig once nothing uses it?

PeterMis (changes requested):
Overall looks very good, just couple of minor comments below please

lib/Controller/Notification.php
```diff
@@ -884,4 +639,159 @@ public function exportAttachment(Request $request, Response $response, $id)
...
         ));
     }
+
+    /**
+     * Build system notifications for the current request context.
+     * These are never stored in the database; they reflect live environment state.
+     * @param Request $request
+     * @return array
+     */
+    private function buildSystemNotifications(Request $request): array
```
PeterMis: I see what we aim to achieve here and I think it's probably OK

My only suggestion would be to change this function name to something like `buildEnvironmentWarnings` and the follow up function to `buildEnvironmentWarning` just to avoid potential confusion with the function that actually creates system notifications ( saved in db ) in NotificationFactory.

lib/Controller/Notification.php
```diff
@@ -884,4 +639,159 @@ public function exportAttachment(Request $request, Response $response, $id)
...
+    {
+        $notifications = [];
+
+        // TODO: Might need further updates for the SPA
```
PeterMis: What do we think should go here?
If we don't know then we probably can remove the TODO
if we do know then we should add what's missing and remove the TODO?

lib/Controller/Notification.php
```diff
@@ -223,249 +204,140 @@ public function grid(Request $request, Response $response): Response|\Psr\Http\M
...
-     * @return \Psr\Http\Message\ResponseInterface|Response
-     * @throws AccessDeniedException
-     * @throws \Xibo\Support\Exception\ControllerNotImplemented
-     * @throws \Xibo\Support\Exception\GeneralException
-     * @throws \Xibo\Support\Exception\NotFoundException
+     * @return Response|ResponseInterface
+     * @throws NotFoundException
      */
-    public function deleteForm(Request $request, Response $response, $id)
+    public function myNotifications(Request $request, Response $response): Response|ResponseInterface
```
PeterMis, line 309: Should we check if user has featureEnabled('drawer') somewhere in this function?
( old Actions.php checked that before adding the environment notifications I believe )

PeterMis (commented):
Looks good, just minor comments below to address please

PeterMis (dismissed):
LGTM

maurofmferrao (dismissed):

mgbaybay:
TEST: all

maurofmferrao (dismissed):

PeterMis (dismissed):

mgbaybay:
TEST: all

maurofmferrao (approved):

mgbaybay:
TEST: all

PR #3582, opened 2026-06-15
Notifications : Fix interrupt notification query logic.

### LINKED ISSUE #3872: Creating a notification with “Interrupt” stops user from accessing the CMS
When an administrator creates a system notification and enables the "Interrupt" flag, user are entirely blocked from accessing the CMS and are met with a error screen: "Sorry there has been an unknown error."

**Steps to reproduce**
1. Create a new notification
2. Set the Date and Time (select a time that will occur immediately or with a 5-minute delay)
3. Enable 'Interrupt'checkbox
4. Click Save
5. Wait for the scheduled notification time to arrive/display.
6. Attempt to log in to the CMS as a user targeted by this notification or the user created the notification

**Actual result**
When the user tries to log in, the authentication flow fails to process the active interrupt flag, and the following error message is shown:
"Sorry there has been an unknown error."
The user is completely locked out of the CMS and cannot bypass this screen.

**Expected result**
The user should be able to log in successfully and cleanly view the full-screen interrupt notification modal before proceeding to their standard dashboard.

Image

PeterMis:
fixes xibosignage/xibo#3872

### DIFF
--- a/lib/Factory/UserNotificationFactory.php
+++ b/lib/Factory/UserNotificationFactory.php
@@ -144,7 +144,7 @@ class UserNotificationFactory extends BaseFactory
 
         $params = ['now' => Carbon::now()->format('U')];
         $select = 'SELECT `lknotificationuser`.lknotificationuserId,
-            `lknotificationuser`.notificationId,
+            `notification`.notificationId,
             `lknotificationuser`.userId,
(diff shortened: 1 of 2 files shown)


mgbaybay (approved):

PeterMis:
TEST: all

PR #3584, opened 2026-06-15
Quickfix: Hisense override settings

chaolen:


frontend/src/pages/Displays/Displays/components/EditDisplayModal.tsx
```diff
@@ -907,12 +957,12 @@ export default function EditDisplayModal({
...
     if (name === 'hisensePictureOptions' && isHisenseType) {
-      const obj: Record<string, number> = {};
+      const obj: Record<string, number | null> = {};
       HISENSE_PICTURE_SETTING_KEYS.forEach((key) => {
-        const v =
-          profileFlat[key] !== undefined ? profileFlat[key] : (profileDefaults[key] ?? null);
-        if (v !== null && v !== undefined && v !== '') {
-          obj[key] = Number(v);
+        if (profileFlat[key] !== undefined) {
+          obj[key] = Number(profileFlat[key]);
```
maurofmferrao: The old version skipped empty-string values; now `Number('')` coerces them to `0`, and this also disagrees with the back-fill in the load effect (lines 777–778) which still filters `pv !== ''`. If a picture-option profile value can ever be `''`, this will show `0`. Was that intended, or should `''` be treated the same as `null` here for consistency?

frontend/src/pages/Displays/Displays/components/EditDisplayModal.tsx
```diff
@@ -222,12 +222,16 @@ function summarizeHisenseTimers(str: string, t: TFunction): string {
...
 function summarizeHisensePictureOptions(str: string, t: TFunction): string {
   try {
-    const parsed = JSON.parse(str) as Record<string, number>;
+    const parsed = JSON.parse(str) as Record<string, number | null>;
     const entries = Object.entries(parsed).filter(([key]) => key !== '');
     if (entries.length === 0) {
       return '—';
     }
     const names = entries.map(([key, val]) => {
+      if (val === null) {
```
maurofmferrao, line 231: `gammaMode` and `dynamicContrast` aren't in `HISENSE_PICTURE_PROPERTY_DEFS` (only the 5 sliders are), so when either is `null` this branch returns the raw key `"gammaMode"` instead of the friendly "Gamma" / "Dynamic Contrast" label the non-null cases below produce. Could you special-case those two keys here, e.g. fall through to a label lookup that mirrors the cases on lines 235–241?

maurofmferrao (changes requested):
LGTM with some warnings

maurofmferrao (approved):
LGTM!

chaolen:
TEST: all

PR #3614, opened 2026-06-23
Logs: fixed filter and message formatting

mgbaybay:
### Frontend Changes

- Modified data fetching so that it only loads result once filters are set (`onApply`)
- Fixed message formatting in logs

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1562


frontend/src/components/ui/FilterInputs.tsx
```diff
@@ -314,6 +316,7 @@ export default function FilterInputs<T>({
             {t('Reset')}
           </Button>
         )}
+        <div className="grid grid-cols-[repeat(auto-fit,minmax(15rem,1fr))] gap-4 items-end">
         {options.map((filter) => {
```
maurofmferrao: This file has a lot of linter errors because of that new div, can we please format it?

frontend/src/pages/Advanced/Logs/LogsConfig.tsx
```diff
@@ -26,6 +26,12 @@ import type { FilterConfigItem } from '@/components/ui/FilterInputs';
 import { TextCell } from '@/components/ui/table/cells';
 import type { LogEntry } from '@/types/log';
 
+function decodeHtmlEntities(encoded: string): string {
```
maurofmferrao, line 29: If `message` ever arrives null/undefined, `textarea.innerHTML = null` coerces to the string `"null"` and renders it in the cell (previously a nullish value rendered as blank). A short guard covers it:

```
function decodeHtmlEntities(encoded: string): string {
  if (!encoded) {
    return '';
  }
  const textarea = document.createElement('textarea');
  textarea.innerHTML = encoded;
  return textarea.value;
}
```

maurofmferrao (changes requested):
LGTM with just some minor changes

maurofmferrao (approved):
LGTM!

mgbaybay:
TEST: all

PR #3636, opened 2026-07-01
Media: fixed background upload card should show on all pages

mgbaybay:
### Frontend Changes

- Moved background upload card to Root so that it shows on all pages during upload or when user navigates away from the media page

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1576


chaolen (approved):

mgbaybay:
TEST: all

PR #3643, opened 2026-07-03
fix: Library enable stat multi should notify

### LINKED ISSUE #3860: 4.4.2 Reports - The dummy layout's enableStat attribute remains stale after enabling stats on a media item that was scheduled before stats were turned on.
When an image or video is scheduled directly from the Media Library (creating a dummy layout) while "Enable stats collection" is off, the dummy layout XLF is built with enableStat="0". If stats collection is subsequently enabled on that media item, the dummy layout is never marked as requiring a rebuild. After a layout build cycle, the XLF retains the old enableStat="0" value — meaning stats are silently not collected despite the user having enabled them.

Steps to reproduce
1. Upload an image to the Media Library.
2. Set Enable stats collection to Off on the media item.
3. From the Media Library, directly schedule the image to a display (e.g. a Windows player) and save.
4. Inspect the resulting dummy layout's XLF — confirm enableStat="0" at both layout and media node levels.
5. Go to Library, edit the media item, set Enable stats collection to On, and save.
6. Wait for or force a layout build cycle (e.g. wait for the display to collect).
7. Inspect the dummy layout's XLF again.

Actual result
The dummy layout XLF retains the stale enableStat="0". The layout is never invalidated after the media item's stat setting is changed, so no rebuild occurs.

Expected result
Saving a change to enableStat on a media item should invalidate any associated dummy layouts and trigger a rebuild. The rebuilt XLF should reflect enableStat="1" at both the layout and media node levels.

Image

dasgarner:
fixes xibosignage/xibo#3860

### DIFF
--- a/lib/Controller/Library.php
+++ b/lib/Controller/Library.php
@@ -22,8 +22,6 @@
 namespace Xibo\Controller;
 
 use Carbon\Carbon;
-use GuzzleHttp\Client;
-use Xibo\Helper\Guzzle\SafeClient;
 use Illuminate\Support\Str;
 use Intervention\Image\ImageManagerStatic as Img;
 use OpenApi\Attributes as OA;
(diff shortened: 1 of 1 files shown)


mgbaybay (approved):

flashdave:
TEST: all

PR #3651, opened 2026-07-07
Remove prototype url prefix

PeterMis:
fixes https://github.com/xibosignageltd/xibo-private/issues/1608


dasgarner (dismissed):

lib/routes.php
maurofmferrao: This route line is pre-existing, but this PR newly exposes it to a collision, the same class you fixed for `/schedule/{id}` and `/library/{id}`.

The SPA route `user/force-change-password` has no PHP route; it's served via the NotFound to `app-spa.twig` fallback. It's reached by a full-page navigation: `LoginApp.tsx` does `window.location.assign('/user/force-change-password') `after login, and `Actions.php` issues a server 302 to the same URL for legacy page loads. Since `force-change-password` moved out from under the old `/prototype/` prefix, Slim now matches this unconstrained route with `id="force-change-password"` and runs `User::searchById()`, so a user required to change their password gets JSON/404 instead of the force-change page. Constrain it to numeric:


```
$app->get('/user/{id:[0-9]+}', ['\Xibo\Controller\User','searchById'])->setName('user.search.id');

```
Worth confirming there are no other unconstrained `GET /<seg>/{id}` routes whose `<seg>` matches a two-segment SPA path.

maurofmferrao (changes requested):
LGTM with a 🤖 finding

lib/routes.php
PeterMis: That's probably a good point, I will get that fixed tomorrow, thanks ~~Clau~~ Mauro!

maurofmferrao (approved):
LGTM!

PeterMis:
TEST: all

PR #3662, opened 2026-07-09
Frontend: Preview Campaign Page

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1637

Added fix related to: https://github.com/xibosignageltd/xibo-private/issues/1613

### DIFF
--- a/frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
+++ b/frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
@@ -19,8 +19,9 @@
  * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { useQuery } from '@tanstack/react-query';
+import { useInfiniteQuery, useQuery } from '@tanstack/react-query';
 import { ArrowLeft, Maximize2 } from 'lucide-react';
+import { useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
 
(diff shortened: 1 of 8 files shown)


frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const {
+    data: layoutsData,
+    isLoading: layoutsLoading,
+    isError: isLayoutsError,
+    error: layoutsError,
+  } = useQuery({
+    queryKey: ['campaign-preview-layouts', campaignId],
+    queryFn: () =>
+      fetchLayouts({
```
maurofmferrao, line 111: Two issues here:

-  No `sortBy`, so layouts come back alphabetically (`LayoutFactory::query` defaults to `layout ASC`) rather than in campaign playback order. For a layout-list campaign the sequence is the point of a preview. The fix needs backend help: `displayOrder` isn't in the factory's `$allowedColumns`, so add it there, then pass `sortBy: 'displayOrder'` here.

-  `length: 100` silently truncates larger campaigns while the header shows the true `numberLayouts`. At minimum surface "showing first 100", ideally paginate.

frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+import type { Layout } from '@/types/layout';
+import { formatDuration } from '@/utils/formatters';
+
+function LayoutPreviewCard({ layout }: { layout: Layout }) {
+  const { t } = useTranslation();
+
+  return (
+    <div className="flex flex-col gap-2 rounded-sm overflow-hidden shadow-sm">
+      <div className="relative w-full aspect-4/3 bg-black rounded overflow-hidden">
+        {layout.previewUrl && (
```
maurofmferrao: `previewUrl` is only populated by the backend when the `layout.view` feature is enabled, but this route is gated on `campaign.view`. A user with campaign.view but not layout.view lands here and sees only black boxes with no explanation. Consider a fallback message (or an empty-state) when `previewUrl` is absent so it doesn't look broken.

frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <iframe
+            sandbox="allow-scripts"
+            src={layout.previewUrl}
+            title={`Layout ${layout.layoutId}`}
```
maurofmferrao: This title string isn't translated, the labels just below use `t()`.

Consider `title={t('Layout {{id}}', { id: layout.layoutId })}` for consistency.

frontend/src/pages/Design/Campaigns/CampaignPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+            className="absolute inset-0 w-full h-full border-0 overflow-hidden"
+            style={{ pointerEvents: 'auto' }}
```
maurofmferrao: `pointerEvents: 'auto'` is the default, is this doing anything? If not, I'd drop the inline style; if you did mean to force interaction, a `className` would fit the rest of the file better.

maurofmferrao (dismissed):
LGTM with warnings

maurofmferrao (approved):
LGTM

chaolen:
TEST: all

chaolen:
TEST: all

PR #3667, opened 2026-07-10
Update branding on the default favicon.ico file.

PeterMis:
- Update favicon.ico with new branding
- Introduce .sidecar hash checks for default files, adjust logic for when we should copy default files from docker/brand 
( this will allows us to check if the image hash is different, so after any image in docker/brand was updated, and if we are on Xibo branded install we copy over the updated file automatically )


maurofmferrao (approved):
LGTM

PeterMis:
TEST: all

maurofmferrao (approved):

PeterMis:
TEST: all

PR #3671, opened 2026-07-13
Reports: various fixes

mgbaybay:
### Frontend Changes

- Updated reports filters order
- Fixed missing input fields when saving a report
- Fixed missing options from the input fields
- Fixed issue on display report filter handling

### Backend Changes
- Added error handling for large file export in the reports
- Updated backend query for display reports data

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1602, https://github.com/xibosignageltd/xibo-private/issues/1648


frontend/src/pages/Reporting/Reports/shared/hooks/useLayoutOptions.ts
```diff
@@ -0,0 +1,50 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect, useState } from 'react';
+
+import type { MultiSelectOption } from '@/components/ui/forms/MultiSelectDropdown';
+import { fetchLayouts } from '@/services/layoutsApi';
+
+const MAX_LAYOUTS = 100;
```
maurofmferrao: This (and the matching MAX_MEDIA / MAX_CAMPAIGNS) hard-caps the dropdown at the first 100 rows with no search or load-more. On larger installs, layouts/media/campaigns past the first 100 become unselectable in the Proof of Play filters and schedule modal. `useDisplayOptions` already solves this with `onSearch`/`onLoadMore`/`hasMore`,  could we follow that pattern here rather than a fixed cap?

frontend/src/pages/Reporting/Reports/ProofOfPlay/components/ProofOfPlayFilters.tsx
```diff
@@ -134,29 +167,31 @@ export default function ProofOfPlayFilters({
...
+                value={(filter.logicalOperator as 'AND' | 'OR') || 'OR'}
+                onChange={(val) => onFilterChange({ logicalOperator: val })}
+              />
+            }
             suffix={
-              <button
-                type="button"
+              <div
+                className="px-3 py-2 flex items-center cursor-pointer justify-center"
                 onClick={() => onFilterChange({ exactTags: !filter.exactTags })}
```
maurofmferrao, line 200: Moving the `onClick`/`title` onto a non-focusable `<div>` and nesting an unlabelled `<Button>` inside means the focusable control has no accessible name and no pressed state for screen readers.

Could we make it a single control instead, put the toggle directly on the `<Button>` with `aria-label={t('Match exact characters only')}` and `aria-pressed={filter.exactTags}`?

Same visual result, better semantics. (Same pattern in `ProofOfPlay/components/AddScheduleModal.tsx` around line 195.)

frontend/src/services/savedReportApi.ts
```diff
@@ -80,3 +80,29 @@ export async function fetchSavedReportData(
...
+  } catch (err) {
+    // When responseType is 'blob', axios wraps the error body as a Blob too.
+    // Convert it back to a plain Error with the server's message.
+    if (err && typeof err === 'object' && 'response' in err) {
+      const axiosErr = err as { response?: { data?: unknown } };
+      if (axiosErr.response?.data instanceof Blob) {
+        try {
+          const text = await (axiosErr.response.data as Blob).text();
+          const json = JSON.parse(text) as { message?: string };
+          if (json.message) throw new Error(json.message);
```
maurofmferrao: This conflates two cases. When the error blob isn't valid JSON, `JSON.parse` throws a `SyntaxError` that this `catch` happily rethrows, so the user sees "Unexpected token '<'…" instead of the real failure. I'd extract the message without rethrowing the parse error:
```ts
        const text = await (axiosErr.response.data as Blob).text();
        let message: string | undefined;
        try { message = (JSON.parse(text) as { message?: string }).message; } catch { /* not JSON */ }
        if (message) throw new Error(message);
```

maurofmferrao (changes requested):
LGTM with some changes

maurofmferrao (approved):
LGTM!

PeterMis (approved):

mgbaybay:
TEST: all

PR #3672, opened 2026-07-13
Add tests for Modules page and modals

flashdave:
https://github.com/xibosignageltd/xibo-private/issues/1547


ifarzana (approved):
I have a small comment: we should consider a test case ensuring that non-superadmins are never able to change settings.

flashdave:
> I have a small comment: we should consider a test case ensuring that non-superadmins are never able to change settings.

Sure, we can include those in our upcoming tests.

flashdave:
TEST: all

PR #3675, opened 2026-07-13
Applications : OAuth code flow form.

PeterMis:
fixes xibosignageltd/xibo-private#1621

### DIFF
--- a/frontend/src/pages/Administration/Applications/Authorize/AuthorizeApplication.tsx
+++ b/frontend/src/pages/Administration/Applications/Authorize/AuthorizeApplication.tsx
@@ -30,6 +30,7 @@ import { withPublicPath } from '@/config/publicPath';
 import { useBranding } from '@/context/BrandingContext';
 import { ErrorBanner } from '@/login/components/ErrorBanner';
 import { LoginCard } from '@/login/components/LoginCard';
+import { isSafeHttpUrl } from '@/utils/url';
 
 import '@/login/styles.css';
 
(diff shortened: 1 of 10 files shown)


frontend/src/pages/Administration/Applications/Authorize/AuthorizeApplication.tsx
```diff
@@ -0,0 +1,162 @@
...
+          </ul>
+
+          {application.description && (
+            <p className="authorize-description">{application.description}</p>
+          )}
+
+          {(application.termsUrl || application.privacyUrl) && (
+            <p className="authorize-links">
+              {application.termsUrl && (
+                <a href={application.termsUrl} target="_blank" rel="noreferrer">
```
maurofmferrao: `termsUrl` and `privacyUrl` are set by the application owner and saved with a plain
`getString()` (no scheme validation, see Applications.php:437-438). This consent screen is
rendered to a different authenticated user during the OAuth flow, so an owner who sets
`termsUrl` to `javascript:...` gets script execution in that user's CMS session when they
click the link. React 19 does not strip `javascript:` hrefs (it only logs a dev warning), and
there is no app-wide URL sanitizer. The click barrier is weak here because the screen actively
invites the user to review the terms before approving.

Preferred fix is server side, so every consumer is covered: validate the scheme on save with an http/https allow-list, e.g.
```
    $url = $sanitizedParams->getString('termsUrl');
    if ($url !== '' && !preg_match('/^https?:\/\//i', $url)) {
        throw new InvalidArgumentException(__('Terms URL must be an http(s) URL'), 'termsUrl');
    }
```

As defence in depth, also guard here so only http(s) links render:

```
    const safe = (u: string | null) => (u && /^https?:\/\//i.test(u) ? u : undefined);
    ...
    {safe(application.termsUrl) && (
      <a href={safe(application.termsUrl)} target="_blank" rel="noreferrer">{t('Terms')}</a>
    )}
```

maurofmferrao (changes requested):
Just one change flagged by Claude

mgbaybay (approved):

maurofmferrao (approved):
LGTM

PeterMis:
TEST: all

PR #3689, opened 2026-07-16
Fix: Missing component on Layout copy modal

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1612

Note: 
- Sub issue: https://github.com/xibosignageltd/xibo-private/issues/1612#issuecomment-4866208401 will be implemented in a different issue. The fix requires a validation for the Tags that requires value which should be updated across all the pages that uses `TagInput`.
- Added minor changes on `Campaigns.tsx` (revert correct column conditions)


maurofmferrao (approved):
LGTM

chaolen:
TEST: all

PR #3692, opened 2026-07-16
Fix: Tag Value Validation

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1726

### DIFF
--- a/frontend/src/components/ui/forms/TagInput.tsx
+++ b/frontend/src/components/ui/forms/TagInput.tsx
@@ -35,6 +35,7 @@ import { useEffect, useId, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { twMerge } from 'tailwind-merge';
 
+import SelectDropdown from './SelectDropdown';
 import { useTagSuggestions } from './hooks/useTagSuggestions';
 
 import { fetchTags } from '@/services/tagApi';
@@ -112,6 +113,7 @@ interface TagInputProps {
   onInputChange?: (value: string) => void;
   allowValues?: boolean;
   suggestions?: boolean;
+  onPendingValueChange?: (isPending: boolean) => void;
 }
 
 function TagInput({
@@ -130,6 +132,7 @@ function TagInput({
   onInputChange,
   allowValues = true,
   suggestions = true,
+  onPendingValueChange,
 }: TagInputProps) {
   const { t } = useTranslation();
   const inputId = useId();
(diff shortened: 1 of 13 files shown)


frontend/src/pages/Design/Campaigns/CampaignEditor.tsx
```diff
@@ -716,7 +718,7 @@ export default function CampaignEditor() {
         <Button variant="secondary" onClick={() => navigate('/design/campaign')}>
           {t('Back')}
         </Button>
-        <Button variant="primary" disabled={saveGeneral.isPending} onClick={handleSaveGeneral}>
+        <Button variant="primary" disabled={saveGeneral.isPending || hasTagPendingValue} onClick={handleSaveGeneral}>
```
maurofmferrao: This line fails the ESLint tests, needs to be fixed with something like:

```
        <Button
          variant="primary"
          disabled={saveGeneral.isPending || hasTagPendingValue}
          onClick={handleSaveGeneral}
        >
```

frontend/src/components/ui/forms/TagInput.tsx
```diff
@@ -445,31 +453,12 @@ function TagInput({
...
-            <select
-              id={valueInputId}
-              autoFocus
-              className={twMerge(
-                'w-full text-sm px-3 py-2.5 rounded-lg border border-gray-200 outline-none focus:border-blue-500 focus:ring-1 focus:ring-blue-500 min-h-11.25 bg-white',
-                valueError && 'border-red-500 focus:border-red-500 focus:ring-red-500',
-              )}
-              aria-invalid={valueError}
-              aria-describedby={valueError ? `${valueInputId}-err` : undefined}
+            <SelectDropdown
```
maurofmferrao, line 456: The old `<select>` had an empty option plus `onBlur → applyValue()` and `Escape → cancelValueEntry`, so an optional tag with options could be left blank.

`SelectDropdown` here has no blank affordance and isn't `clearable`, and since Save is now disabled while pending, the user gets trapped, contradicting the "or leave blank" help text just below. Adding `clearable` fixes both cases cleanly: for optional, selecting the placeholder → `applyValue('')` closes the pending state; for required, `applyValue('')` still sets `valueError` and keeps it open:

```
  (...)
  onSelect={(val) => applyValue(val)}
  clearable
  (...)
```

maurofmferrao (changes requested):
Just 2 minor changes

maurofmferrao (approved):
LGTM

chaolen:
TEST: all

PR #3697, opened 2026-07-16
Add unit tests for Users  page

ifarzana:
I added `aria-label`s to the Users checkboxes and switches so screen readers and vitests can identify each control by name instead of relying on their position in the UI. These are attribute-only changes with no impact on existing functionality.


maurofmferrao (approved):
LGTM

ifarzana:
TEST: all

PR #3702, opened 2026-07-17
Permissions: update user permissions and sharing column behavior

mgbaybay:
### Frontend Changes

- Modified sharing column to use the global sharing cell component
- Added sorting in `Sharing` column
- Fixed permission issues on media, usergroups, playlists, and datasets

 
### Backend Changes
- Updated middleware permissions
- Added `sharing` column sorting in various pages

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1614, https://github.com/xibosignageltd/xibo-private/issues/1605

### DIFF
--- a/frontend/src/components/ui/table/cells/SharingCell.tsx
+++ b/frontend/src/components/ui/table/cells/SharingCell.tsx
@@ -32,7 +32,7 @@ import {
 } from '@floating-ui/react';
 import type { ColumnDef } from '@tanstack/react-table';
 import type { TFunction } from 'i18next';
-import { useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { useLayoutEffect, useRef, useState } from 'react';
 
 import Badge from '../../Badge';
 
@@ -109,7 +109,7 @@ function useVisibleCount(
 }
 
 export function SharingCell({ groups, privatePlaceholder = '' }: SharingCellProps) {
-  const entries = useMemo(() => parseGroups(groups), [groups]);
+  const entries = parseGroups(groups);
 
   const containerRef = useRef<HTMLDivElement>(null);
   const pillRefs = useRef<(HTMLSpanElement | null)[]>([]);
--- a/frontend/src/components/ui/table/cells/SharingCell.tsx
+++ b/frontend/src/components/ui/table/cells/SharingCell.tsx
@@ -1,0 +1,229 @@
+/*
+ * Copyright (C) 2026 Xibo Signage Ltd
+ *
+ * Xibo - Digital Signage - https://xibosignage.com
+ *
+ * This file is part of Xibo.
+ *
+ * Xibo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * Xibo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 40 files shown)


lib/Controller/Schedule.php
```diff
@@ -726,6 +730,28 @@ public function add(Request $request, Response $response): Response|ResponseInte
...
+            // Only images and videos may be scheduled directly - other media types
+            // (audio, pdf, etc) have no meaningful full-screen presentation.
+            if ($type === 'media' && !in_array($scheduledItem->mediaType, ['image', 'video'], true)) {
+                throw new InvalidArgumentException(
+                    __('Only images and videos can be scheduled directly'),
+                    'mediaId'
+                );
+            }
+
+            if ($scheduleWithViewOnItem && !$this->getUser()->checkViewable($scheduledItem)) {
```
PeterMis: I think this is great in Schedule add/edit ( this whole block with media type check as well )
How about the Layout createFullScreenLayout() ?
POST /layout/fullscreen is it's own api exposed route perhaps the same permission checks should be added there?
Unless we handled that in a different way already

lib/Controller/Schedule.php
```diff
@@ -726,6 +730,28 @@ public function add(Request $request, Response $response): Response|ResponseInte
                 );
             }
 
+            // Permission check for the media/playlist
+            $scheduleWithViewOnItem = ($this->getConfig()->getSetting('SCHEDULE_WITH_VIEW_PERMISSION') == 1);
+            $scheduledItem = ($type === 'media')
+                ? $this->mediaFactory->getById($id)
+                : $this->playlistFactory->getById($id);
+
+            // Only images and videos may be scheduled directly - other media types
```
PeterMis: Should we perhaps re-order this, so that we check the permissions first and if it's not denied then we check the media type?

Same for both add and edit schedule routes

PeterMis (changes requested):
Looks great, just 2 comments below

frontend/src/pages/Administration/Users/components/tabs/GeneralTab.tsx
```diff
@@ -121,28 +121,30 @@ export default function GeneralTab({
...
+              className="px-3 text-gray-500 hover:text-gray-700 transition-colors"
+            >
+              {showPassword ? <EyeOff size={16} /> : <Eye size={16} />}
+            </button>
+          }
+        />
+      )}
 
-      {isEdit && (
+      {(isSuperAdmin && isEdit) && (
```
maurofmferrao: ESlint error here: 

```
Replace `(isSuperAdmin && isEdit)` with `isSuperAdmin && isEdit`
```

frontend/src/components/ui/table/cells/SharingCell.tsx
```diff
@@ -0,0 +1,229 @@
...
+    const observer = new ResizeObserver(recompute);
+    observer.observe(container);
+    return () => observer.disconnect();
+  }, [entries, containerRef, pillRefs, plusRef]);
+
+  return visibleCount;
+}
+
+export function SharingCell({ groups, privatePlaceholder = '' }: SharingCellProps) {
+  const entries = useMemo(() => parseGroups(groups), [groups]);
```
maurofmferrao: We don't use `useMemo`, the `React Compiler` memoizes this for us. I'd change it to:

```
const entries = parseGroups(groups);
```

maurofmferrao (changes requested):
Just some minor changes

maurofmferrao (approved):
LGTM

PeterMis (approved):

mgbaybay:
TEST: all

PR #3712, opened 2026-07-21
CMS Permissions

maurofmferrao:
- [x] https://github.com/xibosignageltd/xibo-private/issues/1738
- [x] https://github.com/xibosignageltd/xibo-private/issues/1745
- [x] https://github.com/xibosignageltd/xibo-private/issues/1747
- [x] https://github.com/xibosignageltd/xibo-private/issues/1715
- [x]  https://github.com/xibosignageltd/xibo-private/issues/1711
- [x]  https://github.com/xibosignageltd/xibo-private/issues/1744
- [x]  https://github.com/xibosignageltd/xibo-private/issues/1615
- [x]  https://github.com/xibosignageltd/xibo-private/issues/1635


frontend/src/pages/Administration/Users/UsersConfig.tsx
```diff
@@ -131,45 +139,59 @@ export const getUserItemActions = ({
...
         icon: Edit,
         onClick: () => onEdit(user),
         isQuickAction: true,
         variant: 'primary' as const,
-      },
+      });
 
       // Dropdown menu actions
-      {
+      actions.push({
```
PeterMis, line 155: do we need a per-row check instead of global canModify based on the users.modify?

User::checkEditable() requires super-admin, self-ownership, or shared group membership for a Group Admin.

Delete never checks that the target shares a group with the acting admin. The backend already returns a per-row userPermissions field on the User grid ( which mirrors checkEditable's group-overlap logic exactly) and the frontend User type already declares userPermissions?: UserPermissions - it's just not being read here. 

Every sibling grid (Layout, Playlist, MenuBoard, DisplayGroup) correctly uses userPermissions?.edit/delete; Users is the outlier. A Group Admin will see Edit on users outside their group and get a 403.

frontend/src/pages/Design/Templates/Templates.tsx
```diff
@@ -267,6 +265,11 @@ export default function Templates() {
 
   const columns = getTemplateColumn({
     t,
+    canModify: hasFeature(user, 'template.modify'),
```
PeterMis: should this be `layout.modify` instead? Or maybe both - double check what the route we are trying to reach here actually has as a gate ( likely layout.modify )

frontend/src/pages/Displays/DisplayGroup/DisplayGroupConfig.tsx
```diff
@@ -169,97 +185,122 @@ export const getDisplayGroupItemActions = ({
...
+        label: t('Assign Files'),
+        onClick: () => openAssignFilesModal(displayGroup),
+      });
+      actions.push({
+        label: t('Assign Layouts'),
+        onClick: () => openAssignLayoutsModal(displayGroup),
+      });
+    }
+
+    if ((canModify && canEdit) || canLimitedView) {
```
PeterMis: A user with only displaygroup.limitedView (no displaygroup.modify, no command.view) sees both buttons and gets a 403. 

The Displays-page version doesn't check the displaygroup.* keys, so its "limited view" gating is checking an unrelated feature namespace entirely. 

Fix: gate "Send Command" on canModify || hasFeature(user, 'command.view'), and drop the canLimitedView bypass for "Trigger a web hook" entirely (backend has no OR-path for it).

frontend/src/pages/Displays/DisplayProfile/DisplayProfileConfig.tsx
```diff
@@ -78,46 +78,62 @@ export const getBaseFilterKeys = (t: TFunction): FilterConfigItem<DisplayProfile
...
-    {
-      label: t('Delete'),
-      icon: Trash2,
-      onClick: () => onDelete(displayProfile.displayProfileId),
-      variant: 'danger' as const,
-    },
-  ];
+    const actions: ActionItem[] = [];
+
+    if (canModify) {
```
PeterMis: Are we ignoring the owner check here and only rely on the feature flag? ( same for other if (canModify below )

There are some hard checks for owner ( or super admin ) server side like
```
if ($this->getUser()->userTypeId != 1 && $this->getUser()->userId != $displayProfile->userId) {
    throw new AccessDeniedException(__('You do not have permission to edit this profile'));
}
```
frontend likely should care about that as well

PeterMis (changes requested):
Looks good, just couple of  comments to look at below please

frontend/src/pages/Displays/Displays/DisplaysConfig.tsx
```diff
@@ -580,19 +582,21 @@ export const getDisplayItemActions = ({
       });
     }
 
-    if (limitedEdit) {
+    if (canModify && canEdit) {
```
PeterMis: I think the canModify here on display for send command/trigger webhook might want to check displaygroup.modify ( and command.view ) - to match the backend, otherwise the buttons could be visible, but still return 403 server side

frontend/src/pages/Design/Templates/Templates.tsx
```diff
@@ -60,6 +60,7 @@ export default function Templates() {
   const { user } = useUserContext();
   const canViewFolders = usePermissions()?.canViewFolders;
   const canTag = hasFeature(user, 'tag.tagging');
+  const canMoveToCampaignFolder = hasFeature(user, 'campaign.modify');
```
PeterMis, line 63: I think that's great for the move, we still should address the previous comment about canModify ( it likely want to check layout.modify instead of template.modify ? )

frontend/src/pages/Administration/UserGroups/UserGroupsConfig.tsx
```diff
@@ -86,8 +86,8 @@ export const getUserGroupItemActions = ({
     const canManageMembers =
       isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
     const canEdit = isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
-    const canCopy = isSuperAdmin && hasUsergroupModify;
-    const canDelete = isSuperAdmin && hasUsergroupModify;
+    const canCopy = isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
+    const canDelete = isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
```
PeterMis: I think canDelete wants to be super admin only here
( UserGroup delete() throws 403 for non super admins )

PeterMis (changes requested):
Good changes ( although one small regression in UserGroupsConfig )
just a few comments below:

PeterMis (dismissed):
LGTM

frontend/src/pages/Administration/UserGroups/UserGroupsConfig.tsx
```diff
@@ -85,29 +85,34 @@ export const getUserGroupItemActions = ({
     // specific group, surfaced here as userGroup.userPermissions.edit.
     const canManageMembers =
       isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
+    const canEdit = isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
+    const canCopy = isSuperAdmin || (hasUsergroupModify && !!userGroup.userPermissions?.edit);
```
mgbaybay: Are lines 88-89 intentional or can we simplify it since they have the same condition with the `canManageMembers`?

mgbaybay (commented):
Looks good! just one minor clarification

mgbaybay (approved):

maurofmferrao:
TEST: all

PeterMis (approved):

ifarzana:
TEST: /Schedule

PR #3729, opened 2026-07-23
Search: query optimization

mgbaybay:
### Backend Changes
- Search term optimization for special characters sanitization

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1757

### DIFF
--- a/lib/Factory/BaseFactory.php
+++ b/lib/Factory/BaseFactory.php
@@ -577,8 +577,8 @@ class BaseFactory
         $conditions = [];
 
         foreach ($terms as $term) {
-            if (preg_match('/^[a-zA-Z]+$/', $term) && strlen($term) >= 4) {
-                $fulltextTerms[] = $term;
+            if (preg_match('/^[a-zA-Z]+$/', $term) && strlen($term) >= 3) {
+                $fulltextTerms[] = $this->sanitizeBooleanModeTerm($term);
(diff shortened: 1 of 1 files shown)


lib/Factory/BaseFactory.php
```diff
@@ -577,8 +577,8 @@ public function buildSearchQuery(
         $conditions = [];
 
         foreach ($terms as $term) {
-            if (preg_match('/^[a-zA-Z]+$/', $term) && strlen($term) >= 4) {
-                $fulltextTerms[] = $term;
+            if (preg_match('/^[a-zA-Z]+$/', $term) && strlen($term) >= 3) {
+                $fulltextTerms[] = $this->sanitizeBooleanModeTerm($term);
```
PeterMis, line 581: Is this function reachable?
the previous preg_match we have there likely fails on any non letter characters already, so it would fallback to the LIKE search and never touch the new sanitizeBooleanModeTerm() function ?

Unless this is meant to be just defense in depth in cases where the preg_match returns true?

the term length change makes sense, assuming the mysql setting will be updated ( before the token size is updated, it might create worse matches, but that's probably OK, since we intend to change the size to 3 )

PeterMis (commented):
Just a comment below, if you could please confirm whether that's expected or not

PR #3738, opened 2026-07-23
Fulltext indexes changes

PeterMis:
fixes xibosignageltd/xibo-private#1760


maurofmferrao (changes requested):
All looks good, there are just some Typescript errors that are preventing the App to build

maurofmferrao (dismissed):

lib/Factory/DataSetColumnFactory.php
```diff
@@ -139,14 +139,13 @@ public function query($sortOrder = null, $filterBy = [])
...
 
-        if ($sanitizedFilter->getString('keyword') != null) {
-            // Fulltext search
-            $body .= $this->buildSearchQuery(
-                $sanitizedFilter->getString('keyword'),
-                $params,
-                ['datasetcolumn.heading', 'datasetcolumn.listcontent', 'datasetcolumn.tooltip'],
-                ['datasetcolumn.dataSetColumnId']
-            );
+        if ($sanitizedFilter->getString('heading') != null) {
```
dasgarner, line 142: Does it make sense that filtering by heading also returns matches from list content and tooltip? I can see why you've done this (to match the removed full text index), but it makes the API weird?

PeterMis: yes, it was meant to "replace" the fulltext index, that said yeah perhaps it is a bit odd API wise - happy to make it filter only the heading - that's done in the commit below

dasgarner (approved):

PeterMis:
TEST: /Schedule

PR #3754, opened 2026-07-28
v5: When a Resolution is not Enabled, it still shows in Templates option

maurofmferrao:
relates to xibosignageltd/xibo-private#1358


PeterMis (approved):

maurofmferrao:
TEST: /Library

PR #3759, opened 2026-07-29
Manual guides urls; layout import gate; cleanup old setting.

PeterMis:


maurofmferrao (approved):
LGTM

views/non-authed.twig
```diff
@@ -68,7 +68,7 @@
     <body>
         <div class="container">
             <div class="{% block contentClass %}form-signin text-center{% endblock %}">
-                {% block header %}<p><a href="{{ theme.getThemeConfig("theme_url") }}"><img class="login-logo" src="{{ theme.rootUri() }}brand/{{ brandLogoFile }}" alt="Logo"></a></p>{% endblock %}
+                {% block header %}<p><a href="{{ theme.getThemeConfig("theme_url") }}"><img class="login-logo" src="/brand/{{ brandLogoFile }}" alt="Logo"></a></p>{% endblock %}
```
dasgarner, line 71: This won't work in an alias, will it?

PeterMis: this is specifically changed to work on an alias CMS
as it stands it will not work and needs this change to work

brand is served by Apache alias if it has any subfolder /cms/ etc it wil not work ( hence the change to match other images )

dasgarner (approved):

PeterMis:
TEST: /Displays

PR #3760, opened 2026-07-29
Fix i18n extraction warnings

maurofmferrao:
- frontend/i18next.config.ts: set extract.keySeparator to false to stop dotted keys being treated as nested paths
- lib/Controller/{DisplayGroup,ScheduleReport,SavedReport,Library,Display}.php, lib/Factory/PlaylistFactory.php: join wrapped __() (and one log message) strings onto single lines


PeterMis (approved):

maurofmferrao:
TEST: /Displays

PR #3761, opened 2026-07-29
Fix remaining i18n extraction issues

maurofmferrao:
- frontend/i18next.config.ts: set extract.nsSeparator to false so colon-containing strings aren't split into separate namespace files
- bin/locale.php: remove dead Note.json read, now redundant since translation.json contains everything
- frontend/src/pages/Library/Dataset/subPages/Data/DatasetDataConfig.tsx: drop angle brackets from the HTML-content placeholder label so it doesn't read as markup


PeterMis (approved):

maurofmferrao:
TEST: /Displays

PR #3769, opened 2026-07-30
Display page: Highlight Bandwidth Limit error and fix negative number

maurofmferrao:
relates to xibosignageltd/xibo-private#1471

- Prevent negative bandwidth limit values
- Highlight the Bandwidth Limit field on validation error
- Add client-side validation to Set Bandwidth modal
- Cap bandwidth limit at the DB column's max
- Fix DisplayGroupFactory::setBandwidth() TypeError

### DIFF
--- a/lib/Factory/DisplayGroupFactory.php
+++ b/lib/Factory/DisplayGroupFactory.php
@@ -275,9 +275,8 @@ class DisplayGroupFactory extends BaseFactory
      * Set Bandwidth limit
      * @param int $bandwidthLimit
      * @param array $displayGroupIds
-     * @return DisplayGroup[]
      */
-    public function setBandwidth(int $bandwidthLimit, array $displayGroupIds): array
+    public function setBandwidth(int $bandwidthLimit, array $displayGroupIds): void
     {
         $sql = 'UPDATE `displaygroup` SET bandwidthLimit = :bandwidthLimit WHERE displayGroupId IN (0';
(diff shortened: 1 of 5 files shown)


lib/Factory/DisplayGroupFactory.php
```diff
@@ -291,6 +291,8 @@ public function setBandwidth(int $bandwidthLimit, array $displayGroupIds): array
         $sql .= ')';
 
         $this->getStore()->update($sql, $params);
+
+        return $this->query(null, ['disableUserCheck' => 1, 'displayGroupIds' => $displayGroupIds]);
```
PeterMis: Did we change that because the phpdoc/return on this function suggest it should return DisplayGroup[] / array ?

If that was the only reason, then I think we might want to change the return to be void ( after `:` and in the phpdoc above ) and remove the return statement added here

Reason being, that this is not actually used anywhere, there is one caller for it in Display that does not seem to expect any return values from this function ( and it returns 204, so the request also is not expecting any returns )

And with this change we would make additional not needed query each time this function is called

PeterMis (changes requested):
Frontend looks great, one thing to consider below for the backend please

PeterMis (approved):

mgbaybay (approved):

maurofmferrao:
TEST: all

flashdave:
TEST: all

PR #3772, opened 2026-07-30
QA: Vitest reliability improvements

ifarzana:
- tightening up how mocks are set up and reused
- created tests to cover end-to-end scenarios for creating, editing, and deleting events.


frontend/src/pages/Schedule/Schedule/__tests__/table/scenarios/createEditDelete.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+  });
+
+  test('a user creates a layout event, renames it, then deletes it — each step reflects in the table', async () => {
+    const user = userEvent.setup();
+
+    await act(async () => {
+      renderEventsPage();
+    });
+    await screen.findByRole('columnheader', { name: 'Name' });
+    expect(screen.queryByRole('cell', { name: 'Welcome Layout' })).not.toBeInTheDocument();
```
flashdave, line 224: Must fix: the delete assertion in `createEditDelete.test.tsx` isn't synchronised

The column header is present before the delete, so `findByRole` resolves on the first tick and gives you no synchronisation at all , the absence assertion then races the invalidation and refetch.

To confirm rather than guess, I added 50ms of latency to the test's own fake `fetchEvent` (a properly synchronised assertion is immune to that). It failed **2/2, deterministically**, at line 266:

```
Error: expect(element).not.toBeInTheDocument()
 ❯ .../createEditDelete.test.tsx:266:11
```

This fix passes 2/2 with the same latency still applied:

```js
import { screen, within, act, waitForElementToBeRemoved } from '@testing-library/react';
...
await waitForElementToBeRemoved(() =>
  screen.queryByRole('cell', { name: 'Welcome Layout (renamed)' }),
);
```

One related observation, flagged honestly because it might just be my machine: that same test timed out twice under the broad 38-file parallel run, though it passes 3/3 in isolation and passes within the full 40-file Schedule suite. It's the heaviest test we have (~3s body against a 15s timeout). Worth watching on CI — the fix above should help, and splitting the journey is an option if it keeps flaking.

flashdave (changes requested):

frontend/src/pages/Schedule/Schedule/__tests__/table/scenarios/createEditDelete.test.tsx
ifarzana: Thanks, in this test, the delete finishes almost instantly, so the item is often already gone by the time we start checking,, now the test properly waits for the delete to finish before deciding pass or fail, no matter how fast or slow that takes.

flashdave (approved):
Looks good to me

ifarzana:
TEST: all

PR #3776, opened 2026-07-31
Sessions: updated session logout redirect

mgbaybay:
### Frontend Changes
- Fixed the issue where the user is not actually logged out of the session
- Updated confirmation dialog to match the one in 4.4
- Redirect own user session expiry to login page to match the one in 4.4

### Backend Changes
- Updated session expiry handling

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1699

### DIFF
--- a/frontend/src/pages/Advanced/Sessions/Sessions.tsx
+++ b/frontend/src/pages/Advanced/Sessions/Sessions.tsx
@@ -40,17 +40,22 @@ import FilterButton from '@/components/ui/FilterButton';
 import FilterInputs from '@/components/ui/FilterInputs';
 import TabNav from '@/components/ui/TabNav';
 import { DataTable } from '@/components/ui/table/DataTable';
-import { useUserContext } from '@/context/UserContext';
+import { withPublicPath } from '@/config/publicPath';
 import { useDateFormatter } from '@/hooks/useDateFormatter';
 import { useFilteredTabs } from '@/hooks/useFilteredTabs';
 import { useTableState } from '@/hooks/useTableState';
 import type { Session } from '@/types/session';
 import { countActiveFilters } from '@/utils/filters';
 
+// Matches the legacy design: your own current session logs out via the real /logout
+// route directly, not the admin-facing bulk endpoint.
+const redirectToLogout = () => {
+  window.location.href = withPublicPath('logout');
+};
+
 export default function Sessions() {
   const { t } = useTranslation();
   const { formatDateTime } = useDateFormatter();
-  const { user } = useUserContext();
   const queryClient = useQueryClient();
 
   const {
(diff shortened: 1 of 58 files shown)


frontend/src/pages/Advanced/Sessions/Sessions.tsx
```diff
@@ -150,8 +153,15 @@ export default function Sessions() {
       return;
     }
 
-    setSessionToLogout([session]);
     setLogoutError(null);
+
+    // Logging out your own session needs no confirmation
+    if (id === user?.userId) {
```
maurofmferrao: Since this bypasses `openModal('logout')`, `logoutError` never gets rendered anywhere for the self-logout path,  `LogoutSessionModal` (the only consumer of `logoutError`) only mounts when `activeModal === 'logout'`.

If `logoutSession` throws here (expired token, network blip), `useSessionActions` sets the error state but the UI shows nothing; the user just sees the button do nothing. Worth surfacing the failure some other way (e.g. still open the modal on failure, or a toast) rather than only on the happy path.

maurofmferrao (changes requested):
LGTM with warnings

lib/Controller/Sessions.php
```diff
@@ -221,8 +232,12 @@ public function logout(Request $request, Response $response, $id)
             throw new AccessDeniedException();
         }
 
-        // We log out all of this user's sessions.
-        $this->sessionFactory->expireByUserId($id);
+        if ((int) $id === $this->getUser()->userId) {
+            $this->completeLogoutFlow($this->getUser(), $this->session, $this->getLog(), $request);
```
PeterMis: I'm not sure if that's quite correct

`$this->session` would be bound to the current user session issuing the request not the session that was selected in the grid, I think?
 ( which may happen to be the same, if there is only one, but that's not ideal for multi sessions which we are trying to solve here )

So that means an admin would have no way to logout their own other session, it would always log them out of their current session - which I presume is not what we wanted to achieve? I feel like that would be confusing.

lib/Controller/Sessions.php
PeterMis: Also minor note about the OA on this logout function
currently the description reads `Logs out all sessions for the given user`
which would not be quite true for the self logout case only for someone else case

Should we adjust that?

PeterMis (changes requested):
Looks good, just two comments below to address please

PeterMis (dismissed):

PeterMis:
We will need to address the tests that prevent the container from being built please

maurofmferrao (changes requested):
`createMockSession` in `frontend/src/pages/Advanced/Sessions/__tests__/Setup.tsx` needs the `isCurrentSession` property, otherwise the build fails.

```
export const createMockSession = (overrides?: Partial<Session>): Session => ({
  userId: Math.floor(Math.random() * 1000),
  userName: 'john_doe',
  remoteAddress: '192.168.1.1',
  userAgent: 'Chrome 120',
  isExpired: false,
  isCurrentSession: false,
  lastAccessed: '2024-01-01T10:00:00Z',
  expiresAt: '2024-01-02T10:00:00Z',
  ...overrides,
});
```

Other than that, LGTM

PeterMis (approved):

maurofmferrao (approved):
LGTM

ifarzana:
TEST: /Advanced/Sessions

PR #3777, opened 2026-07-31
Widget: Cache fixes and adjustments.

### LINKED ISSUE #3876: Weather widget: displays showing different weather data
CMS: V4.4.0

When you have multiple screens showing a weather widget (different layouts) set to the same location, they do not always update at the same time. One screen will show the latest weather while the other screen continues showing old data.

## Prerequisites
- Two layouts, each containing a weather widget configured with the same location (same coordinates, same units, "Use display location" turned off) - one layout per display

## Steps

1. Create two layouts, each with a weather widget set to the same location.
2. Assign each layout to a different display and leave them running.
3. Wait an hour for the weather data to refresh.
4. One display will show the updated weather. The other will still be showing the old data.

## Expected

Both displays update at the same time since they are showing weather for the same location.

## Actual

Only one display updates. The other can stay behind by 2 or more hours. The issue is that the XMR push is only sent to one display.

## Notes

If anyone opens the layout preview in the CMS while the weather is due to refresh, neither display will update. The preview silently resets the refresh timer without telling any display that new data is available.

PeterMis:
fixes xibosignage/xibo#3876

### DIFF
--- a/lib/Controller/Widget.php
+++ b/lib/Controller/Widget.php
@@ -1205,90 +1205,108 @@ class Widget extends Base
         ) {
             $this->getLog()->debug('getData: Pulling fresh data');
 
-            $dataProvider->clearData();
-            $dataProvider->clearMeta();
-            $dataProvider->addOrUpdateMeta('showFallback', $showFallback);
-
             try {
-                if ($widgetInterface !== null) {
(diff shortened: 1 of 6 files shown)


dasgarner (approved):

PeterMis:
TEST: all

PR #3783, opened 2026-08-03
Display: Fix register with invalid default folderId.

### LINKED ISSUE #3890: Display Registration: fails and outputs raw error response when DISPLAY_DEFAULT_FOLDER is empty/invalid
**Description:** 
If the `DISPLAY_DEFAULT_FOLDER` setting holds a value that is not a number or an empty string, then XMDS `RegisterDisplay` throws an uncaught `TypeError` for every new Display, and PHP's SOAP layer returns that error's raw text to the player.

PeterMis:
fixes xibosignage/xibo#3890

Not many lines changed, but couple of fixes here:
- By default `DISPLAY_DEFAULT_FOLDER` is not in settings table, make it so on first save it's set to 1 (root folder) not an empty string
- Protect the `DISPLAY_DEFAULT_FOLDER` getSetting() in register wrap it in intval ( would fix empty string here, cast to 0, then on save it would get folderId 1 )
- Wrapped display save() in try/catch to ensure we do not return raw response to the player.

### DIFF
--- a/lib/Controller/Settings.php
+++ b/lib/Controller/Settings.php
@@ -640,12 +640,12 @@ class Settings extends Base
         if ($this->getConfig()->isSettingEditable('DISPLAY_DEFAULT_FOLDER')) {
             $this->handleChangedSettings(
                 'DISPLAY_DEFAULT_FOLDER',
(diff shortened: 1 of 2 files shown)


dasgarner (approved):

flashdave:
TEST: all

PR #3788, opened 2026-08-04
Reports: fixed missing time filter

mgbaybay:
### Frontend Changes
- Added the missing time filter for the reports

### Backend Changes
- Added spacing in the SQL to fix the grouping error

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1770


frontend/src/components/ui/DateRangeFilter.tsx
```diff
@@ -167,10 +213,18 @@ export default function DateRangeFilter({
                   mode="range"
                   disableFutureDates
                   isJalali={isJalali}
+                  showTimePicker={showTimePicker}
+                  value={parseRangeValue(value, timeZone)}
                   onCancel={() => setOpenDatePicker(false)}
                   onApply={(v) => {
                     if (v.type === 'range') {
-                      onChange(name, `range:${toLocalDateKey(v.from)}|${toLocalDateKey(v.to)}`);
+                      const fromKey = toLocalDateKey(v.from);
```
maurofmferrao: `formatTime` reads the hour/minute in the CMS timezone, but `toLocalDateKey` reads the day in the browser's timezone. If those two differ, the saved date can be off by a day. Both should use the same timezone.

maurofmferrao (changes requested):
The date and time in the range filter can come from different timezones, so the saved date can be off by a day when the CMS timezone differs from the browser's. Everything else looks good.

maurofmferrao (dismissed):

frontend/src/components/ui/DateRangeFilter.tsx
```diff
@@ -156,25 +202,38 @@ export default function DateRangeFilter({
...
+                    value={parseRangeValue(value, timeZone)}
+                    onCancel={() => setOpenDatePicker(false)}
+                    onApply={(v) => {
+                      if (v.type === 'range') {
+                        const fromKey = toLocalDateKey(
+                          v.from,
+                          showTimePicker ? timeZone : undefined,
+                        );
+                        const toKey = toLocalDateKey(v.to, showTimePicker ? timeZone : undefined);
+                        const rangeValue = showTimePicker
```
maurofmferrao, line 227: Flagged by Claude:

`timeZone` here is the user's personal `defaultTimezone` setting, so `statsFromDtTime`/`statsToDtTime` are sent as `HH:mm` in the user's own zone. But `lib/Report/ProofOfPlay.php:382-390` applies that `HH:mm` onto a Carbon date anchored to the CMS's global timezone (set via `date_default_timezone_set` in `lib/Middleware/State.php:211`), not the user's personal one. If those two differ, the report window shifts by the offset between them.

This was dead backend code until this PR, so it's a new user-facing bug, not a pre-existing one. I don't think it's fixable purely on the frontend, worth flagging to whoever owns the backend side.

maurofmferrao (changes requested):
There might be a bug that needs to be fixed, in a comment I left from a finding made by Claude.

maurofmferrao (approved):
LGTM

flashdave:
TEST: all

PR #3793, opened 2026-08-05
Translations: Several navigation labels and data fields won't translate into other languages

maurofmferrao:
relates to xibosignageltd/xibo-private#1798


mgbaybay (approved):

ifarzana:
TEST: all

PR #3770, opened 2026-07-30
Bulk move to folder fixes

maurofmferrao:
relates to xibosignageltd/xibo-private#1650

- Fix inconsistent Move icon on Display Groups (Folder → FolderInput)
- Fix bulk move across multiple pages sending concurrent requests, causing DB deadlocks
- Fix deadlock retries being swallowed, causing silent partial failures
- Fix Displays bulk move missing row selection reset

PR #3795, opened 2026-08-05
QA: Add hook-level regression tests for bulk-move-to-folder

ifarzana:


flashdave (changes requested):
The `maxInFlight` approach here is the right call and I verified it works — but it's currently wired up to only seven of the nine pages the bug affects, and I can show the gap is silent.

**First, the harness genuinely catches the bug.** I reverted #3770's fix in the Media hook and re-ran that file:

```
× sends move requests sequentially, not concurrently
  AssertionError: expected 3 to be 1
Tests  1 failed | 4 passed (5)
```

Four of the five tests passed with the bug live. Call count, both notification paths, the empty guard — all identical whether requests go out serially or all at once. `tracker.maxInFlight` is the only observable difference, which is exactly why this test needed to exist.

**The problem: two pages have no test.**

```
src/pages/Library/Dataset/hooks/useDatasetActions.ts
src/pages/Library/MenuBoard/hooks/useMenuBoardActions.ts
```

Both were broken by the original concurrency bug, both were fixed in #3770, neither is covered here.

To check whether that actually matters, I reintroduced the identical bug in both and ran every existing test those two pages have:

```
Test Files  22 passed (22)
Tests  267 passed (267)
```

267 tests, none of them noticed. So if someone reintroduces `Promise.all` on Datasets or Menu Boards later, the suite stays green and the deadlocks come back. Those two pages have the same protection today as before this PR — none.

Given the harness already exists, each one should be close to a copy of an existing file.

**Second, worth a decision on the duplication.** 1,219 added lines, 254 unique — the 20-line notify-mock block is byte-identical in all seven files, and the rest differs only in hook name, fixture, id field and `targetType`. A shared `testUtils/notifyMock.ts` alone drops ~140 lines. I don't feel strongly about it — explicit test files are a legitimate style — but adding Dataset and MenuBoard makes it nine copies, so it's worth deciding rather than defaulting.

Happy to approve once Dataset and MenuBoard are in; the duplication is your call.

ifarzana:
Thank you! I've added the missing files, and extracted only the shared notification mock, while leaving the individual concurrency tests as they are.

flashdave (approved):
LGTM!

ifarzana:
TEST: all

PR #3797, opened 2026-08-05
Fix: Layout Code dropdown only showing first 10 results

maurofmferrao:
relates to xibosignage/xibo#3888

- fetchLayoutCodes() never sent start/length or read x-total-count, so the backend silently capped results at 10 with no search/load-more
- Resolution and Sync Group per-display Layout dropdowns had the same cap-with-no-pagination bug (capped at 100), fixed the same way


chaolen (approved):
LGTM!

ifarzana (approved):
Thank you for adding the mock. Approved.

Kent0425:
TEST: all

Kent0425:
Test failed but still safe to merge @maurofmferrao

PR #3802, opened 2026-08-06
Fix: Handle API 500 errors inline

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1799

Changes:
  - Set `throwOnError: false` globally in QueryClient so API 500 errors are stored in query state (isError) rather than thrown during render — pages can now handle them inline with error banners instead of hitting the full-screen ErrorPage
  - Removed redundant `throwOnError` callbacks from 35 individual data hooks (the global default now covers all of them)
  - Added inline error banners to StatusDashboard and PlaylistDashboard, which were the only pages missing error state handling
  - Fixed `getUserSession()` in route loaders to re-throw 500+ server errors so they reach the `errorElement`, instead of treating them as "not authenticated" and redirecting to login


frontend/src/routes/loaders.ts
```diff
@@ -57,7 +58,11 @@ async function getUserSession(): Promise<User | null> {
       if (isValidUser(res.data)) {
         return res.data;
       }
-    } catch {
+    } catch (err) {
+      // Server error — let it bubble to errorElement
```
maurofmferrao: This doesn't actually bubble anywhere, there's no `errorElement` configured on any route in `routes/index.tsx` (I checked with `grep -rn "errorElement" frontend/src`, the only hit is this comment). This throw becomes an uncaught loader error rendered by react-router's default unstyled error page, which is exactly the full-screen-crash experience this PR is trying to eliminate elsewhere, and it happens on the very first 500 with zero retries, unlike the 401/network path just below it. I'd drop this special case and let 500s fall through to the existing retry-then-null path, unless there's a wired-up `errorElement` I'm missing.

maurofmferrao (changes requested):
Do not merge, 2 blockers

- `routes/loaders.ts` re-throws 500s from `/user/me` straight into an uncaught crash with no `errorElement` to catch it. Branch `bugfix/update_error_element` already has the fix, so this needs that branch merged first.

- 12 changed files have import/order lint warnings. Run `eslint --fix` can fix this automatically.

maurofmferrao (dismissed):

flashdave:
TEST: all

maurofmferrao (approved):

flashdave:
TEST: all

PR #3804, opened 2026-08-06
Tags: fixed clicking a tag no longer filters the grid

mgbaybay:
### Frontend Changes
- Added tag filtering functionality
- UI change: updated scrolling on the tags filter input field
- Fixed minimum media item in a playlist

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1753


maurofmferrao (changes requested):
The code and logic looks fine, but when testing it I notice that if select a tag with value (e.g. `animal|dog` ) it adds that to the filter, and all the rows with the tagId will be selected ( `animal|dog`, `animal|cat`, `animal|sheep`, etc ).

We either only add the tagId to the filter, with no value, or we need to make sure the tag filter also filters by value.

maurofmferrao (dismissed):

flashdave:
TEST: all

maurofmferrao (approved):

mgbaybay:
TEST: all

PR #3810, opened 2026-08-07
Displays: fixed bulk edit functionality

mgbaybay:
### Frontend Changes
- Added error handling for bulk tag add/edit/delete updates
- Added notification for bulk tag updates

### Backend Changes
- Added try/catch for the bulk edit

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1789

### DIFF
--- a/frontend/src/components/ui/modals/EditTagsMultipleModal.tsx
+++ b/frontend/src/components/ui/modals/EditTagsMultipleModal.tsx
@@ -104,7 +104,6 @@ export default function EditTagsMultipleModal({
       }
 
       await onSuccess?.();
-      onClose();
     } catch (err) {
       console.error(err);
       setError(
--- a/frontend/src/components/ui/modals/EditTagsMultipleModal.tsx
+++ b/frontend/src/components/ui/modals/EditTagsMultipleModal.tsx
@@ -19,6 +19,7 @@
  * along with Xibo.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { isAxiosError } from 'axios';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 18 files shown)


frontend/src/services/tagApi.ts
```diff
@@ -106,16 +106,28 @@ export interface EditMultipleTagsPayload {
...
   params.append('targetIds', payload.ids.join(','));
   params.append('addTags', payload.addTags ?? '');
   params.append('removeTags', payload.removeTags ?? '');
 
-  await http.put(`/tag/${payload.targetType}/multi`, params.toString(), {
+  const response = await http.put(`/tag/${payload.targetType}/multi`, params.toString(), {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
   });
+
+  return {
```
maurofmferrao, line 127: **Blocker**: the payload is nested under `response.data.data` (same envelope every other API file here uses), not `response.data`. `failedCount` is always 0, so the partial-failure toast never fires. 

Suggested fix:
```tsx
    return {
      failedCount: response.data?.data?.failedCount ?? 0,
      failedNames: response.data?.data?.failedNames ?? [],
    };
```

frontend/src/components/ui/modals/EditTagsMultipleModal.tsx
```diff
@@ -83,19 +84,34 @@ export default function EditTagsMultipleModal({
...
+            count: result.failedCount,
+            total: ids.length,
+          }),
+          { description: result.failedNames.join(', ') },
+        );
+      } else {
+        notify.success(t('Tags updated for {{count}} selected items.', { count: ids.length }));
+      }
+
+      await onSuccess?.();
```
maurofmferrao, line 106: Every caller's `onSuccess` already calls `closeModal()`, so this closes twice.

Can we drop this `onClose()`?

maurofmferrao (changes requested):
1 Blocker and 1 Suggestion to fix before we merge, everything else looks good to me.

PeterMis (approved):

maurofmferrao (approved):

flashdave:
TEST: all

PR #3813, opened 2026-08-07
Media: fixed expiry column text

mgbaybay:
### Frontend Changes
- Updated expiry column to use the new `expiresFormatted` from the backend

### Backend Changes
- Added `expiresFormatted` 
- Removed unused properties

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1775


lib/Controller/Library.php
```diff
@@ -2665,9 +2666,20 @@ private function decorateMediaProperties($request, $parsedQueryParams, $media):
...
-        $media->setUnmatchedProperty('mediaExpiresIn', __('Expires %s'));
-        $media->setUnmatchedProperty('mediaExpiryFailed', __('Expired '));
-        $media->setUnmatchedProperty('mediaNoExpiryDate', __('Never'));
+        $expiresTimestamp = $media->expires;
+        $now = Carbon::now();
+
+        if ($expiresTimestamp == 0) {
+            $expiresFormatted = __('Never');
+        } elseif ($expiresTimestamp > $now->timestamp) {
+            $relative = Carbon::createFromTimestamp($expiresTimestamp)
```
PeterMis: Nothing wrong here as it stands, but we will need to keep in mind that the `Carbon::createFromTimestamp()` will be handled differently once the composer updates PR lands - at which points this will need to be updated

( unless that PR with composer updates lands first, then we should update it here in this PR )

PeterMis (dismissed):
LGTM, one comment below - might need a follow up fix or fix here depending on the timing with the other PR merge

frontend/src/pages/Library/Media/MediaConfig.tsx
```diff
@@ -622,14 +622,10 @@ export const getMediaColumns = (props: MediaActionsProps): ColumnDef<Media>[] =>
       },
     },
     {
-      accessorKey: 'expires',
+      accessorKey: 'expiresFormatted',
```
maurofmferrao, line 625: This rename changes the column's id too. `Media.tsx` line 113 still sets the default column visibility with the old key, `expires: false`, so it no longer matches and the column will default to visible. Can you update it to `expiresFormatted: false`.

maurofmferrao (changes requested):
Just a single note about the new column id not being addressed in the Media.tsx file.

maurofmferrao (dismissed):

PeterMis (dismissed):

PeterMis (approved):

maurofmferrao (approved):

flashdave:
TEST: all

PR #3818, opened 2026-08-07
User Preferences : Fix settings that had no effect on the CMS

maurofmferrao:
relates to xibosignage/xibo#3898

- Fixed "clear all auto submit form choices" needing a reload to take effect
- Removed isAlwaysUseManualAddUserForm and navigationMenuPosition


PeterMis (dismissed):

mgbaybay (dismissed):

PeterMis (dismissed):

PeterMis (dismissed):

flashdave:
TEST: all

flashdave (changes requested):
**One file was missed.** `DisplayGroup.edit.test.tsx` → *"opens the Edit modal when the Edit action is
clicked on a row"* still fails with the same `Unable to find role="button" and name /edit/i`. Its row
action is gated on `isHydrated` just like the Filters button, so it needs the same treatment
`openFilters()` gives.

The bug the tests were flagging is still there.** `setIsFolderIdHydrated(true)`
(`useTableState.ts:83`) is only reachable from the promise's fulfilled branch, and `isHydrated` is now
`isHydrated && isFolderIdHydrated` (`:216`). Fail that single `/json/user/pref` request — 500, expired
session, blip — and all 48 pages using the hook sit in a permanent skeleton: no retry, no error shown, and
on Media the grid query never fires at all. The sibling React Query call retries 3×; this one is attempted
once.

flashdave (dismissed):

maurofmferrao:
TEST: all

flashdave (approved):
LGTM!

maurofmferrao:
TEST: all

maurofmferrao:
TEST: all

PR #3820, opened 2026-08-07
Logs : Reports Client-Side Paging Implementation & CSV Export

maurofmferrao:
relates to xibosignage/xibo#3896

- Replaced the separate COUNT(*) query with SQL_CALC_FOUND_ROWS/FOUND_ROWS() so totals can't race with concurrent writes and still work when the page returns zero rows
- Default "Exclude Log" filter to on

### DIFF
--- a/lib/Factory/LogFactory.php
+++ b/lib/Factory/LogFactory.php
@@ -74,8 +74,12 @@ class LogFactory extends BaseFactory
         $params = [];
         $limit = '';
 
+        $isPaged = $filterBy !== null
+            && $parsedFilter->getInt('start') !== null
+            && $parsedFilter->getInt('length', ['default' => 10]) !== null;
+
         $select = '
-            SELECT `logId`,
+            SELECT ' . ($isPaged ? 'SQL_CALC_FOUND_ROWS ' : '') . '`logId`,
                 `runNo`,
                 `logDate`,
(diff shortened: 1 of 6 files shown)


lib/Factory/LogFactory.php
```diff
@@ -74,8 +74,12 @@ public function query(?array $sortOrder = null, array $filterBy = []): array
         $params = [];
         $limit = '';
 
+        $isPaged = $filterBy !== null
+            && $parsedFilter->getInt('start') !== null
+            && $parsedFilter->getInt('length', ['default' => 10]) !== null;
+
         $select = '
-            SELECT `logId`,
+            SELECT ' . ($isPaged ? 'SQL_CALC_FOUND_ROWS ' : '') . '`logId`,
```
PeterMis: I'm not too sure about that one

It looks like it's deprecated and expected to be removed in the future mysql versions, so we may want to avoid using that here?

Separate query with `count(*)` is what mysql docs recommend

PeterMis (changes requested):
Looks good overall, we likely should address the comment below though

mgbaybay (dismissed):

PeterMis (approved):

flashdave:
TEST: all

PR #3824, opened 2026-08-10
Schedule: fixed invalid api route error handling

mgbaybay:
### Backend Changes
- Added the `:[0-9]+` regex constraint to every numeric ID route placeholder in `lib/routes.php` that was missing it (matching the pattern already used by routes like  `/schedule/{id:[0-9]+}`). Non-numeric values now fail to match the route at all, so Slim's router returns a proper 404 instead of the request reaching the controller and erroring out.
- Updated handling for missing date parameters

-------
Relates to https://github.com/xibosignage/xibo/issues/3894


PeterMis (changes requested):
I think those changes make sense, but I'm unsure if it actually fully addresses the reported issue.

I feel like that 500 `unexpected error`  on `/api/schedule/{id}/events` was more so due to `date` or `startDate` and `endDate` missing rather than the displayGroupId being a string

As it stands the eventList(), reads to me like either
`singlePointInTime` and `date` 
OR
`startDate` and `endDate`

Have to be provided, otherwise the unconditional `->second(0)` will throw an error - all of those are marked as optional in the OA

Should this default to a sensible date if none of the above is provided?
Or should it be required to provide date? ( so that at least one of the branches can go through )

Perhaps the `$id` should be type hinted to `int` in the eventList() function as well? Just so it's fully clear what's expected.

Could we re-investigate this please?

PeterMis (approved):

mgbaybay:
TEST: all

PeterMis (approved):

mgbaybay:
TEST: all

PR #3830, opened 2026-08-11
QA: Fix Vitest suite

ifarzana:
Fix stale assertions in Media/AuditTrail vitest suites, remove flaky Layouts dropdown test


flashdave (commented):
One thing I want to flag before approving: the diff deletes all of Layouts.column.test.tsx, which is 10 tests, not just the flaky dropdown one. Five of them don't touch a dropdown at all, and four of those are the only layout.add/layout.modify coverage we have anywhere — it also leaves mockDisplayManagerUser orphaned in layoutTestUtils.

Was there a reason for taking the whole file rather than just the flaky test? Entirely possible you hit something I haven't, if the rest of the file was failing too, or the permission tests were unreliable for a reason that isn't obvious from the diff, that changes the answer and I'd rather know before pushing back.

If not, there's an easier route: 11 files in the repo already mock DataTableRowActions to swap the dropdown for plain buttons, including Layouts.delete.modal.test.tsx right next door. That'd make the flaky ones deterministic and keep all ten.

ifarzana:
The layout column tests will need to be revisited, so I’ve deleted them for now.

flashdave (dismissed):
Looks good

ifarzana:
TEST: all

maurofmferrao (dismissed):
LGTM

ifarzana:
TEST: all

flashdave (dismissed):
LGTM

maurofmferrao (approved):

ifarzana:
TEST: all

PR #3832, opened 2026-08-11
Schedule grid returns 500 for monthly recurring events that repeat on a weekday-of-month

maurofmferrao:
relates to xibosignageltd/xibo-private#1807


PeterMis (approved):

ifarzana:
TEST: all

PR #3866, opened 2026-08-20
Fix: Menu Board create throws HTTP 500 during audit logging

maurofmferrao:
relates to xibosignageltd/xibo-private#1823

- MenuBoard::add() bound modifiedDt only as a SQL param, never assigning it back onto the entity, so the in-memory value stayed null after insert

- save()'s audit-on-create call falls back to toArray() when there are no originalValues, which unconditionally formatted datesToFormat fields, crashing Carbon::createFromTimestamp() on the null value

- Also guard EntityTrait::toArray() against empty date values, matching the check already in getChangedProperties(), to close the same failure mode for any other entity

### DIFF
--- a/lib/Entity/EntityTrait.php
+++ b/lib/Entity/EntityTrait.php
@@ -307,7 +307,7 @@ trait EntityTrait
         $objectAsJson = $this->jsonSerialize();
 
         foreach ($objectAsJson as $key => $value) {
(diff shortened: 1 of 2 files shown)


PeterMis (approved):

Kent0425:
TEST: all

Kent0425:
Tested!

PR #3868, opened 2026-08-20
fix: new trustedProxyIps setting. xibosignageltd/xibo-private#1822

dasgarner:
Fixes: https://github.com/xibosignageltd/xibo-private/issues/1822


lib/Middleware/TrustedProxyIpAddress.php
```diff
@@ -0,0 +1,135 @@
...
+ * and each successive hop peeled off a multi-hop chain, is confirmed to be on the trusted list (see
+ * IpTrust::isTrusted() — exact IP, IPv4 CIDR, and wildcard entries all supported).
+ *
+ * Replaces the third-party RKA\Middleware\IpAddress package (removed so this and the other
+ * trusted-proxy consumers — HttpsDetect::isShouldIssueSts(), Session::getIp(), Xmds\Soap::getIp() —
+ * share one matcher instead of two, keeping CIDR/wildcard behaviour consistent everywhere).
+ */
+class TrustedProxyIpAddress implements Middleware
+{
+    private const HEADERS = ['Forwarded', 'X-Forwarded-For', 'X-Forwarded', 'X-Cluster-Client-Ip', 'Client-Ip'];
```
PeterMis: resolve() checks these 5 headers in fixed priority, with no check that the specific header it picks is the one the operator's real proxy actually sets.
e.g.
If the proxy only sets `X-Forwarder-For` but never touches `Forwarded`, it is possible to send `Forwarded: for=<anything>` header which wins priority and gets treated as trusted.

lib/Service/ConfigService.php
```diff
@@ -205,6 +206,25 @@ public function getWhitelistHosts(): string
...
+     */
+    public function getTrustedProxyIps(): string
+    {
+        return $this->trustedProxyIps;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getTrustedProxyIpList(): array
```
PeterMis, line 220: The new deployment-only `$trustedProxyIps` is unioned with the pre-existing `WHITELIST_LOAD_BALANCERS` db setting, which is editable in CMS settings - this seem to directly contradict updated SECURITY.md rationale for keeping the trust list our of the admin UI.

lib/Middleware/TrustedProxyIpAddress.php
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        return $remoteAddr;
+    }
+
+    /**
+     * Strip a trailing :port from an IPv4 address, and unwrap a bracketed IPv6 address
+     * (`[::1]` or `[::1]:8080`) — IPv6 addresses without brackets are left untouched.
+     */
+    private function extractIp(string $ip): string
```
PeterMis, line 102: do we cover stripping quotes from a quoted `Forwarded: for="1.2.3.4"` here? ( value no port case )

lib/Helper/Session.php
```diff
@@ -557,15 +567,22 @@ private function updateSessionHistory(): void
      */
     private function getIp(): string
```
PeterMis, line 568: We seem to validate the `X-forwarded-for` as a single IP instead of chain-walking, so multi-hop trusted proxy chains fail and get logged under wrong IP.

web/index.php
```diff
@@ -93,7 +93,17 @@
...
+// Trusted reverse-proxy list for X-Forwarded-For client IP resolution (used for login/2FA/password-reset
+// rate limiting and audit logging). Combines the settings.php-only $trustedProxyIps (supports CIDR/wildcard)
+// with the admin-editable WHITELIST_LOAD_BALANCERS setting (Settings > Network), so operators who already
+// listed their load balancer there (e.g. for HSTS trust) get it trusted here too.
+// setDependencies() must be called explicitly here (mirroring web/xmds.php) because getSetting() requires
+// it, and State::setState() only calls it per-request — too late, since this list is fixed at bootstrap.
+$configService = $container->get('configService');
+$configService->setDependencies($container->get('store'), $container->get('rootUri'));
+$trustedProxyIps = $configService->getTrustedProxyIpList();
+$app->add(new \Xibo\Middleware\TrustedProxyIpAddress($trustedProxyIps));
```
PeterMis, line 106: Should this be registered for API entrypoint as well or was it explicitly not included?

lib/Helper/HttpsDetect.php
```diff
@@ -186,29 +187,63 @@ public static function isHttps(): bool
...
 
+    /**
+     * Used by isHttpsTrusted() and isShouldIssueSts(): is a forwarded "https" proto claim
+     * trustworthy for the current request — i.e. does it come via a REMOTE_ADDR on the operator's
+     * trusted-proxy list? An empty list means nothing is trusted, not "trust everything" (see
+     * $trustedProxyIps / WHITELIST_LOAD_BALANCERS in SECURITY.md).
+     * @param \Xibo\Service\ConfigServiceInterface $config
+     * @return bool
+     */
+    private static function isForwardedHttpsTrusted(ConfigServiceInterface $config): bool
```
PeterMis, line 239: HSTS issuance bypasses the empty-trust-list permissive fallback that other schema/port checks get, saliently regressing HSTS on upgrade for installs without either setting configured.

lib/Helper/IpTrust.php
```diff
@@ -0,0 +1,207 @@
...
+            return false;
+        }
+        [$subnet, $bits] = explode('/', $cidr, 2);
+        if (!is_numeric($bits)) {
+            return false;
+        }
+
+        $packedIp = @inet_pton($ip);
+        $packedSubnet = @inet_pton($subnet);
+        if ($packedIp === false || $packedSubnet === false || strlen($packedIp) !== strlen($packedSubnet)) {
```
PeterMis, line 100: IPv4-mapped-IPv6 REMOTE_ADDR (::ffff:10.0.0.5) never matches a plain IPv4 trusted entry due to differing inet_pton byte lengths.

lib/Xmds/Soap.php
```diff
@@ -2761,17 +2762,22 @@ private function isInsideOperatingHours(): bool
      */
     protected function getIp()
```
PeterMis, line 2763: Same comment as in getIp() in Session

Could we have this logic in one shared place perhaps?

PeterMis (changes requested):
Looks great overall, there were several things flagged, please see comments below

The `TrustedProxyIpAddress` resolve() and `ConfigService` issues were flagged with higher impact/priority.
Other comments perhaps not blocking, but might be worth another look

lib/Service/ConfigService.php
dasgarner: I think this is OK, the setting is admin only (in global settings) and can only be configured once logged in.

web/index.php
dasgarner: We only use it for user facing functions, so far as I know. I.e. things that can be accessed over the web portal. Sessions, redirects, etc. Not applicable to the API.

lib/Middleware/TrustedProxyIpAddress.php
dasgarner: That's OK isn't it? The user supplied value has to match what's inside the trusted proxy ips configured locally?

PeterMis (approved):

ifarzana:
TEST: all

PR #3870, opened 2026-08-20
CSV Export: fixed exported filename

mgbaybay:
### Frontend Changes
- Added table name to be used in exported CSV file name

-------
Relates to https://github.com/xibosignage/xibo/issues/3899


maurofmferrao (changes requested):
I found 2 bugs for Dataset:

**1st**: the Export from the table works, but the Export CSV from the action menu was returning a random string. In `datasetApi.ts` the line `link.setAttribute('donwload', fileName);` should be `link.setAttribute('download', fileName);` instead.

**2nd**: if the Dataset has a name like "Dataset." the Export CSV file will be "Dataset..csv". We need to sanitize the name. My suggestion it to add the method `sanitizeFileName` to `stringUtils.ts` like:

```
export function sanitizeFileName(name: string): string {
  return name
    .replace(/[/\\:*?"<>|]/g, '')
    .replace(/\s+/g, ' ')
    .trim()
    .replace(/\.+$/, '')
    .trim();
}
```

And then import it `DataTable.tsx` and `datasetApi.ts`. And then in `exportDatasetCsv` in `datasetApi.ts` we can construct the filename like this:

```
  if (contentDisposition) {
    const fileNameMatch = contentDisposition.match(/filename="?(.+)"?/);
    if (fileNameMatch && fileNameMatch.length === 2) {
      const rawFileName = fileNameMatch[1];
      const extensionMatch = rawFileName.match(/\.[^.]+$/);
      const extension = extensionMatch ? extensionMatch[0] : '';
      const base = extension ? rawFileName.slice(0, -extension.length) : rawFileName;
      const sanitizedBase = sanitizeFileName(base);

      if (sanitizedBase) {
        fileName = `${sanitizedBase}${extension}`;
      }
    }
  }
```

maurofmferrao (approved):

flashdave:
TEST: all

PR #3876, opened 2026-08-21
Permissions : Fix getPermission folder fallback check.

### LINKED ISSUE #3922: Campaign edit/delete issue for users on folder shares through user group
A user cannot edit or delete a campaign that lives inside a folder shared with them (view/edit/delete) via a user group, unless they are also the owner of that campaign. The campaign is visible in the folder, but the Edit/Delete actions do not work for it.

## Steps to Reproduce
1. Create a folder and add a campaign to it.
2. Create a user group with all campaign features enabled.
3. Create a standard user and add them to the above user group.
4. Set shares on the folder from step 1 for the user group only (view/edit/delete).
5. Log in as that user and go to the folder. The campaign is visible, but cannot be edited or deleted.

## Expected
The user should be able to edit and delete the campaign, since their user group has view/edit/delete shares on the folder containing it.

## Actual
The campaign is visible in the folder, but the Edit and Delete actions do not work for the user; they are not authorized in the UI even though the folder share grants edit/delete.

PeterMis:
fixes xibosignage/xibo#3922

### DIFF
--- a/lib/Entity/User.php
+++ b/lib/Entity/User.php
@@ -1266,9 +1266,8 @@ class User implements \JsonSerializable, UserEntityInterface
                 // Group Admin and group object in the user's array of groups
                 return $this->permissionFactory->getFullPermissions();
             }
-        } else if (
-            $this->userTypeId == 2
-            && count(array_intersect($this->groups, $this->userGroupFactory->getByUserId($object->getOwnerId())))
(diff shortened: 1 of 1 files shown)


maurofmferrao (approved):
LGTM

ifarzana:
TEST: all

PR #3880, opened 2026-08-25
Layout: Fix issues with auto publish enabled.

### LINKED ISSUE #3924: Enabling the DEFAULT_LAYOUT_AUTO_PUBLISH_CHECKB setting breaks Layout Designer.
With DEFAULT_LAYOUT_AUTO_PUBLISH_CHECKB  set to 1 in settings, we attempt to set publishedDate on a draft Layout
When adding a new Layout or editing a layout without `publishedDate` set on it this will cause an error when creating a timestamp from the `$parent->publishedDate` as it runs unconditionally with this setting enabled.

This will then cause lock issues and eventually timeout leaving user unable to publish any Layout. 

Temporary workaround in current 4.5.1 CMS version would be to set this setting to 0.

PeterMis:
fixes xibosignage/xibo#3924

### DIFF
--- a/lib/Entity/Layout.php
+++ b/lib/Entity/Layout.php
@@ -2211,28 +2211,39 @@ class Layout implements \JsonSerializable
 
             // Layout auto Publish
             if ($this->config->getSetting('DEFAULT_LAYOUT_AUTO_PUBLISH_CHECKB') == 1 && $this->isChild()) {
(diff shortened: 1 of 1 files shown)


maurofmferrao (approved):
LGTM

mgbaybay (approved):

flashdave:
TEST: all

PR #3883, opened 2026-08-27
Weather: Round coordinates to 2 decimal places.

### LINKED ISSUE #3880: Weather lat/long passed to OpenWeatherMap at full precision
Affected CMS Version: 4.x 
 
The CMS currently sends full-precision latitude and longitude values when requesting weather data from the OpenWeather API.

Example:

Display A: `51.5073512`, `-0.1277583`
Display B (1 metre away): `51.5073601`, `-0.1277590`

These are treated as different locations, so:

- Separate API requests are made.
- Separate cache entries are created.
- More API tokens are consumed.
- Nearby displays can show slightly different weather because they're using different cached responses.

**Expected:** Both displays resolve to the same rounded coordinates, share one  cached API response, and show identical current observations.

PeterMis:
fixes xibosignage/xibo#3880


dasgarner (approved):

ifarzana:
TEST: all

PR #3885, opened 2026-08-27
Schedule: Fix Sync event displayGroup assignment.

### LINKED ISSUE #3927: Schedule : Sync event does not assign any displayGroup to the event.
In release44 frontend was sending `displayGroupIds` array in the payload with each Display specific displayGroupId from the selected sync group.

That was lost when frontend was migrated to React - as a result no displayGroup is ever assigned to the sync event, meaning they will be never returned to players in getForXmds().

This issue was masked by the SeedDatabaseTask/phpunit test we have for sync events, as displaygroups were explicitly assigned in the task itself.

PeterMis:
fixes xibosignage/xibo#3927

Frontend fix should suffice, backend side fix only acts as an additional defense fallback - happy to remove it if we don't think it's needed.


maurofmferrao (approved):
LGTM

ifarzana:
TEST: all

PR #3891, opened 2026-08-28
Settings: Fixed unable to clear elevate logging value

mgbaybay:
### Backend Changes
-  Fixed clearing the "Elevate Log Until" setting not persisting
- Fixed "Elevate Log Until" audit log entries showing 1970-01-01 instead of "Never" for unset values

-------
Relates to https://github.com/xibosignage/xibo/issues/3913


PeterMis (approved):

flashdave:
TEST: all

PR #3894, opened 2026-09-01
XMDS : Fix schedule with recurring command events.

### LINKED ISSUE #3931: 4.5.1: a **repeating** Command event stops every player on that display getting its schedule
**Summary**
Schedule a Command event **with Repeats set** to a display, and the CMS's XMDS `Schedule` call for
that display starts returning HTTP 500. Every player on that display then shows
_> Schedule: Unable to get the schedule. Please check the CMS logs._

and stops picking up anything new — new layouts, republished layouts, schedule changes. It keeps
playing its last-known-good schedule, so from a distance the screen looks fine.
The failure is server-side PHP and every XMDS version (`v4`–`v7`) faults identically, so any player of any type on any version behaves the same way.

**Steps to reproduce**

1. Schedule any layout to the display. Confirm the player is updating normally.
2. **Schedule → Events → Add Event.** Event Type **Command**, pick your command, pick the display,
   and on step **4 Optional → Repeats** set **Repeats = `Day`**, **Repeat every = `1`**. Save.
3. Let the player poll, or call XMDS directly
4. Observe the response, and the player's CMS Status screen.
5. Now edit that same event, set **Repeats = `None`**, save, and repeat step 3.

**Expected**
HTTP **200** and a schedule containing the layout and the command:

```xml
<schedule generated="…" filterFrom="…" filterTo="…">
  <layout file="6" fromdt="…" todt="…" scheduleid="…" … />
  <command date="2026-09-01 14:00:00" scheduleid="…" code="bug_reboot"/>
  …
</schedule>
```
**Actual**

HTTP **500** and a SOAP fault:

```xml
<faultstring>Carbon\Carbon::createFromTimestamp(): Argument #1 ($timestamp) must be of type
string|int|float, null given, called in /var/www/cms/lib/Xmds/Soap.php on line 1377</faultstring>
```
Step 5 (Repeats = None) returns **HTTP 200** again, with the one-off Command event still scheduled.
That single toggle is the whole difference.

**Which Repeats settings are affected**

Each tested alone on 4.5.1, cache dropped before every call:

| Repeats | Event created **today** | Notes |
|---|---|---|
| None (one-off) | ✅ 200 | the control — commands themselves are fine |
| Minute | ❌ **500** | |
| Hour | ❌ **500** | |
| Day | ❌ **500** | |
| Week | ❌ **500** | |
| Week + *Repeats on* Mon/Wed | ❌ **500** | a second, separate code path — worth its own test case |
| Month | ✅ 200 | **⚠️ see below — not immune** |
| Year | ✅ 200 | **⚠️ see below — not immune** |
| Day repeat on an **Always** daypart | ✅ 200 | short-circuits before the faulting code |

> **⚠️ Do not read Month/Year as safe.** They passed only because an event created *today* has no
> repeat instance inside the window yet — the one instance in range is the original. Back-date the
> original so a *repeat* lands in the window and both fault:
>
> | Fixture | Result |
> |---|---|
> | Month/1, original dated **2026-08-01** | ❌ **500** |
> | Year/1, original dated **2025-09-15** | ❌ **500** |
>
> In other words: **every Repeat type is affected**; a monthly event simply breaks next month rather
> than today. A tester who creates a monthly command event and sees it work would sign off wrongly.
> This is the single most important thing to get right in the test case.

PeterMis:
fixes xibosignage/xibo#3931

### DIFF
--- a/lib/Xmds/Soap.php
+++ b/lib/Xmds/Soap.php
@@ -1374,13 +1374,17 @@ class Soap
                     if ($isSyncTimezone) {
                         $fromDt = Carbon::createFromTimestamp($scheduleEvent->fromDt, $this->display->timeZone)
                             ->format(DateFormatHelper::getSystemFormat());
-                        $toDt = Carbon::createFromTimestamp($scheduleEvent->toDt, $this->display->timeZone)
-                            ->format(DateFormatHelper::getSystemFormat());
+                        $toDt = ($scheduleEvent->toDt === null)
+                            ? null
+                            : Carbon::createFromTimestamp($scheduleEvent->toDt, $this->display->timeZone)
+                                ->format(DateFormatHelper::getSystemFormat());
                     } else {
                         $fromDt = DateFormatHelper::createFromTimestamp($scheduleEvent->fromDt)
                             ->format(DateFormatHelper::getSystemFormat());
-                        $toDt = DateFormatHelper::createFromTimestamp($scheduleEvent->toDt)
-                            ->format(DateFormatHelper::getSystemFormat());
+                        $toDt = ($scheduleEvent->toDt === null)
+                            ? null
+                            : DateFormatHelper::createFromTimestamp($scheduleEvent->toDt)
+                                ->format(DateFormatHelper::getSystemFormat());
                     }
 
                     $scheduleId = $row['eventId'];


mgbaybay (approved):

flashdave:
TEST: all

PR #3900, opened 2026-09-03
XMR: Fix cannot seal message error.

### LINKED ISSUE #3932: XMR: Request Screen Shot is delayed
When **Request Screen Shot** is used on a Tizen player, the screenshot is not received immediately.

The CMS accepts the request, but the immediate message to the player fails. The screenshot is only received when the player next communicates with the CMS during its normal **Collect Interval**.

### Steps to reproduce

1. Register a Tizen player against the CMS.
2. Go to **Displays**.
3. Open the display menu and select **Request Screen Shot**.
4. Check the CMS logs.
5. Wait for the player's next Collect Interval.

### Actual result

The request appears to be accepted successfully in the CMS, but the screenshot does not arrive immediately.

The CMS log contains:

```text
Player action connection failed. E = Cannot seal message
```

The screenshot eventually appears after the player's normal Collect Interval.

### Expected result

The screenshot should be requested from the player immediately and should become available shortly after clicking **Request Screen Shot**, rather than waiting for the next Collect Interval.

### Scope

* Android `client_code < 408` uses the legacy path are affected.
* Android `client_code >= 408` uses WebSocket XMR are unaffected.
* Tizen (`sssp`) has no equivalent version check and always uses the legacy path.

PeterMis:
fixes xibosignage/xibo#3932

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -164,6 +164,7 @@ ARG GIT_COMMIT=prod
 
 # Setup persistent environment variables
 ENV CMS_DEV_MODE=false \
+    OPENSSL_CONF=/etc/ssl/openssl-legacy.cnf \
     INSTALL_TYPE=docker \
     XMR_HOST=xmr \
     CMS_SERVER_NAME=localhost \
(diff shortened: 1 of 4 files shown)


dasgarner (approved):

ifarzana:
TEST: all

PR #3905, opened 2026-09-03
Library: Fix unusedOnly filter.

### LINKED ISSUE #3937: Media Manager Dashboard: Unused Media returns no results when Dataset Image columns contain NULL values
The **Unused Only** filter on the Media page can return an empty list when an Image-type Dataset column contains a `NULL` value.

### Steps to Reproduce

1. Create a Dataset with an **Image** column.
2. Add at least one Dataset row without a value in the Image column, resulting in `NULL`.
3. Upload a new image to the Media Library.
4. Do not use the image anywhere in the CMS.
5. Navigate to **Media Manager Dashboard**.
6. Observe the Unused Media section.

### Actual

The unused media list is empty, even though there are media items that are not being used anywhere.
No error or warning is displayed.

### Expected

The **Unused Only** filter should return media that are genuinely unused.
`NULL` values in Dataset Image columns should not cause all unused media to be excluded.

PeterMis:
fixes xibosignage/xibo#3937

### DIFF
--- a/lib/Factory/MediaFactory.php
+++ b/lib/Factory/MediaFactory.php
@@ -695,7 +695,8 @@ class MediaFactory extends BaseFactory
                     $dataSetId = $sanitizedDataSet->getInt('dataSetId');
                     $heading = $sanitizedDataSet->getString('heading');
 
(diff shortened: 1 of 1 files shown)


dasgarner (approved):

ifarzana:
TEST: all

PR #3906, opened 2026-09-04
Layout: fixed publish modal doesn't pre-fill existing scheduled publishedDate

mgbaybay:
### Frontend Changes
  - Fixed the Publish modal on the Layouts list page always defaulting to "Publish Now" with no date, even when the layout already had a scheduled publish date set (e.g. via the Designer's publish form)
  - Applied the same fix to the Templates list page's Publish modal, which shared the same gap
  - Removed a stale comment claiming scheduled publishing "is still not implemented in the BE"
-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1840


maurofmferrao (approved):
LGTM

Kent0425:
TEST: all

PR #3908, opened 2026-09-04
Library: fixed nav permission visibility

mgbaybay:
### Frontend Changes  
- Fixed the "Library" nav item (and other permission-gated sections/routes) staying visible and accessible in the SPA after an admin disabled the corresponding feature for a user's group, when that user already had the SPA open — `user.features` was only ever fetched once at initial load and never refreshed afterward
- Extended the existing focus/visibility-triggered session check (`SessionExpiredModal`) to also refresh the user's permissions, instead of only checking session validity — reuses the same trigger and guard already in place rather than adding polling or router-loader revalidation

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1762


chaolen (approved):

ifarzana:
TEST: all

PR #3913, opened 2026-09-07
Layout: Add additional info on layout status

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1839

### DIFF
--- a/frontend/src/pages/Design/Layouts/LayoutConfig.tsx
+++ b/frontend/src/pages/Design/Layouts/LayoutConfig.tsx
@@ -256,6 +256,8 @@ export interface LayoutActionsProps {
   canSaveTemplate?: boolean;
   canTag?: boolean;
   formatDateTime: (value: DateLike) => string;
+  timeZone?: string;
+  locale?: string;
   onPreview?: (row: Layout) => void;
   onMiniPreview?: (row: Layout, kind: 'published' | 'draft') => void;
   onDelete: (id: number) => void;
@@ -515,6 +517,8 @@ export const getLayoutColumns = (props: LayoutActionsProps): ColumnDef<Layout>[]
     t,
     showDescriptionId,
     formatDateTime,
+    timeZone,
+    locale,
     canTag = false,
     onTagClick,
     selectedTagIds,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Design/Layouts/LayoutConfig.tsx
```diff
@@ -598,10 +600,32 @@ export const getLayoutColumns = (props: LayoutActionsProps): ColumnDef<Layout>[]
...
       cell: (info) => {
-        const status = info.getValue<string>();
-        return <StatusCell label={status} type={status === 'Published' ? 'success' : 'neutral'} />;
+        const row = info.row.original;
+        const status = row.publishedStatus;
+        let badgeType: UIStatus = status === 'Published' ? 'success' : 'neutral';
+        let extraInfo = '';
+
+        if (row.publishedDate) {
+          const published = DateTime.fromSQL(row.publishedDate);
```
maurofmferrao: `publishedDate` comes back as a raw string in the CMS's configured `defaultTimezone`, not the browser's timezone. Parsing it with a bare `DateTime.fromSQL` interprets it in the browser's local zone instead, so on any install where those differ, this diff is off by the offset and can flip "Publishing" into "Publish failed" (or vice versa).

Suggestion:
```
const published = DateTime.fromSQL(row.publishedDate, { zone: timeZone });
```

frontend/src/pages/Design/Layouts/LayoutConfig.tsx
```diff
@@ -598,10 +600,32 @@ export const getLayoutColumns = (props: LayoutActionsProps): ColumnDef<Layout>[]
...
+          const diffMinutes = published.diff(DateTime.now(), 'minutes').minutes;
+
+          if (diffMinutes < -5) {
+            extraInfo = t('Publish failed.');
+            badgeType = 'danger';
+          } else {
+            extraInfo = `${t('Publishing')} ${published.toRelative()}`;
```
maurofmferrao: Two i18n problems here. `toRelative()` with no locale always renders in the browser's language, not the CMS language the user picked, and gluing a translated word to an untranslated fragment breaks word order for languages that don't put the verb first.

`formatDateTime` is already threaded into this file's props from `useDateFormatter()` (line 517), I'd thread `locale` (and `timeZone`, needed for the diff calc above) the same way, then build one interpolated string using this codebase's existing pattern (see `t('Last seen {{date}}', ...)` in TimeConnectedRow.tsx):
```ts
extraInfo = t('Publishing {{time}}', {
  time: published.setLocale(locale ?? 'en').toRelative(),
});

maurofmferrao (changes requested):
LGTM, just 2 minor suggestions.

maurofmferrao (approved):
LGTM

flashdave:
TEST: all

PR #3916, opened 2026-09-08
Applications: Added auditing

mgbaybay:
### Backend Changes  
- Adds audit logging to the OAuth Application entity's add/edit/delete lifecycle (Added, Saved, Key changed, Deleted)
- Adds set-based audit diffing for scope and redirect-URI changes, sourced from pre-change snapshots the controller captures via setOriginalValue() to avoid an extra DB query.

-------
Relates to https://github.com/xibosignageltd/xibo-private/issues/1368

### DIFF
--- a/lib/Entity/Application.php
+++ b/lib/Entity/Application.php
@@ -183,13 +183,16 @@ class Application implements \JsonSerializable, ClientEntityInterface
         // Get scopes
         $this->scopes = $this->applicationScopeFactory->getByClientId($this->key);
 
+        // Snapshot so save() can audit what changed, regardless of caller.
+        $this->setOriginalValue('redirectUris', array_map(fn($uri) => $uri->redirectUri, $this->redirectUris));
+        $this->setOriginalValue('scopeIds', array_map(fn($scope) => $scope->id, $this->scopes));
+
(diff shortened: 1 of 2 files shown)


lib/Entity/Application.php
```diff
@@ -188,18 +189,36 @@ public function load(): static
...
-        if ($this->key == null || $this->key == '') {
+        $isNew = ($this->key == null || $this->key == '');
+
+        if ($isNew) {
             // Make a new secret.
             $this->resetSecret();
 
             // Add
             $this->add();
+            $this->audit(0, 'Added', ['name' => $this->name]);
```
dasgarner: Why is the first parameter 0?

dasgarner (changes requested):

lib/Entity/Application.php
mgbaybay: Audit's `entityId` column is an integer, but the Application's ID (key) is a string, so we can't pass it directly. I think `0` is also used for the Display bandwidth update, for the same reason.

dasgarner (approved):

flashdave:
TEST: all

PR #3918, opened 2026-09-08
Permissions: Fixed folder permissions and resolution query

mgbaybay:
### Frontend Changes
- Added a "Settings Updated" success toast when the Edit User modal's save completes
- Fixes fetchGroupFolderPermissions() to request unpaginated results instead of a folder-count-sized page, so a user's own permission group can no longer be silently excluded from the response.

### Backend Changes
- Fixes User::permissionsMultiGrid() so each user group's per-folder permission entry is keyed by its real folder ID instead of being appended positionally, preventing permissions from displaying against the wrong folder when the Folder Permission tab is reopened.
- Fixes Resolution query so its width/height-based sort actually reaches the ORDER BY clause instead of being silently dropped and falling back to alphabetical order

-------
Relates to https://github.com/xibosignage/xibo/issues/3938, https://github.com/xibosignageltd/xibo-private/issues/1706

### DIFF
--- a/frontend/src/pages/Administration/Users/components/AddEditUserModal.tsx
+++ b/frontend/src/pages/Administration/Users/components/AddEditUserModal.tsx
@@ -503,7 +503,7 @@ export default function AddEditUserModal({
           await saveFolderPermissions(savedUser.groupId);
         }
 
-        notify.success(t('Settings Updated'));
+        notify.success(t('User saved'));
         onSuccess();
         onClose();
       } catch (err: unknown) {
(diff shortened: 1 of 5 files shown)


maurofmferrao (dismissed):
LGTM

dasgarner (dismissed):

frontend/src/pages/Administration/Users/components/AddEditUserModal.tsx
```diff
@@ -502,6 +503,7 @@ export default function AddEditUserModal({
           await saveFolderPermissions(savedUser.groupId);
         }
 
+        notify.success(t('Settings Updated'));
```
maurofmferrao: Should this be "User Saved" instead?

maurofmferrao (dismissed):
LGTM, just a minor copy suggestion that I'm not sure makes sense.

maurofmferrao (approved):

dasgarner (approved):

Kent0425:
TEST: all

PR #3920, opened 2026-09-08
Tags: Comma in group name is wrong in "Sharing" column

maurofmferrao:
relates to xibosignage/xibo#3945


dasgarner:
I don't think this is a sensible change as it stands here. Every view in the CMS goes from 1 query to 2 queries (or from N queries to N+1 queries).

Can we explore other options?

For example, could we concatenate the string from the inline view with a different delimiter (|| perhaps) and then make an array PHP side instead?

dasgarner (approved):

ifarzana:
TEST: all

PR #3928, opened 2026-09-09
Search text is not persisted when navigating away and returning to a page

maurofmferrao:
relates to xibosignageltd/xibo-private#1712


mgbaybay (approved):

flashdave:
TEST: all

PR #3934, opened 2026-09-10
Schedule: validate recurrence end is after event start date

chaolen:
relates to: https://github.com/xibosignageltd/xibo-private/issues/1849

### DIFF
--- a/frontend/src/components/ui/modals/ScheduleEventModal.tsx
+++ b/frontend/src/components/ui/modals/ScheduleEventModal.tsx
@@ -1207,6 +1207,9 @@ export default function ScheduleEventModal({
         mapped.shareOfVoice
       ) {
         setCurrentStep(timeStepIndex);
+      } else if (mapped.recurrenceDetail || mapped.recurrenceRange) {
+        setCurrentStep(optionalStepIndex);
+        setOptionalTab('repeats');
       }
       return;
     }
@@ -2358,6 +2361,7 @@ export default function ScheduleEventModal({
                         value={draft.recurrenceRange}
                         onChange={(value) => updateDraft('recurrenceRange', value)}
                         helpText={t('Optionally select the date this event should stop repeating.')}
+                        error={formErrors.recurrenceRange}
                       />
                     </>
                   )}
--- a/frontend/src/schema/scheduleEvent.ts
+++ b/frontend/src/schema/scheduleEvent.ts
@@ -210,6 +210,19 @@ export const getScheduleEventSchema = (
         });
       }
 
+      if (
+        data.recurrenceType &&
+        data.recurrenceRange &&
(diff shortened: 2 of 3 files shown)


maurofmferrao (changes requested):
This PR adds validation for the recurrence end date, but the field it targets isn't wired to display it. Two things in `ScheduleEventModal.tsx` (not part of this diff, so flagging here instead of inline):

1. The "Until" `DatePickerInput` (around line 2356) doesn't pass an `error` prop, unlike every other date field in the form (see fromDt/toDt at lines 2014, 2024, 2035, 2041). Right now `formErrors.recurrenceRange` is set but never rendered, so the message only shows up in the generic toast, not next to the field:

```
    <DatePickerInput
      label={t('Until')}
      value={draft.recurrenceRange}
      onChange={(value) => updateDraft('recurrenceRange', value)}
      helpText={t('Optionally select the date this event should stop repeating.')}
      error={formErrors.recurrenceRange}
    />
```

2. The step-navigation logic in `handleFinish` (around lines 1202-1210) doesn't have a branch for `mapped.recurrenceRange`, so if that's the only validation error, the wizard doesn't navigate the user to the "Repeats" tab where the field lives, they're left on whatever step they were on with just the toast to go on.

maurofmferrao (dismissed):

chaolen:
requesting another review for this PR resolving the new flagged issue:

- Issue 1 — Daypart save fails when associated events have expiring recurrence                                                                                                           
Changing a daypart's times triggers a schedule split that could create an event with Until before its start date, failing the new recurrence validation.
Fix: Skip creating the split schedule when the recurrence window has already passed relative to the new start date.                                                                    
                                                                                                                                                                                          
- Issue 2 — "Make a Copy" fails on events with expired recurrence range                                                                                                                  
Copying an event that has Until before its start date fails validation with no fields available to correct it.
Fix: Clear the inherited recurrenceRange before saving the copy when it's invalid.
 
- Issue 3 — Repeats tab hidden from users without Schedule Reminders feature
The Repeats and Reminder tabs were gated behind a single permission check, so users without Schedule Reminders couldn't see or edit recurrence settings.
Fix: Show the Repeats tab based on showRepeatReminder alone, keeping the Reminder tab gated behind canSetReminders.

maurofmferrao (dismissed):
LGTM

chaolen:
Another change has been applied:
Earlier daypart edits could create schedules where Until < start date. The previous commit prevents new ones, but existing ones block future edits. This fix deletes those dead schedules during daypart update.

maurofmferrao (dismissed):
LGTM

maurofmferrao (approved):
LGTM

flashdave:
TEST: all

PR #3941, opened 2026-09-14
Schedule: Preserve reminders when editing events

chaolen:
relates to: https://github.com/xibosignage/xibo/issues/3942

Editing a schedule event deletes and recreates all reminders, resetting `lastReminderDt` to 0 and causing already-fired reminders to fire again. The frontend never sent `reminder_scheduleReminderId` in the edit payload, so the backend treated every reminder as new.
                                                                                                                                       
- Thread `scheduleReminderId` through draft type, modal payload, API call, and PHP controller so reminders are updated in place
- Backend always loads reminders for GET and PUT schedule endpoints instead of requiring the embed param
- Enrichment fetch now merges only missing fields (reminders, criteria, geoLocation, etc.) instead of replacing the entire draft, preventing user edits from being overwritten
- Added loading indicators to Reminder, Geo Location, and Criteria tabs while event details are being fetched
- Display group multi-select shows _"Loading..."_ instead of raw numeric IDs while labels resolve


frontend/src/components/ui/modals/ScheduleEventModal.tsx
maurofmferrao: `canFinish` is true from the first render in edit mode, so Save is clickable before the enrichment fetch resolves. Saving that fast sends empty `reminders`, which makes the backend delete all existing reminders, exactly the bug this PR fixes. I would `isEnriching` here too.

```
disabled: isPending || isEnriching || !isStepValid,
```

frontend/src/pages/Schedule/Schedule/components/DisplayGroupMultiSelect.tsx
```diff
@@ -400,6 +400,10 @@ export function DisplayGroupMultiSelect({
   const { getReferenceProps, getFloatingProps } = useInteractions([click, dismiss]);
 
   const totalSelected = value.displaySpecificGroupIds.length + value.displayGroupIds.length;
+  const allLabelsResolved =
```
maurofmferrao, line 417: If a label fetch fails or a display/group was deleted, its `labelsCache` entry never fills in and this stays "Loading..." forever, no fallback to the raw ID like before. Could we fall back to the id once the fetch attempt finishes either way?

maurofmferrao (changes requested):
Just 2 quick changes.

chaolen:
- ScheduleEventModal: Disable the Save/Finish button while isEnriching is true. In edit mode, canFinish is true from the first render, so without this guard the user could save before the enrichment fetch resolves, sending empty reminders and deleting existing ones.
- DisplayGroupMultiSelect: Fall back to the raw ID when a label fetch fails or the API returns fewer rows than requested (e.g. a deleted display/group). Previously the component would stay stuck on "Loading..." forever. Extracted a shared `fillMissing` helper to deduplicate the fallback logic across both display and group fetch paths.

maurofmferrao (dismissed):
LGTM

flashdave:
TEST: all

maurofmferrao (dismissed):
LGTM

maurofmferrao (approved):

ifarzana:
TEST: all

PR #3942, opened 2026-09-15
Schedule: Fix date picker timezone day offset

chaolen:
relates to: https://github.com/xibosignage/xibo/issues/3951

When the CMS timezone differs from the browser's local timezone, the date picker in the schedule event modal shows the wrong date. Selecting "today" in the calendar displays a day before or after in the preview and input field, because the browser-local Date was converted to an ISO string without anchoring it to the CMS timezone first.

- Added `anchorDateOnly()` in DatePicker to anchor calendar-picked dates to midday in the CMS timezone before converting to ISO, ensuring the UTC instant stays on the same calendar day regardless of timezone offset
- Applied the anchor to both single and range modes when the time picker is not shown (named daypart events)
- Fixed the footer preview to anchor the date before formatting, so it matches what the input field will display after clicking Apply


frontend/src/components/ui/DatePicker.tsx
```diff
@@ -229,11 +229,27 @@ export default function DatePicker({
     !!rangeValues &&
     rangeValues.from > rangeValues.to;
 
+  const anchorDateOnly = (date: Date): Date => {
```
maurofmferrao, line 232: `anchorDateOnly` is a `const` declared here at line 232, but `getRangeValues()` is called immediately at line 225, before that declaration executes on this render pass.

The fix is to move the `anchorDateOnly` declaration above `getRangeValues` (before line 217), so it's initialized before anything calls it.

maurofmferrao (changes requested):
There a blocker with a reference to a variable before it's declared.

maurofmferrao (approved):
LGTM

ifarzana:
TEST: all

PR #3943, opened 2026-09-15
Display Groups: fixed delimeter in sharing column

mgbaybay:
### Backend Changes  
- Updated display group query delimeter

-------
Relates to https://github.com/xibosignage/xibo/issues/3945


PeterMis (approved):
LGTM

Since `@` can technically still collide, it can potentially hit the same issue as `,` did.
We could look at using `JSON_ARRAYAGG` as initially indented just with a single-level correlated subquery ( i.e. `JSON_ARRAYAGG(group.group)` without the derived table wrapper )

That said, it's not a blocker, perhaps something to keep in mind for the future, as it stands your fix is safe, minimal and unblocks the regression introduced in the earlier PR, so good to go

ifarzana:
TEST: Displays

PR #3944, opened 2026-09-15
Single item download loader and save as fix

maurofmferrao:
relates to xibosignage/xibo#3946
relates to xibosignage/xibo#3947

- Fix zip file download


mgbaybay (approved):

ifarzana:
TEST: all

PR #3946, opened 2026-09-16
Not found Page: Serve branded React page for unmatched routes

chaolen:
relates to: https://github.com/xibosignage/xibo/issues/3917


PeterMis (dismissed):

chaolen:
Few more changes to this PR:

- The branded 404 page issued a new CSRF token on every visit, but because routing throws before the session middleware runs, the token was never stored — overwriting the browser's real cookie and breaking the next save. Skip CSRF issuance for 404 since the page has no forms.

- The 404 icon also used hardcoded Xibo blue instead of the brand color. Replaced with `color-mix()` derived from 
`--brand-primary` so it follows the brand on re-branded installs.

PeterMis (approved):

flashdave:
TEST: all

PR #3950, opened 2026-09-17
4.5.3 enhancements

maurofmferrao:
relates to xibosignageltd/xibo-private#1870

- Add Schedule button to Layout Editor
- Android video engine default:  default instead of exoplayer
- Display Groups quick action: Members instead of Edit
- Schedule form: populate Display/Group names on load instead of showing raw IDs


frontend/src/pages/Displays/DisplayGroup/DisplayGroupConfig.tsx
```diff
@@ -206,13 +206,24 @@ export const getDisplayGroupItemActions = ({
       }
     };
 
+    if (canModify && canEdit && Number(displayGroup.isDynamic) === 0) {
+      actions.push({
+        label: t('Members'),
+        icon: Users,
+        onClick: () => openMembersModal(displayGroup),
+        isQuickAction: true,
```
PeterMis, line 214: I *think* with this quick action swap between members and edit, may cause two tests in the DisplayGroup.edit.test.tsx might fail ( as they rely on edit being the quickaction button )

PeterMis (changes requested):
Looks great, but we may need to look at the tests before we merge, specifically the displayGroup test as per the comment below

ifarzana (approved):
looks good to me

PeterMis (approved):

ifarzana:
TEST: all

